PR #1, opened 2023-12-07
Make cleanAll idempotent

keithc-ca:
Use "rm -f" so it doesn't fail if the header files don't exist.

(Discovered while reviewing https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/238).


keithc-ca:
This make `jgskit.mak` consistent with all other `jgskit.*.mak` files.

johnpeck-us-ibm (approved):

PR #8, opened 2023-12-12
Change the OCK and JGskit load path on Windows platform

taoliult:
When loading the OCK and JGskit libraries on Windows platform, they will be loaded from the "bin" directory, instead of the "lib" directory on other platforms. So updated the libraries load paths to use "bin" directory on Windows platform.


farshadasl (approved):

johnpeck-us-ibm (approved):
Make sure this gets updated in readme so that customers know were they need to place things in the JVM

jasonkatonica (approved):
LGTM

PR #16, opened 2024-02-21
Update dependant mvn libraries

jasonkatonica:
This update splits out the `common-compress` library into its own
dependency definition. This allows us to update to the latest
version of this library.

This update also updates `maven-compiler-plugin` and `junit-bom`
libraries to the latest versions available.


KostasTsiounis (approved):
LGTM

PR #20, opened 2024-02-23
Use KnownOIDs class for known curve constants

jasonkatonica:
While developing necessary changes for Java 21 we noticed that the CurveUtil class needed to make use of the sun.security.util.KnownOIDs class. This update simply matches that required behavior for JDK 21 to define various OID constants.


taoliult (approved):
LGTM

WilburZjh (approved):

PR #23, opened 2024-02-29
Remove RSA Cipher from OpenJCEPlusFIPS

WilburZjh:
According to the OCK policy, RSA cipher is not allowed to be used in FIPS140-3. Should be eliminated.

Updated the tests to force the code executing `Cipher.getInstance("RSA", "OpenJCEPlusFIPS")`, and catch the exception to ensure the OpenJCEPlusFIPS provider doesn't contain the `RSA` crypto service in Cipher.

Port from [main](https://github.com/IBM/OpenJCEPlus/commit/1c30d2888570daa95f214003aa5f8b1db2b4d954)
Signed-off-by: WilburZjh <wilburzjh@gmail.com>


jasonkatonica (approved):
LGTM

PR #24, opened 2024-03-01
Remove MD5 and SHA1 algorithms from OpenJCEPlusFIPS provider

jasonkatonica:
MD5 and SHA1 algorithms do not strictly adhere to FIPS policies.

These algorithms should be removed from registration within the OpenJCEPlusFIPS provider.

### DIFF
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestRSASignatureChunkUpdate.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestRSASignatureChunkUpdate.java
@@ -10,6 +10,7 @@ package ibm.jceplus.junit.base;
 
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
+import java.security.NoSuchAlgorithmException;
 import java.security.Signature;
 import java.security.SignatureException;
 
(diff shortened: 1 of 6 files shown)


src/test/java/ibm/jceplus/junit/base/BaseTestRSASignatureChunkUpdate.java
```diff
@@ -38,7 +39,17 @@ public void setUp() throws Exception {
...
+            if (providerName.contains("FIPS")) {
+                assertEquals("no such algorithm: SHA1WithRSA for provider OpenJCEPlusFIPS", nsae.getMessage());
+                return;
+            } else {
+                throw nsae;
+            }
+        }
+        
         testSignatureChunkUpdate(1024, "SHA256WithRSA", 100);
         testSignatureChunkUpdate2(1024, "SHA1WithRSA", 100); // Test signing failure with OpenJCEPlusFIPS
```
KostasTsiounis, line 54: Shouldn't this be removed too?

KostasTsiounis (changes requested):

WilburZjh:
Should we remove this [TestMD5.java](https://github.com/IBM/OpenJCEPlus/blob/main/src/test/java/ibm/jceplus/junit/openjceplusfips/TestMD5.java) files or force the code to run `MessageDigest.getInstance("MD5", providerName)` and compare the exception message?

src/test/java/ibm/jceplus/junit/base/BaseTestRSASignatureChunkUpdate.java
jasonkatonica: Looking into these paths it appears that this does not currently fail because the underlying method `testSignatureChunkUpdate2` appears to have a FIPS specific check to ensure a error message:

https://github.com/IBM/OpenJCEPlus/blob/050187a432d31b0ca96eaae01922378ff11c8aa6/src/test/java/ibm/jceplus/junit/base/BaseTestRSASignatureChunkUpdate.java#L145C21-L145C22

I would be ok either way although this seems to be ok as is handling the exception in that location.

jasonkatonica:
> Should we remove this [TestMD5.java](https://github.com/IBM/OpenJCEPlus/blob/main/src/test/java/ibm/jceplus/junit/openjceplusfips/TestMD5.java) files or force the code to run `MessageDigest.getInstance("MD5", providerName)` and compare the exception message?

Yes good catch here. I updated this PR to run the TestMD5.class file and the TestMD5 class simply expects an error to occur now trying to instantiate a MD5 instance from `OpenJCEPlusFIPS` provider.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestSHA1.java
```diff
@@ -27,6 +31,15 @@ public TestSHA1() {
         super(Utils.TEST_SUITE_PROVIDER_NAME);
     }
 
+    public static void testSHA1Cipher() throws Exception {
+        try {
+            Cipher.getInstance("SHA1", Utils.TEST_SUITE_PROVIDER_NAME);
+        } catch (NoSuchAlgorithmException nsae) {
+            assertEquals("No such algorithm: SHA1", nsae.getMessage());
+            return;
```
KostasTsiounis, line 39: Not sure whether the `return` is needed here.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestMD5.java
```diff
@@ -27,6 +31,15 @@ public TestMD5() {
         super(Utils.TEST_SUITE_PROVIDER_NAME);
     }
 
+    public static void testMD5Cipher() throws Exception {
+        try {
+            Cipher.getInstance("MD5", Utils.TEST_SUITE_PROVIDER_NAME);
+        } catch (NoSuchAlgorithmException nsae) {
+            assertEquals("No such algorithm: MD5", nsae.getMessage());
+            return;
```
KostasTsiounis, line 39: Not sure whether the `return` is needed here.

KostasTsiounis (changes requested):

PR #27, opened 2024-03-04
Check the RSA keysize in Signature

JinhangZhang:
At least 2048 bits of RSA key can be
used for Sign in Signature. However,
current openjceplusfips provider can
accept a RSA key which size is smaller
than 1024.

Add a check in the engineInitSign()
function to filter the keysize.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/RSAKeyFactory.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/RSAKeyFactory.java
@@ -25,6 +25,9 @@ import java.security.spec.RSAPrivateCrtKeySpec;
 import java.security.spec.RSAPrivateKeySpec;
 import java.security.spec.RSAPublicKeySpec;
 import java.security.spec.X509EncodedKeySpec;
+import java.util.Arrays;
+import java.util.List;
+
 import com.ibm.crypto.plus.provider.RSAUtil.KeyType;
 
 @SuppressWarnings({"removal", "deprecation"})
@@ -34,6 +37,8 @@ class RSAKeyFactory extends KeyFactorySpi {
     public final static int MIN_MODLEN_FIPS = 2048;
     public final static int MIN_MODLEN_FIPS_PUB = 1024; //FIPS currently allows signature verification on this size key
     public final static int MAX_MODLEN = 16384;
+    public final static List<Integer> ALLOWABLE_MODLEN_FIPS_SIGN = Arrays.asList(2048, 3072, 4096);
+    public final static List<Integer> ALLOWABLE_MODLEN_FIPS_VERIFY = Arrays.asList(1024, 2048, 3072, 4096);
 
     /*
      * If the modulus length is above this value, restrict the size of the
@@ -121,6 +126,40 @@ class RSAKeyFactory extends KeyFactorySpi {
             throw new InvalidKeyException(
                     "RSA exponents can be no longer than " + MAX_RESTRICTED_EXPLEN + " bits "
                             + " if modulus is greater than " + MAX_MODLEN_RESTRICT_EXP + " bits");
+        }
+    }
+
+    /**
+     * Check the length of an RSA key modulus/exponent to make sure it is not
+     * too short or long. Some impls have their own min and max key sizes that
+     * may or may not match with a system defined value.
+     *
+     * @param modulusLen
+     *            the bit length of the RSA modulus.
+     * @param exponent
+     *            the RSA exponent
+     * @param minModulusLen
+     *            if > 0, check to see if modulusLen is at least this long,
+     *            otherwise unused.
+     * @param maxModulusLen
+     *            caller will allow this max number of bits. Allow the smaller
+     *            of the system-defined maximum and this param.
+     * @param specificModulesLen
+     *            specific module length for sign/verify.
+     *
+     * @throws InvalidKeyException
+     *             if any of the values are unacceptable.
+     */
+    static void checkKeyLengths(int modulusLen, BigInteger exponent, int minModulusLen,
+            int maxModulusLen, List<Integer> specificModulesLen, String flag) throws InvalidKeyException {
+
+        checkKeyLengths(modulusLen, exponent, minModulusLen, maxModulusLen);
+        if ((specificModulesLen != null) && (!specificModulesLen.contains(modulusLen))) {
+            if (flag.equals("verify")) {
+                throw new InvalidKeyException("In FIPS mode, only 1024, 2048, 3072, or 4096 size of RSA key is accepted.");
+            } else if (flag.equals("sign")){
+                throw new InvalidKeyException("In FIPS mode, only 2048, 3072, or 4096 size of RSA key is accepted.");
+            }
         }
     }
 
--- a/src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2024
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -17,6 +17,9 @@ import java.security.PublicKey;
 import java.security.SignatureException;
 import java.security.SignatureSpi;
 import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.RSAKeyGenParameterSpec;
+import java.util.List;
+
 import com.ibm.crypto.plus.provider.RSAUtil.KeyType;
 import com.ibm.crypto.plus.provider.ock.Signature;
 
@@ -47,6 +50,18 @@ abstract class RSASignature extends SignatureSpi {
             throw new InvalidKeyException("Key is not an RSAPublicKey");
         }
 
+        List<Integer> specificModulesLen = null;
+        int minModulusLen = RSAKeyFactory.MIN_MODLEN_NONFIPS;
+        if (provider.isFIPS()) {
+            minModulusLen = RSAKeyFactory.MIN_MODLEN_FIPS_PUB;
+            specificModulesLen = RSAKeyFactory.ALLOWABLE_MODLEN_FIPS_VERIFY;
+        }
+        RSAKeyFactory.checkKeyLengths(((java.security.interfaces.RSAPublicKey) publicKey).getModulus().bitLength(),
+                                        RSAKeyGenParameterSpec.F4,
+                                        minModulusLen,
+                                        64 * 1024,
+                                        specificModulesLen,
+                                        "verify");
 
         RSAPublicKey rsaPublic = (RSAPublicKey) RSAKeyFactory.toRSAKey(provider, publicKey);
         try {
@@ -83,6 +98,19 @@ abstract class RSASignature extends SignatureSpi {
         if (!(privateKey instanceof java.security.interfaces.RSAPrivateKey)) {
             throw new InvalidKeyException("Key is not an RSAPrivateKey");
         }
+
+        List<Integer> specificModulesLen = null;
+        int minModulusLen = RSAKeyFactory.MIN_MODLEN_NONFIPS;
+        if (provider.isFIPS()) {
+            minModulusLen = RSAKeyFactory.MIN_MODLEN_FIPS;
+            specificModulesLen = RSAKeyFactory.ALLOWABLE_MODLEN_FIPS_SIGN;
+        }
+        RSAKeyFactory.checkKeyLengths(((java.security.interfaces.RSAPrivateKey) privateKey).getModulus().bitLength(),
+                                        RSAKeyGenParameterSpec.F4,
+                                        minModulusLen,
+                                        64 * 1024,
+                                        specificModulesLen,
+                                        "sign");
 
         //RSAPrivateCrtKey rsaPrivate = (RSAPrivateCrtKey) RSAKeyFactory.toRSAKey(provider, privateKey);
         PrivateKey rsaPrivate = (PrivateKey) RSAKeyFactory.toRSAKey(provider, privateKey);
--- a/src/test/java/ibm/jceplus/junit/openjceplusfips/TestAll.java
+++ b/src/test/java/ibm/jceplus/junit/openjceplusfips/TestAll.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2024
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -39,7 +39,7 @@ import junit.framework.Test;
         TestSHA384.class, TestSHA512.class, TestRSAPSSInterop2.class, TestRSAPSSInterop3.class,
         TestSHA3_224.class, TestSHA3_256.class, TestSHA3_384.class, TestSHA3_512.class,
         TestRSAKeyInterop.class, TestRSAKeyInteropBC.class, TestRSAPSS2.class,
-        TestFIPSVerifyOnlyTest.class})
+        TestFIPSVerifyOnlyTest.class, TestRSASignatureWithSpecificSize.class})
 
 public class TestAll {
 
--- a/src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureWithSpecificSize.java
+++ b/src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureWithSpecificSize.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright IBM Corp. 2024
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution.
+ */
+package ibm.jceplus.junit.openjceplusfips;
+
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.Signature;
+import java.security.InvalidKeyException;
+
+import junit.framework.Test;
+import junit.framework.TestSuite;
+
+public class TestRSASignatureWithSpecificSize extends ibm.jceplus.junit.base.BaseTest {
+
+    //--------------------------------------------------------------------------
+    //
+    //
+    static final byte[] origMsg = "this is the original message to be signed I changed to a very long message to make sure enough bytes are there for copying."
+            .getBytes();
+
+    static {
+        Utils.loadProviderTestSuite();
+    }
+
+    //--------------------------------------------------------------------------
+    //
+    //
+    public TestRSASignatureWithSpecificSize() {
+        super(Utils.TEST_SUITE_PROVIDER_NAME);
+    }
+
+    // RSA signature sign allows at least 2048 bits of RSA key to be used for sign a signature.
+    public byte[] doSign(String sigAlgo, byte[] message, PrivateKey privateKey) throws Exception {
+        Signature sign = Signature.getInstance(sigAlgo, Utils.TEST_SUITE_PROVIDER_NAME);
+        try {
+            sign.initSign(privateKey);
+        } catch (InvalidKeyException ike) {
+            if (((java.security.interfaces.RSAPrivateKey) privateKey).getModulus().bitLength() < 2048 ) {
+                if ("RSA keys must be at least 2048 bits long".equals(ike.getMessage())) {
+                    System.out.println("Expected exception msg: <RSA keys must be at least 2048 bits long> is caught for sign.");
+                    return null;
+                }
+            } else {
+                if ("In FIPS mode, only 2048, 3072, or 4096 size of RSA key is accepted.".equals(ike.getMessage())) {
+                    System.out.println("Expected exception msg: <In FIPS mode, only 2048, 3072, or 4096 size of RSA key is accepted.> is caught for sign.");
+                    return null;
+                }
+            }
+            throw ike;
+        }
+        sign.update(message);
+        byte[] signedBytes = sign.sign();
+        return signedBytes;
+    }
+
+    // RSA signature verify allows at least 2048 bits of RSA key to be used for sign a signature.
+    public void doVerify(String sigAlgo, byte[] message, PublicKey publicKey, 
+            byte[] signedBytes) throws Exception {
+        Signature verify = Signature.getInstance(sigAlgo, Utils.TEST_SUITE_PROVIDER_NAME);
+        try {
+            verify.initVerify(publicKey);
+        } catch (InvalidKeyException ike) {
+            if (((java.security.interfaces.RSAPublicKey) publicKey).getModulus().bitLength() < 1024 ) {
+                if ("RSA keys must be at least 1024 bits long".equals(ike.getMessage())) {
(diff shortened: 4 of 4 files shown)


JinhangZhang:
OpenJCEPlus pipeline test[142] is passed

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
```diff
@@ -84,6 +86,22 @@ protected void engineInitSign(PrivateKey privateKey) throws InvalidKeyException
             throw new InvalidKeyException("Key is not an RSAPrivateKey");
         }
 
+        try {
+            if (provider.isFIPS()) {
+                RSAKeyFactory.checkKeyLengths(((java.security.interfaces.RSAPrivateKey) privateKey).getModulus().bitLength(),
+                        RSAKeyGenParameterSpec.F4,
```
jasonkatonica, line 92: Could we limit the size of the modulus used for verification also? I believe that this should be 1024, 2048, 3072, and 4096 for FIPS. We may need an additional test to validate these sizes for the RSA FIPS verification process.

jasonkatonica (changes requested):

jasonkatonica:
Also please rebase upon the latest `main` branch when you get the chance.

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
```diff
@@ -84,6 +86,22 @@ protected void engineInitSign(PrivateKey privateKey) throws InvalidKeyException
             throw new InvalidKeyException("Key is not an RSAPrivateKey");
         }
 
+        try {
+            if (provider.isFIPS()) {
+                RSAKeyFactory.checkKeyLengths(((java.security.interfaces.RSAPrivateKey) privateKey).getModulus().bitLength(),
```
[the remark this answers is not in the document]
KostasTsiounis: Would it be better to overload `checkKeyLengths()` on `RSAKeyFactory` to accept a list of values instead of min and max?

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright IBM Corp. 2023
```
KostasTsiounis, line 2: Update the copyright.

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
```diff
@@ -17,6 +17,8 @@
 import java.security.SignatureException;
 import java.security.SignatureSpi;
 import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.RSAKeyGenParameterSpec;
```
KostasTsiounis, line 20: Update the copyright.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestAll.java
```diff
@@ -39,7 +39,7 @@
         TestSHA384.class, TestSHA512.class, TestRSAPSSInterop2.class, TestRSAPSSInterop3.class,
         TestSHA3_224.class, TestSHA3_256.class, TestSHA3_384.class, TestSHA3_512.class,
         TestRSAKeyInterop.class, TestRSAKeyInteropBC.class, TestRSAPSS2.class,
-        TestFIPSVerifyOnlyTest.class})
+        TestFIPSVerifyOnlyTest.class, TestRSASignatureInteropNonFIPS.class})
```
KostasTsiounis, line 42: Update the copyright.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
```diff
@@ -0,0 +1,82 @@
...
+    //
+    public TestRSASignatureInteropNonFIPS() {
+        super(Utils.TEST_SUITE_PROVIDER_NAME, Utils.PROVIDER_SunRsaSign, 1024);
+    }
+
+    // RSA signature allows at least 2048 bits of RSA key to be used for sign a signature.
+    @Override
+    protected void doSignVerify(String sigAlgo, byte[] message, PrivateKey privateKey,
+            PublicKey publicKey) throws Exception {
+        Signature signing = Signature.getInstance(sigAlgo, Utils.TEST_SUITE_PROVIDER_NAME);
```
KostasTsiounis, line 40: Maybe change the name to `signature`?

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
```diff
@@ -0,0 +1,82 @@
...
+        try {
+            signing.initSign(privateKey);
+        } catch (java.security.InvalidParameterException ipe) {
```
KostasTsiounis, line 43: Would it make sense to import this?

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
```diff
@@ -0,0 +1,82 @@
...
+                assertEquals("RSA keys must be at least 2048 bits long", ipe.getMessage());
+                return;
+            } else {
+                throw ipe;
+            }
+        }
+        signing.update(message);
+        byte[] signedBytes = signing.sign();
+
+        Signature verifying = Signature.getInstance(sigAlgo, Utils.TEST_SUITE_PROVIDER_NAME);
```
KostasTsiounis, line 54: Maybe `verification` would be a better name?

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
```diff
@@ -0,0 +1,82 @@
...
+
+    // RSA signature allows at least 2048 bits of RSA key to be used for sign a signature.
+    @Override
+    protected void doSignVerify(String sigAlgo, byte[] message, PrivateKey privateKey,
+            PublicKey publicKey) throws Exception {
+        Signature signing = Signature.getInstance(sigAlgo, Utils.TEST_SUITE_PROVIDER_NAME);
+        try {
+            signing.initSign(privateKey);
+        } catch (java.security.InvalidParameterException ipe) {
+            if (providerName.equals("OpenJCEPlusFIPS")) {
```
KostasTsiounis, line 44: Isn't the provider always FIPS in this one? If we want to check anyway, maybe we should move the change to the `BaseTestSignatureInterop` and avoid duplicate code.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
```diff
@@ -0,0 +1,82 @@
...
+    //--------------------------------------------------------------------------
+    //
+    //
+    public TestRSASignatureInteropNonFIPS() {
+        super(Utils.TEST_SUITE_PROVIDER_NAME, Utils.PROVIDER_SunRsaSign, 1024);
+    }
+
+    // RSA signature allows at least 2048 bits of RSA key to be used for sign a signature.
+    @Override
+    protected void doSignVerify(String sigAlgo, byte[] message, PrivateKey privateKey,
```
KostasTsiounis, line 38: I think we are missing a few interop cases here, as the method that is overridden was calling other methods that would use different providers for different things. In this case, `OpenJCEPlusFIPS` is used for both sign/verify, while `Sun` is used for key generation.

KostasTsiounis (changes requested):

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
JinhangZhang: Openjceplusfips does not allow a 1024 size of key pair, we have to use other providers such as SunRsaSign to create a 1024 keysize key. 

In this doSignVerify test, atm, we only test whether sign procedure still accepts 1024 key size for RSA or not.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
JinhangZhang: This is a sign procedure, it aims to differentiate with the verify procedure.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
JinhangZhang: Sure, let me rename it.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
KostasTsiounis: So, maybe we'll need a different name, as this is not `interop`. We would need a provider to sign and a different one to verify in order to be `interop`.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
KostasTsiounis: Could be something like `creation`? We can leave it as is if we don't find anything better...

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
KostasTsiounis: Also, maybe you don't need 2 instances of the `Signature`, since the same provider does both. You can just initialize again, I think.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
JinhangZhang: Yes, we use the same provider to do the sign and verify, but a different provider to generate the key pair. 

Didnt come up with a good name for this test, any idea?

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
JinhangZhang: updated

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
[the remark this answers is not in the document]
JinhangZhang: overloaded

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
JinhangZhang: limited

src/test/java/ibm/jceplus/junit/openjceplusfips/TestAll.java
JinhangZhang: updated

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
JinhangZhang: updated

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
```diff
@@ -47,6 +49,23 @@ protected void engineInitVerify(PublicKey publicKey) throws InvalidKeyException
...
+                        RSAKeyFactory.ALLOWABLE_MODLEN_FIPS_VERIFY);
+            } else {
+                RSAKeyFactory.checkKeyLengths(((java.security.interfaces.RSAPublicKey) publicKey).getModulus().bitLength(),
+                        RSAKeyGenParameterSpec.F4,
+                        RSAKeyFactory.MIN_MODLEN_NONFIPS,
+                        64 * 1024);
+            }
+
+        } catch (InvalidKeyException e) {
+            throw new InvalidParameterException(e.getMessage());
```
jasonkatonica, line 67: Please add a caused by to this exception by passing the original exception to the newly thrown exception.

src/main/java/com/ibm/crypto/plus/provider/RSAKeyFactory.java
```diff
@@ -124,6 +128,59 @@ static void checkKeyLengths(int modulusLen, BigInteger exponent, int minModulusL
...
+     *            of the system-defined maximum and this param.
+     * @param specificModulesLen
+     *            specific module length for sign/verify.
+     *
+     * @throws InvalidKeyException
+     *             if any of the values are unacceptable.
+     */
+    static void checkKeyLengths(int modulusLen, BigInteger exponent, int minModulusLen,
+            int maxModulusLen, int[] specificModulesLen) throws InvalidKeyException {
+
```
jasonkatonica, line 154: Can this code make use of line 103 above to avoid duplicate code? Perhaps it checks all the parameters except for specificModulesLen using the method above then checks the values within `specificModulesLen`

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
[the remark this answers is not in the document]
jasonkatonica: looks good to me.

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
```diff
@@ -47,6 +49,23 @@ protected void engineInitVerify(PublicKey publicKey) throws InvalidKeyException
             throw new InvalidKeyException("Key is not an RSAPublicKey");
         }
 
+        try {
+            if (provider.isFIPS()) {
+                RSAKeyFactory.checkKeyLengths(((java.security.interfaces.RSAPublicKey) publicKey).getModulus().bitLength(),
```
jasonkatonica, line 54: Do we need to add a few tests that try some key values with the FIPS provider that must be in the list `RSAKeyFactory.ALLOWABLE_MODLEN_FIPS_VERIFY)` or else we get an exception. I didnt notice any new tests for this area along with the new `checkKeyLengths` implemented above.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
KostasTsiounis: I think it would be `TestRSASignatureVerifySmall` or something. Because it's still FIPS, but with an externally created key, right?

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
```diff
@@ -84,6 +103,23 @@ protected void engineInitSign(PrivateKey privateKey) throws InvalidKeyException
             throw new InvalidKeyException("Key is not an RSAPrivateKey");
         }
 
+        try {
+            if (provider.isFIPS()) {
```
KostasTsiounis, line 107: We could also do something like:
```
int[] specificModulesLen = null;
if (provider.isFIPS()) {
   specificModulesLen = RSAKeyFactory.ALLOWABLE_MODLEN_FIPS_SIGN;
}
RSAKeyFactory.checkKeyLengths(((java.security.interfaces.RSAPrivateKey) privateKey).getModulus().bitLength(),
                         RSAKeyGenParameterSpec.F4,
                         RSAKeyFactory.MIN_MODLEN_NONFIPS,
                         64 * 1024,
                         specificModulesLen);
```
And then in `RSAKeyFactory`, only do check if not `null`:
```
if (specificModulesLen != null && !Arrays.asList(specificModulesLen).contains(modulusLen)) {
    throw new InvalidKeyException("In FIPS mode, only 2048, 3072, or 4096 size of RSA key is accepted.");
}
```

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
JinhangZhang: Good point there, should test specific key values. Adding tests

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
```diff
@@ -0,0 +1,83 @@
...
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.Signature;
+import java.security.InvalidParameterException;
+
+import junit.framework.Test;
+import junit.framework.TestSuite;
+
+public class TestRSASignatureInteropNonFIPS 
+    extends ibm.jceplus.junit.base.BaseTestRSASignatureInterop {
```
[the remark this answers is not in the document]
johnpeck-us-ibm: I thought there were Key size tests going on.  Maybe not explicitly, but different key sizes are tested. Maybe we add a couple of boundary tests for the key size if there are not there.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
JinhangZhang: Changed the name to `TestRSASignatureWithSpecificSize`

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
JinhangZhang: this should be throw the `e` exception directly.

src/main/java/com/ibm/crypto/plus/provider/RSAKeyFactory.java
JinhangZhang: I updated the function by adding additional parameters, i think it'll be clear to use another overload function for sign and verify.

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
JinhangZhang: tests added

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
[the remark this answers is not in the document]
JinhangZhang: I updated the parent class. I dont think `BaseTestRSASignatureInterop` should be extended either.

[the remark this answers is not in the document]
JinhangZhang: @johnpeck-us-ibm you are right, I just updated the code to include more tests for various key sizes.

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
JinhangZhang: In FIPS mode, we have two lists actually, one for sign and one for verify.

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
jasonkatonica: Seems like we can get rid of the entire try catch block here as any `InvalidKeyException` can be directly thrown from this method correct?

src/main/java/com/ibm/crypto/plus/provider/RSAKeyFactory.java
jasonkatonica: It does still seem like there is duplicate code between this method and the other `checkKeyLengths` id suggest this new method implemented could call the other to check the values for `int modulusLen, BigInteger exponent, int minModulusLen,
int maxModulusLen` then add an additional check in this method for the specific modulus length check you are performing?

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureWithSpecificSize.java
```diff
@@ -0,0 +1,166 @@
...
+    public TestRSASignatureWithSpecificSize() {
+        super(Utils.TEST_SUITE_PROVIDER_NAME);
+    }
+
+    // RSA signature sign allows at least 2048 bits of RSA key to be used for sign a signature.
+    public byte[] doSign(String sigAlgo, byte[] message, PrivateKey privateKey) throws Exception {
+        Signature sign = Signature.getInstance(sigAlgo, Utils.TEST_SUITE_PROVIDER_NAME);
+        try {
+            sign.initSign(privateKey);
+        } catch (InvalidKeyException ike) {
```
jasonkatonica, line 44: The logic here and on line 66 seems to need to expect certain behavior/exceptions for certain key sizes. 

Can this be changed to expect the exception with message `In FIPS mode, only 2048, 3072, or 4096 size of RSA key is accepted.` only when testing key values other then 2048, 3072, or 4096? 

Similar for `RSA keys must be at least 2048 bits long` we only expect that failure on key lengths less then 2048.

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureInteropNonFIPS.java
JinhangZhang: updated

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
JinhangZhang: right

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
KostasTsiounis: We can change to:
```
int[] specificModulesLen = null;
int minModulusLen = RSAKeyFactory.MIN_MODLEN_NONFIPS;
if (provider.isFIPS()) {
   specificModulesLen = RSAKeyFactory.ALLOWABLE_MODLEN_FIPS_SIGN;
   minModulusLen = RSAKeyFactory.MIN_MODLEN_FIPS_PUB;
}
RSAKeyFactory.checkKeyLengths(((java.security.interfaces.RSAPrivateKey) privateKey).getModulus().bitLength(),
                         RSAKeyGenParameterSpec.F4,
                         minModulusLen,
                         64 * 1024,
                         specificModulesLen);
```

NOTICES.md
```diff
@@ -0,0 +1,42 @@
+# Notices for OpenJCEPlus Module
```
jasonkatonica, line 1: Please remove this NOTICES.md file.

JinhangZhang: It's just for building a personal jdk. will remove it after the test is done.

src/main/java/com/ibm/crypto/plus/provider/RSAKeyFactory.java
JinhangZhang: updated

src/test/java/ibm/jceplus/junit/openjceplusfips/TestRSASignatureWithSpecificSize.java
JinhangZhang: updated

jasonkatonica (approved):
LGTM

jasonkatonica:
x86_64_windows, x86_64_linux, s390x_linux, ppcle_linux

KostasTsiounis (approved):
LGTM

PR #30, opened 2024-03-08
Updates for JDK 11

jasonkatonica:
Modifications were made to support a Java 11 environment. This includes the following updates as compared to Java 17:

- The `mvn` configuration was modified to enforce Java 11 usage on this
branch of the code.
- The algortihms `Ed25519` and `Ed448` are lacking framework support in
Java 11. These algorithms and associated tests were removed.
- Construction of various key types has changed. The Java 11 frameworks
encoding and decoding methods needed to be called differently.
- The `java.nio.ByteBuffer` class needed adjustments to get/put bytes
into a specific position in the buffer. The existing method for Java 17
does not exist on Java 11.
- The `jgskit_resource.rc` file was modified appropriately for Java 11.
- The test `BaseTestXDHKeyImport.java` was modifed to expect a different
known answer value. There is a subtle difference between Java 11 and
Java 17 in terms of how the XDH DER encodings are represented. This
update describes the differences.
- Updated the github actions workflow to use Java 11.


src/test/java/ibm/jceplus/junit/base/BaseTestXDHKeyImport.java
```diff
@@ -125,18 +125,18 @@ public void createKeyPairXDHParamImport(String alg) throws Exception {
                     "515095759487624245475052955143775821008860566299846242602604658885545774489");
             scalar = Base64.getDecoder().decode("RtcJeC6VA0w1s2ly58eimwgmjNnIvdBQ0WwXodcOKx0=");
             actualPbk = Base64.getDecoder()
-                    .decode("MCowBQYDK2VuAyEAmU2lHYDsKSa7YoPkUeOxazej9sWb9m2hYbHEl6uIIwE=");
+                    .decode("MCwwBwYDK2VuBQADIQCZTaUdgOwpJrtig+RR47FrN6P2xZv2baFhscSXq4gjAQ==");
```
KostasTsiounis, line 128: Should we change all versions to this new value?

KostasTsiounis (changes requested):

README.md
```diff
@@ -234,9 +234,6 @@ KeyAgreement                | XDH                        |                 |X
 KeyFactory                  | DSA                        |X                |X             |
 KeyFactory                  | DiffieHellman              |X                |X             |
 KeyFactory                  | EC                         |X                |X             |
-KeyFactory                  | Ed25519                    |                 |X             |
```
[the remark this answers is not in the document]
jasonkatonica: I believe it to be cleaner to only have a row for algorithms that are supported in either `OpenJCEPlus` or `OpenJCEPlusFIPS`. If no support for a given algorithm then we would leave that out of the documentation entirely.

src/test/java/ibm/jceplus/junit/base/BaseTestXDHKeyImport.java
jasonkatonica: I updated the Javadoc associated with this change to document what the difference is. I will also plan on making a very similar update to other releases if this looks ok to you.

KostasTsiounis: Makes sense to me.

KostasTsiounis (approved):
LGTM

JinhangZhang (approved):

JinhangZhang:
LGTM

PR #68, opened 2024-05-28
Rename initialize method to ockInitialize

jasonkatonica:
The method name `initialize` is too generic and may cause conflicts with other libraries who have the same symbol name loaded globally.


KostasTsiounis (approved):
LGTM

PR #75, opened 2024-06-03
Change XDH double octet encoding parsing logic

jasonkatonica:
A intermittent problem has been observed as follows:

```
java.security.spec.InvalidKeySpecException: Inappropriate key specification: Failed to create XEC private key
	at openjceplus@11/com.ibm.crypto.plus.provider.XDHKeyFactory.engineGeneratePrivate(XDHKeyFactory.java:112)
	at java.base/java.security.KeyFactory.generatePrivate(KeyFactory.java:384)
	at openjceplus@11/ibm.jceplus.junit.base.BaseTestXDH.runBasicTest(BaseTestXDH.java:246)
	at openjceplus@11/ibm.jceplus.junit.base.BaseTestXDH.runBasicTests(BaseTestXDH.java:205)
	at openjceplus@11/ibm.jceplus.junit.base.BaseTestXDH.testXDH_runBasicTests(BaseTestXDH.java:60)
```

Key encodings that have been encountered for XDH based keys include:

1. [octet-string[octet-string[key-bytes]]]
2. [octet-string[key-bytes]]

Case 1 above occurs when keys are encoding from the SunEC provider on releases older then Java 17.

Case 2 above occurs when keys are generated by OpenJCEPlus or the SunEC provider on Java 17 or higher.

The encoding logic was updated to fix a corner case where a random key value contained a nested octet string just by chance. When this occured a key value was parsed that was not the correct length causing the above bad key specification when the key bytes were formatted for use with the Open Cryptography Kit C library. This update checks that key lengths are correct after parsing a nested octet string. If the length is not correct we will attempt to use a single octet format encoding.

An additional interoperability test was added to perform XDH encoding works with encoding produced by the SunEC provider for X448 and X25519 named curves.

The CurveUtil class was simplified to store known key sizes and known der encoding sizes for a given named curve.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
@@ -21,38 +21,31 @@ class CurveUtil {
         X25519, X448, FFDHE2048, FFDHE3072, FFDHE4096, FFDHE6144, FFDHE8192, Ed25519, Ed448
     }
 
-    private static HashMap<CURVE, Integer> publicCurveSizes = new HashMap<CURVE, Integer>(); // key sizes of curves in bytes
-    private static HashMap<Integer, CURVE> sizesToPublicCurves = new HashMap<Integer, CURVE>(); // reverse of above
-    private static HashMap<CURVE, Integer> privateCurveSizes = new HashMap<CURVE, Integer>(); // key sizes of curves in bytes
-    private static HashMap<Integer, CURVE> sizesToPrivateCurves = new HashMap<Integer, CURVE>(); // reverse of above
-    private static HashMap<Integer, CURVE> sizesToCurves = new HashMap<Integer, CURVE>(); // maps the total key size (I think?)
+    private static Map<CURVE, Integer> curveSizes = new HashMap<CURVE, Integer>(); // key sizes of curves in bytes
+    private static Map<CURVE, Integer> DEREncodingSizes = new HashMap<CURVE, Integer>(); // key sizes of der encoded private key values.
+    private static Map<Integer, CURVE> sizesToCurves = new HashMap<Integer, CURVE>(); // maps the total key size (I think?)
                                                                                           // to algorithm (used in constructor)
     static {
-        publicCurveSizes.put(CURVE.X25519, 32);
-        publicCurveSizes.put(CURVE.X448, 56);
-        publicCurveSizes.put(CURVE.FFDHE2048, 557);
-        publicCurveSizes.put(CURVE.FFDHE3072, 813);
-        publicCurveSizes.put(CURVE.FFDHE4096, 1069);
-        publicCurveSizes.put(CURVE.FFDHE6144, 1581);
-        publicCurveSizes.put(CURVE.FFDHE8192, 2093);
-        publicCurveSizes.put(CURVE.Ed25519, 32);
-        publicCurveSizes.put(CURVE.Ed448, 57);
 
-        for (Map.Entry<CURVE, Integer> entry : publicCurveSizes.entrySet())
-            sizesToPublicCurves.put(entry.getValue(), entry.getKey());
+        curveSizes.put(CURVE.X25519, 32);
+        curveSizes.put(CURVE.X448, 56);
+        curveSizes.put(CURVE.FFDHE2048, 256);
+        curveSizes.put(CURVE.FFDHE3072, 384);
+        curveSizes.put(CURVE.FFDHE4096, 512);
+        curveSizes.put(CURVE.FFDHE6144, 768);
+        curveSizes.put(CURVE.FFDHE8192, 1024);
+        curveSizes.put(CURVE.Ed25519, 32);
+        curveSizes.put(CURVE.Ed448, 57);
 
-        privateCurveSizes.put(CURVE.X25519, 48);
-        privateCurveSizes.put(CURVE.X448, 72);
-        privateCurveSizes.put(CURVE.FFDHE2048, 327);
-        privateCurveSizes.put(CURVE.FFDHE3072, 461);
-        privateCurveSizes.put(CURVE.FFDHE4096, 595);
-        privateCurveSizes.put(CURVE.FFDHE6144, 857);
-        privateCurveSizes.put(CURVE.FFDHE8192, 1117);
-        privateCurveSizes.put(CURVE.Ed25519, 32);
-        privateCurveSizes.put(CURVE.Ed448, 57);
-
-        for (Map.Entry<CURVE, Integer> entry : privateCurveSizes.entrySet())
-            sizesToPrivateCurves.put(entry.getValue(), entry.getKey());
+        DEREncodingSizes.put(CURVE.X25519, 48);
+        DEREncodingSizes.put(CURVE.X448, 72);
+        //DEREncodingSizes.put(CURVE.FFDHE2048, 327);
+        //DEREncodingSizes.put(CURVE.FFDHE3072, 461);
+        //DEREncodingSizes.put(CURVE.FFDHE4096, 595);
+        //DEREncodingSizes.put(CURVE.FFDHE6144, 857);
+        //DEREncodingSizes.put(CURVE.FFDHE8192, 1117);
+        DEREncodingSizes.put(CURVE.Ed25519, 32);
+        DEREncodingSizes.put(CURVE.Ed448, 57);
 
         sizesToCurves.put(255, CURVE.X25519);
         sizesToCurves.put(448, CURVE.X448);
@@ -63,23 +56,16 @@ class CurveUtil {
         sizesToCurves.put(8192, CURVE.FFDHE8192); // this is my assumption
     }
 
-    public static int getPublicCurveSize(CURVE curve) throws InvalidParameterException {
-        if (!publicCurveSizes.containsKey(curve))
+    public static int getCurveSize(CURVE curve) throws InvalidParameterException {
+        if (!curveSizes.containsKey(curve))
             throw new InvalidParameterException("Curve (" + curve + ") is not supported");
-        return publicCurveSizes.get(curve);
+        return curveSizes.get(curve);
     }
 
-    public static int getPrivateCurveSize(CURVE curve) throws InvalidParameterException {
-        if (!privateCurveSizes.containsKey(curve))
+    public static int getDEREncodingSize(CURVE curve) throws InvalidParameterException {
+        if (!DEREncodingSizes.containsKey(curve))
             throw new InvalidParameterException("Curve (" + curve + ") is not supported");
-        return privateCurveSizes.get(curve);
-    }
-
-    public static CURVE getPublicCurveOfSize(int size) throws InvalidParameterException {
-        if (!sizesToPublicCurves.containsKey(size))
-            throw new InvalidParameterException(
-                    "Public key size (" + size + ") does not correspond to a supported curve");
-        return sizesToPublicCurves.get(size);
+        return DEREncodingSizes.get(curve);
     }
 
     public static CURVE getCurveOfSize(int size) throws InvalidParameterException {
--- a/src/main/java/com/ibm/crypto/plus/provider/EdDSAKeyPairGenerator.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/EdDSAKeyPairGenerator.java
@@ -102,9 +102,9 @@ public class EdDSAKeyPairGenerator extends KeyPairGeneratorSpi {
     @Override
     public KeyPair generateKeyPair() {
         try {
-            int pub_size = CurveUtil.getPublicCurveSize(curve);
+            int keySize = CurveUtil.getCurveSize(curve);
             XECKey xecKey = XECKey.generateKeyPair(provider.getOCKContext(),
-                    this.curve.ordinal(), pub_size);
+                    this.curve.ordinal(), keySize);
             EdDSAPublicKeyImpl pubKey = new EdDSAPublicKeyImpl(provider, xecKey,
                     this.curve);
             EdDSAPrivateKeyImpl privKey = new EdDSAPrivateKeyImpl(provider, xecKey);
--- a/src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
@@ -90,15 +90,15 @@ public final class EdDSAPrivateKeyImpl extends PKCS8Key implements EdECPrivateKe
             }
 
             if (this.key == null) {
-                int pub_size = CurveUtil.getPublicCurveSize(curve);
+                int keySize = CurveUtil.getCurveSize(curve);
                 this.xecKey = XECKey.generateKeyPair(provider.getOCKContext(),
-                        this.curve.ordinal(), pub_size);
+                        this.curve.ordinal(), keySize);
             } else {
                 this.algid = CurveUtil.getAlgId(this.curve);
                 byte[] der = buildOCKPrivateKeyBytes();
-                int priv_size = CurveUtil.getPrivateCurveSize(curve);
+                int encodingSize = CurveUtil.getDEREncodingSize(curve);
                 this.xecKey = XECKey.createPrivateKey(provider.getOCKContext(), der,
-                        priv_size);
+                        encodingSize);
             }
         } catch (Exception exception) {
             InvalidParameterException ike = new InvalidParameterException(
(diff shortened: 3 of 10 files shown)


src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
```diff
@@ -21,38 +20,31 @@ public enum CURVE {
         X25519, X448, FFDHE2048, FFDHE3072, FFDHE4096, FFDHE6144, FFDHE8192, Ed25519, Ed448
     }
 
-    private static HashMap<CURVE, Integer> publicCurveSizes = new HashMap<CURVE, Integer>(); // key sizes of curves in bytes
-    private static HashMap<Integer, CURVE> sizesToPublicCurves = new HashMap<Integer, CURVE>(); // reverse of above
-    private static HashMap<CURVE, Integer> privateCurveSizes = new HashMap<CURVE, Integer>(); // key sizes of curves in bytes
-    private static HashMap<Integer, CURVE> sizesToPrivateCurves = new HashMap<Integer, CURVE>(); // reverse of above
+    private static HashMap<CURVE, Integer> curveSizes = new HashMap<CURVE, Integer>(); // key sizes of curves in bytes
```
KostasTsiounis, line 23: Should we change the definition to `Map`?

src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
```diff
@@ -21,38 +20,31 @@ public enum CURVE {
...
-        privateCurveSizes.put(CURVE.FFDHE6144, 857);
-        privateCurveSizes.put(CURVE.FFDHE8192, 1117);
-        privateCurveSizes.put(CURVE.Ed25519, 32);
-        privateCurveSizes.put(CURVE.Ed448, 57);
-
-        for (Map.Entry<CURVE, Integer> entry : privateCurveSizes.entrySet())
-            sizesToPrivateCurves.put(entry.getValue(), entry.getKey());
+        curveSizes.put(CURVE.X25519, 32);
+        curveSizes.put(CURVE.X448, 56);
+        curveSizes.put(CURVE.FFDHE2048, 557);
```
KostasTsiounis, line 31: I'm not sure we're even using those, but the numbers seem a bit off. Should we check what the expectation is?

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
```diff
@@ -131,10 +131,10 @@ public EdDSAPrivateKeyImpl(OpenJCEPlusProvider provider, byte[] encoded)
 
     void checkLength(CURVE curve) throws InvalidKeyException {
 
-        if (CurveUtil.getPrivateCurveSize(curve) != this.h.get().length) {
+        if (CurveUtil.getDEREncodingSize(curve) != this.h.get().length) {
```
KostasTsiounis, line 134: Not sure this is right here, because the private size map for `EdDSA` had the actual size of the curve and not the `DER` encoding size.

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
```diff
@@ -131,10 +131,10 @@ public EdDSAPrivateKeyImpl(OpenJCEPlusProvider provider, byte[] encoded)
...
             throw new InvalidKeyException(
                     "key length is " + this.h.get().length + ", key length must be "
-                            + CurveUtil.getPrivateCurveSize(curve));
+                            + CurveUtil.getDEREncodingSize(curve));
```
KostasTsiounis, line 137: Same as above.

src/main/java/com/ibm/crypto/plus/provider/EdDSAKeyPairGenerator.java
```diff
@@ -102,7 +102,7 @@ public void initialize(AlgorithmParameterSpec params)
     @Override
     public KeyPair generateKeyPair() {
         try {
-            int pub_size = CurveUtil.getPublicCurveSize(curve);
+            int pub_size = CurveUtil.getCurveSize(curve);
```
KostasTsiounis, line 105: Should we change the name of the variable to `keySize` or something similar, since this is not just the size of the public key?

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
```diff
@@ -90,13 +90,13 @@ public EdDSAPrivateKeyImpl(OpenJCEPlusProvider provider,
             }
 
             if (this.key == null) {
-                int pub_size = CurveUtil.getPublicCurveSize(curve);
+                int pub_size = CurveUtil.getCurveSize(curve);
```
KostasTsiounis, line 93: Should we change the name of the variable to `keySize`?

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
```diff
@@ -90,13 +90,13 @@ public EdDSAPrivateKeyImpl(OpenJCEPlusProvider provider,
...
                 this.xecKey = XECKey.generateKeyPair(provider.getOCKContext(),
                         this.curve.ordinal(), pub_size);
             } else {
                 this.algid = CurveUtil.getAlgId(this.curve);
                 byte[] der = buildOCKPrivateKeyBytes();
-                int priv_size = CurveUtil.getPrivateCurveSize(curve);
+                int priv_size = CurveUtil.getDEREncodingSize(curve);
```
KostasTsiounis, line 99: Should we change the name of the variable to `encodingSize`?

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
```diff
@@ -118,7 +118,7 @@ public EdDSAPrivateKeyImpl(OpenJCEPlusProvider provider, byte[] encoded)
             // to fit with GSKit and sets params
 
             checkLength(this.curve);
-            int priv_size = CurveUtil.getPrivateCurveSize(curve);
+            int priv_size = CurveUtil.getDEREncodingSize(curve);
```
KostasTsiounis, line 121: Should we change the name of the variable to `encodingSize`?

src/main/java/com/ibm/crypto/plus/provider/XDHKeyPairGenerator.java
```diff
@@ -119,7 +119,7 @@ private void initializeImpl(NamedParameterSpec params) {
     @Override
     public KeyPair generateKeyPair() {
         try {
-            int pub_size = CurveUtil.getPublicCurveSize(serviceCurve);
+            int pub_size = CurveUtil.getCurveSize(serviceCurve);
```
KostasTsiounis, line 122: Should we change the name of the variable to `keySize`?

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -95,7 +95,7 @@ public XDHPrivateKeyImpl(OpenJCEPlusProvider provider, byte[] encoded)
         try {
             byte[] alteredEncoded = processEncodedPrivateKey(encoded); // Sets params, key, and algid, and alters encoded
             // to fit with GSKit and sets params
-            int priv_size = CurveUtil.getPrivateCurveSize(curve);
+            int priv_size = CurveUtil.getDEREncodingSize(curve);
```
KostasTsiounis, line 98: Should we change the name of the variable to `encodingSize`?

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -141,12 +141,12 @@ public XDHPrivateKeyImpl(OpenJCEPlusProvider provider, AlgorithmParameterSpec pa
             this.key = scalar.get();
         try {
             if (this.key == null) {
-                int pub_size = CurveUtil.getPublicCurveSize(curve);
+                int pub_size = CurveUtil.getCurveSize(curve);
```
KostasTsiounis, line 144: Should we change the name of the variable to `keySize`?

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -141,12 +141,12 @@ public XDHPrivateKeyImpl(OpenJCEPlusProvider provider, AlgorithmParameterSpec pa
...
                 this.xecKey = XECKey.generateKeyPair(provider.getOCKContext(), this.curve.ordinal(), pub_size);
             } else {
                 this.algid = CurveUtil.getAlgId(this.params.getName());
                 byte[] der = buildOCKPrivateKeyBytes();
-                int priv_size = CurveUtil.getPrivateCurveSize(curve);
+                int priv_size = CurveUtil.getDEREncodingSize(curve);
```
KostasTsiounis, line 149: Should we change the name of the variable to `encodingSize`?

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -313,25 +313,60 @@ private byte[] processEncodedPrivateKey(byte[] encoded) throws IOException {
...
-            // otherwise, it causes interop issue. JCK issue 569
-            this.key = derStream.getOctetString(); // Try J17 format [octet-string[octer-string[key-bytes]]]
-        } catch (IOException e) {
-            this.key = keyBytes; // Try J11 format [octer-string[key-bytes]]
+            // XDH private key in SunEC new Java 17 design requires [octet-string[octet-string[key-bytes]]] format,
+            // otherwise, it causes interop issue.
+            if (isCorrectlyFormedOctetString(keyBytes)) {
+                this.key = derStream.getOctetString(); // We know we are working with the format [octet-string[octet-string[key-bytes]]]
+            } else {
+                throw new IOException("Octet is malformed, use entire key byte array.");
```
KostasTsiounis, line 321: Do we need to throw the exception here? We just catch it and assign the key. I think this could just be an `else` statement.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -313,25 +313,60 @@ private byte[] processEncodedPrivateKey(byte[] encoded) throws IOException {
...
+    /**
+     * Determines if a given array is a properly formed DER octet string.
+     * 
+     * @param keyBytes The byte array to check for a complete octet.
+     * @return Returns true if the first byte of the array indicates an octet ( 0x04 ) and the
+     * value continues to the end of the array indicating a complete octet filling the 
+     * given byte array. Returns false otherwise.
+     * @throws IOException Throws an IOException when a failure occurs decoding the DER encoded bytes.
+     */
+    private boolean isCorrectlyFormedOctetString(byte[] keyBytes) throws Exception {
```
KostasTsiounis, line 353: I think we should specify the exact type of exception expected.

src/main/java/com/ibm/crypto/plus/provider/XDHPublicKeyImpl.java
```diff
@@ -175,7 +175,7 @@ public XDHPublicKeyImpl(OpenJCEPlusProvider provider, AlgorithmParameterSpec par
 
         try {
             if (u == null) {
-                int pub_size = CurveUtil.getPublicCurveSize(curve);
+                int pub_size = CurveUtil.getCurveSize(curve);
```
KostasTsiounis, line 178: Should we change the name of the variable to `keySize`?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
jasonkatonica: Looking up Map it appears that it is a interface not an implementation that we could use to store these values. I kept HashMap since we were using these before and i didnt see too much of a downside there.

src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
jasonkatonica: I did research some the sizes of keys for X25519, X448, and FFDHE2048. Teh x22519 and X448 values seemed reasonable to me and were in use prior to this update. The FFDHE2048 size I also didnt change for this update. Do you see a path here that may be causing issues with the 557 size? That 557 size does seem odd to me as i would expect 2048 for the key size instead.

src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
KostasTsiounis: I meant something like `private static Map<CURVE, Integer> curveSizes = new HashMap<CURVE, Integer>();`

jasonkatonica: Id prefer to use HashMap as its more explicitly defined unless there is a reason I am not aware of?

KostasTsiounis: It's just common practice to declare the variable with the interface and use the concrete class for the initialization. But I'm good either way.

jasonkatonica: OK good point here. Will update to use `Map` in the delcarations of our `HashMap` values.

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
jasonkatonica: Given that the public and private keys are equivilant I am ok using `keySize` as the variable name here.

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
jasonkatonica: Given that this really does represent an encoding I am ok using `encodingSize` as the variable name here.

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
jasonkatonica: Yes.

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
jasonkatonica: I updated this method to use getCurveSize instead of the DER encoding sizes.

src/main/java/com/ibm/crypto/plus/provider/XDHKeyPairGenerator.java
jasonkatonica: Yes.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
jasonkatonica: Yes.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
jasonkatonica: Yes.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
jasonkatonica: Yes.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
jasonkatonica: Yes i think this is better too as we dont use exceptions for flow control which is usually a bad practice.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
jasonkatonica: Agreed will update to use IOException.

src/main/java/com/ibm/crypto/plus/provider/XDHPublicKeyImpl.java
jasonkatonica: Yes.

src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
jasonkatonica: Done

src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
jasonkatonica: I updated the keysizes to match both of our expectations here for the key size pattern.

jasonkatonica: Hi @KostasTsiounis I did comment out the `DEREncodingSizes` for FFDHE as they are not in use. I did however find that I need to keep the values for `curveSizes` associated with FFDHE curves. Please let me know if this looks ok to you.

KostasTsiounis: LGTM

KostasTsiounis (approved):
LGTM

KostasTsiounis:
Could you add the `sign-off` to satisfy the DCO check?

JinhangZhang:
LGTM

JinhangZhang (approved):

PR #77, opened 2024-06-07
Updated Initialize function name - issue 76

johnpeck-us-ibm:
Updated code to change name of initialize method. All other methods look to have unique names


johnpeck-us-ibm:
will redo

Issue #76, opened 2024-06-06
JNI library names should have less common names to avoid symbol resolution conflicts


jasonkatonica:
The JNI library in use contains various method names that are visible globally within the dll. These names should be named with a method name that has a lower chance of conflicting with other method names that may be loaded prior to our JNI. The method in particular that is known to be a problem is the `initialize` method.

PR #81, opened 2024-06-07
Updated function name for Initiialize - ISSUE 76

johnpeck-us-ibm:
Updated initialize functional to avoid collisions. Checked other native files.

Signed-off-by: John Peck <johnpeck@us.ibm.com>


KostasTsiounis:
I think we should put the `Fixes: https://github.com/IBM/OpenJCEPlus/issues/76` in the description, instead of the title of the PR.

jasonkatonica:
The `Signed-off-by` has an extra space causing the PR to fail the DCO check to fail.

PR #98, opened 2024-06-13
Update MessageDigest to implement the Clonable interface

KostasTsiounis:
The `MessageDigest`, as well as the `Digest`, class is now implementing the `Cloneable` interface, allowing it to be used by classes requiring that functionality, such as the `PKCS12` key stores, or the underlying `Hmac` ones.


src/test/java/ibm/jceplus/junit/TestMultithread.java
```diff
@@ -50,6 +50,7 @@ public class TestMultithread extends TestCase {
             "ibm.jceplus.junit.openjceplus.multithread.TestDSASignatureInteropSUN",
             "ibm.jceplus.junit.openjceplus.multithread.TestHmacMD5",
             "ibm.jceplus.junit.openjceplus.multithread.TestHmacMD5InteropSunJCE",
+            "ibm.jceplus.junit.openjceplus.multithread.TestMessageDigestClone",
```
jasonkatonica, line 53: It seems like this new test runs only SHA-256. I think we want to test cloning on all different MessageDigests that we support. To do this we could create a new `BaseTestMessageDigest` and in this class have it contain the tests you created below for cloning testing. 

For this to work a new constructor would have to be created to set the digest algorithm from parent classes then the variable would have to be used throughout all the base tests and the new  `BaseTestMessageDigest` class. This includes at least `BaseTestMD5, BaseTestSHA1, BaseTestSHA3*, BaseTestSHA224, BaseTestSHA256,BaseTestSHA284, BaseTestSHA512_*, BaseTestSHA512`

jasonkatonica (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestMessageDigestClone.java
```diff
@@ -0,0 +1,82 @@
...
+        MessageDigest md = MessageDigest.getInstance("SHA-256", providerName);
+
+        md.update(input_1);
+
+        MessageDigest mdCopy = (MessageDigest) md.clone();
+
+        byte[] digest1 = md.digest(input_2);
+        byte[] digest2 = mdCopy.digest(input_2);
+
+        assertTrue("Digest of original did not match clone's digest", Arrays.equals(digest1, digest2));
```
jasonkatonica, line 45: Instead of `assertTrue` and `assertFalse` can we use `assertArrayEquals("Digest of original did not match clone's digest",digest1, digest2)` when possible? Would prefer these as a habit as it gives more information on failure and clues as to how arrays are different.

src/test/java/ibm/jceplus/junit/TestMultithread.java
KostasTsiounis: I added cloning tests for all algorithms with [this](https://github.com/IBM/OpenJCEPlus/pull/98/commits/28c49359b572fdada72a1bfbefd6961500b2b248) commit.

src/test/java/ibm/jceplus/junit/base/BaseTestMessageDigestClone.java
KostasTsiounis: I substituted `assertTrue` with suggestion. I couldn't find anything similar for the `assertFalse` cases.

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
```diff
@@ -343,4 +337,20 @@ protected static boolean validId(long id) {
...
+        
+        copy.contextFromQueue = false;
+        try {
+            copy.digestId = NativeInterface.DIGEST_copy(
+                this.ockContext.getId(), getId());
+            if (copy.digestId == 0) {
+                throw new CloneNotSupportedException("Copy of native digest context failed.");
+            }
+        } catch (OCKException e) {
+            throw new CloneNotSupportedException(e.getMessage());
```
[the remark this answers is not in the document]
KostasTsiounis: The `CloneNotSupportedException` only accepts a `String` parameter. I could do `e.getStackTrace()` and use a `StringBuilder` to turn the array returned into a single `String` to pass. Would that make sense?

src/main/java/com/ibm/crypto/plus/provider/MessageDigest.java
```diff
@@ -9,9 +9,10 @@
 package com.ibm.crypto.plus.provider;
 
 import java.security.MessageDigestSpi;
+
```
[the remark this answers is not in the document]
KostasTsiounis: Removed.

src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
```diff
@@ -15,6 +15,7 @@
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.security.ProviderException;
+
```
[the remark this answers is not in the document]
KostasTsiounis: Removed.

src/main/native/Digest.c
```diff
@@ -142,6 +142,90 @@ JNIEXPORT jlong JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_DI
...
+    gslogFunctionEntry(functionName);
+  }
+  if (ockDigest == NULL) {
+    if( debug ) {
+      gslogFunctionExit(functionName);
+    }
+    return 0;
+  }
+#ifdef DEBUG_DIGEST_DETAIL
+  if ( debug ) {
```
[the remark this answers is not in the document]
KostasTsiounis: I tried it and it seems to compile and run fine.

src/main/native/Digest.c
```diff
@@ -142,6 +142,90 @@ JNIEXPORT jlong JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_DI
...
+JNIEXPORT jlong JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_DIGEST_1copy
+  (JNIEnv *env, jclass thisObj, jlong ockContextId, jlong digestId)
+{
+  static const char * functionName = "NativeInterface.DIGEST_copy";
+
+  ICC_CTX *   ockCtx = (ICC_CTX *)((intptr_t) ockContextId);
+  OCKDigest * ockDigest = (OCKDigest *)((intptr_t) digestId);
+  OCKDigest * ockDigestCopy = (OCKDigest *)malloc(sizeof(OCKDigest));
+  jlong       digestCopyId = 0;
+  int         rc = ICC_OSSL_SUCCESS;
```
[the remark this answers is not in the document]
KostasTsiounis: I was only using it in one place, so I removed it altogether and just checked the result directly.

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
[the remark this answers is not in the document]
jasonkatonica: OK i see. Yes i think converting the stack trace into a string would be a good idea so that information can be found. Reading through this code it seems like it would be hard for this exception to be thrown but better safe then sorry in this spot.

src/main/native/Digest.c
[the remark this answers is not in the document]
jasonkatonica: OK thanks.

src/test/java/ibm/jceplus/junit/base/BaseTestMessageDigestClone.java
jasonkatonica: I couldnt find the reverse of `assertArrayEquals` either so using `assertFalse` seems ok.

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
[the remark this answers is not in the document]
jasonkatonica: LGTM

jasonkatonica:
Please fix the compilation error as seen in the github action which is failing:
```
 Error:  /home/runner/work/OpenJCEPlus/OpenJCEPlus/src/test/java/ibm/jceplus/junit/openjceplus/TestAll.java:[40,62] cannot find symbol
  symbol: class TestMD5Clone
  ```

**Comment by jasonkatonica** [comment]
Also please rebase into a single commit, or add DCO to each of the above commits.

**Comment by KostasTsiounis** [comment]
Fixed and squashed.

File: src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java [hunk 8]
```diff
@@ -92,10 +88,8 @@ void getContext() throws OCKException {
             }
         }
 
-        if (this.nativeResources != null) {
+        if (this.digestId != 0) {
```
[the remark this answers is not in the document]
johnpeck-us-ibm: Not sure why a class was used here. Talked with Karthik and simple field is much better, since the class causes memory to be held on too and can cause an out of memory issue.

jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
```diff
@@ -343,4 +339,24 @@ protected static boolean validId(long id) {
         //OCKDebug.Msg(debPrefix, methodName,  "Id : " + id);
         return (id != 0L);
     }
+
```
JinhangZhang, line 342: Perhaps we can add a override Annotation there.

KostasTsiounis: Done. I, also, added the annotation to `MessageDigest`.

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
```diff
@@ -48,10 +50,6 @@ public final class Digest {
 
     static private int runtimeContextNum[];
 
-    class NativeResources {
-        private long digestId = 0;
-    }
-
     class ConcurrentLinkedQueueLong extends ConcurrentLinkedQueue<Long> {
```
JinhangZhang, line 53: ConcurrentLinkedQueueLong contexts[] are also a partial Digest, though we should implements cloneable for this inner class as well?

KostasTsiounis: We don't need to clone that, because it is static and therefore not instance specific.

JinhangZhang: right, miss the static

src/main/native/Digest.c
```diff
@@ -142,6 +142,88 @@ JNIEXPORT jlong JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_DI
...
+#endif
+    throwOCKException(env, 0, "Error allocating copy of OCKDigest");
+    if( debug ) {
+      gslogFunctionExit(functionName);
+    }
+    return 0;
+  } else {
+    ockDigestCopy->md = ockDigest->md;
+    ockDigestCopy->mdCtx = ICC_EVP_MD_CTX_new(ockCtx);
+    if ( NULL == ockDigest->mdCtx ) {
```
JinhangZhang, line 189: why we check `ockDigest->mdCtx` not `ockDigestCopy-mdCtx`?

taoliult (approved):
LGTM

src/main/native/Digest.c
KostasTsiounis: Good catch. Fixed it.

JinhangZhang (approved):
LGTM

JinhangZhang (approved):

jasonkatonica:
ppcle linux, s390x linux, x86 linux, x86 windows

PR #108, opened 2024-06-28
Address Java compilation warnings

KostasTsiounis:
Warnings produced during compilation of both source code and tests are addressed. Those warnings pertain to:
- Redundant casts
- Deprecated APIs that are substituted with the updated alternatives
- Incorrect visibility of classes and methods
- Use of internal proprietary APIs

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -221,6 +221,7 @@
                     <source>${jdk.build.target}</source>   
                     <target>${jdk.build.target}</target>
                     <compilerArgs>
+                        <arg>-XDignore.symbol.file</arg>
                         <arg>-Xlint:all</arg> 
                         <arg>--add-exports </arg> 
                         <arg>java.base/sun.security.internal.spec=openjceplus</arg>
@@ -237,6 +238,7 @@
                         <arg>--add-exports </arg>
                         <arg>java.base/jdk.internal.logger=openjceplus</arg>
                     </compilerArgs>
+                    <fork>true</fork>
                     <debug>true</debug>  
                 </configuration>
             </plugin>
@@ -498,6 +500,16 @@
             <artifactId>bcprov-jdk18on</artifactId>
             <version>1.78.1</version>
         </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk18on</artifactId>
+            <version>1.78.1</version>
+        </dependency>
+        <dependency>
+          <groupId>org.bouncycastle</groupId>
+          <artifactId>bcprov-ext-jdk18on</artifactId>
+          <version>1.78.1</version>
+      </dependency>
         <dependency>
             <groupId>org.jacoco</groupId>
             <artifactId>jacoco-maven-plugin</artifactId>
--- a/src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2023, 2024
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -103,7 +103,7 @@ public final class AESCCMCipher extends CipherSpi implements AESConstants, CCMCo
 
     public AESCCMCipher(OpenJCEPlusProvider provider) {
 
-        if (!provider.verifySelfIntegrity(this.getClass())) {
+        if (!OpenJCEPlusProvider.verifySelfIntegrity(this)) {
             throw new SecurityException("Integrity check failed for: " + provider.getName());
         }
 
@@ -569,7 +569,7 @@ public final class AESCCMCipher extends CipherSpi implements AESConstants, CCMCo
         if (params != null) {
             CCMParameterSpec ivSpec = null;
             try {
-                ivSpec = (CCMParameterSpec) params.getParameterSpec(CCMParameterSpec.class);
+                ivSpec = params.getParameterSpec(CCMParameterSpec.class);
             } catch (InvalidParameterSpecException ipse) {
                 throw new InvalidAlgorithmParameterException(
                         "Wrong parameter " + "type: CCM " + "expected");
--- a/src/main/java/com/ibm/crypto/plus/provider/AESKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESKey.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2023, 2024
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -13,12 +13,11 @@ import java.security.KeyRep;
 import java.util.Arrays;
 import javax.crypto.SecretKey;
 import javax.security.auth.DestroyFailedException;
-import javax.security.auth.Destroyable;
 
 /**
  * This class represents an AES key.
  */
-final class AESKey implements SecretKey, Destroyable {
+final class AESKey implements SecretKey {
 
     static final long serialVersionUID = -8899864838936117258L;
 
@@ -35,7 +34,7 @@ final class AESKey implements SecretKey, Destroyable {
      * @exception InvalidKeyException
      *                if the given key has wrong size
      */
-    public AESKey(byte[] key) throws InvalidKeyException {
+    AESKey(byte[] key) throws InvalidKeyException {
         if ((key == null) || !AESUtils.isKeySizeValid(key.length)) {
             throw new InvalidKeyException("Wrong key size");
         }
@@ -62,7 +61,7 @@ final class AESKey implements SecretKey, Destroyable {
 
         // Return a copy of the key, rather than a reference,
         // so that the key data cannot be modified from outside
-        return (byte[]) this.key.clone();
+        return this.key.clone();
     }
 
     /**
--- a/src/main/java/com/ibm/crypto/plus/provider/ChaCha20Parameters.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ChaCha20Parameters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2023, 2024
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -19,13 +19,11 @@ import javax.crypto.spec.ChaCha20ParameterSpec;
  */
 public final class ChaCha20Parameters extends AlgorithmParametersSpi implements ChaCha20Constants {
 
-    private OpenJCEPlusProvider provider = null;
-
     private byte[] nonce;
     int counter;
 
-    public ChaCha20Parameters(OpenJCEPlusProvider provider) {
-        this.provider = provider;
+    public ChaCha20Parameters() {
+        super();
     }
 
     @Override
@@ -39,7 +37,7 @@ public final class ChaCha20Parameters extends AlgorithmParametersSpi implements 
             throw new InvalidParameterSpecException(
                     "Nonce not " + ChaCha20_NONCE_SIZE + " bytes long");
         }
-        this.nonce = (byte[]) nonce.clone();
+        this.nonce = nonce.clone();
 
         this.counter = ((ChaCha20ParameterSpec) paramSpec).getCounter();
     }
--- a/src/main/java/com/ibm/crypto/plus/provider/DHParameters.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/DHParameters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2023, 2024
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -21,7 +21,8 @@ import sun.security.util.DerValue;
  * This class implements the parameter set used by the Diffie-Hellman key
  * agreement as defined in the PKCS #3 standard.
  */
-public final class DHParameters extends AlgorithmParametersSpi {
+public final class DHParameters extends AlgorithmParametersSpi implements java.io.Serializable {
+    private static final long serialVersionUID = 7137508373627164657L;
 
     private OpenJCEPlusProvider provider;
 
--- a/src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2023, 2024
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
  * this file except in compliance with the License.  You can obtain a copy
@@ -9,7 +9,6 @@
 package com.ibm.crypto.plus.provider;
 
 import java.io.IOException;
-import java.io.Serializable;
 import java.math.BigInteger;
 import java.security.AlgorithmParameters;
 import java.security.InvalidKeyException;
@@ -18,7 +17,6 @@ import java.security.spec.ECParameterSpec;
 import java.security.spec.InvalidParameterSpecException;
 import java.util.Arrays;
 import javax.security.auth.DestroyFailedException;
-import javax.security.auth.Destroyable;
 import com.ibm.crypto.plus.provider.ock.ECKey;
 
 import sun.security.pkcs.PKCS8Key;
@@ -28,8 +26,7 @@ import sun.security.util.DerValue;
 import sun.security.util.ObjectIdentifier;
 import sun.security.x509.AlgorithmId;
 
-final class ECPrivateKey extends PKCS8Key implements java.security.interfaces.ECPrivateKey,
-        java.security.interfaces.ECKey, Serializable, Destroyable {
+final class ECPrivateKey extends PKCS8Key implements java.security.interfaces.ECPrivateKey {
 
     /**
      * 
@@ -39,7 +36,7 @@ final class ECPrivateKey extends PKCS8Key implements java.security.interfaces.EC
 
     private OpenJCEPlusProvider provider = null;
     private BigInteger s;
-    private ECParameterSpec params;
+    private transient ECParameterSpec params;
 
     private ECPublicKey publicKey = null;
     private byte[] privateKeyBytesEncoded = null;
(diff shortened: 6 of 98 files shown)


src/main/java/com/ibm/crypto/plus/provider/DHParameters.java
```diff
@@ -21,7 +21,8 @@
  * This class implements the parameter set used by the Diffie-Hellman key
  * agreement as defined in the PKCS #3 standard.
  */
-public final class DHParameters extends AlgorithmParametersSpi {
+public final class DHParameters extends AlgorithmParametersSpi implements java.io.Serializable {
+    private static final long serialVersionUID = 1L;
```
jasonkatonica, line 25: Not positive this is safe to do? Does this class remain serializable? This comment also applies to a few other locations where the class no longer extends serializable.

src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
```diff
@@ -39,7 +36,7 @@ final class ECPrivateKey extends PKCS8Key implements java.security.interfaces.EC
 
     private OpenJCEPlusProvider provider = null;
     private BigInteger s;
-    private ECParameterSpec params;
+    private transient ECParameterSpec params;
```
jasonkatonica, line 39: Does this require the variable to be `transient` because a ECParameterSpec instance cannot be serialzed? If so is there any change in behavior here in terms of a serializable ECPrivateKey instance? I have a similar question in other classes where `transient` was added to various fields.

src/main/java/com/ibm/crypto/plus/provider/PSSParameters.java
```diff
@@ -68,13 +69,19 @@
...
-    protected int trailerField;
+    private AlgorithmId hashAlgorithm;
+    private AlgorithmId maskGenAlgorithm;
+    private AlgorithmParameterSpec mgfParameterSpec;
+    private int saltLength;
+    private int trailerField;
     private PSSParameterSpec spec;
 
+    private static final PSSParameterSpec DEFAULT_SPEC = new PSSParameterSpec("SHA-1",
+                                                                      "MGF1",
```
jasonkatonica, line 80: Indentation seems off here.

src/main/java/com/ibm/crypto/plus/provider/ock/CCMCipher.java
```diff
@@ -37,12 +36,8 @@ public final class CCMCipher {
...
 
     static {
-        disableCCMAcceleration = AccessController.doPrivileged(new PrivilegedAction<Boolean>() {
-            public Boolean run() {
-                return ("true"
-                        .equalsIgnoreCase(System.getProperty(DISABLE_CCM_ACCELERATION, "false")));
-            }
-        });
+        disableCCMAcceleration = Boolean.parseBoolean(
+            System.getProperty("jdk.nativeCryptoTrace", "false"));
```
jasonkatonica, line 40: Why change the variable name here? Seems you updated this to match what is in openjdk for openssl support within Semeru?

src/main/java/com/ibm/misc/Debug.java
```diff
@@ -194,7 +195,7 @@ public Object run() {
      *   <code>TYPE_ENTRY_EXIT</code> has been split into
      *   <code>TYPE_ENTRY</code> and <code>TYPE_EXIT</code>.
      */
-
+    @Deprecated
```
jasonkatonica, line 198: Why deprecate this field?

src/test/java/ibm/jceplus/junit/base/BaseTestAESGCM_IntIV.java
```diff
@@ -66,46 +63,11 @@ protected void setUp() throws Exception {
 
         key = aesKeyGen.generateKey();
```
jasonkatonica, line 65: Why remove classGCMParameterSpec and ctorGCMParameterSpec and the code setting those values. Are these fields unused?

src/test/java/ibm/jceplus/junit/base/BaseTestAESGCM_IntIV.java
```diff
@@ -66,46 +63,11 @@ protected void setUp() throws Exception {
...
-            }
-        }
-
         byte[] iv = new byte[16];// com.ibm.crypto.plus.provider.AESConstants.AES_BLOCK_SIZE];
         SecureRandom rnd = new java.security.SecureRandom();
         rnd.nextBytes(iv);
 
-        aParamSpec = (AlgorithmParameterSpec) ctorGCMParameterSpec.newInstance(DEFAULT_TAG_LENGTH,
-                iv);
+        aParamSpec = new javax.crypto.spec.GCMParameterSpec(DEFAULT_TAG_LENGTH,iv);
```
jasonkatonica, line 70: Space after `,`

src/test/java/ibm/jceplus/junit/base/BaseTestEdDSASignatureInterop.java
```diff
@@ -81,8 +86,15 @@ private byte[] doSign(String sigAlgo, byte[] message, PrivateKey privateKey) thr
     //Verify the message with BouncyCastle provided Ed25519
     protected void doVerifyEd25519(byte[] message, byte[] signedBytes, PublicKey publicKey)
             throws Exception {
```
jasonkatonica, line 88: This code block along with others in this file look quite different. Could you explain where this code came from?

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/ChaCha20Parameters.java
```diff
@@ -19,13 +19,11 @@
...
 public final class ChaCha20Parameters extends AlgorithmParametersSpi implements ChaCha20Constants {
 
-    private OpenJCEPlusProvider provider = null;
-
     private byte[] nonce;
     int counter;
 
-    public ChaCha20Parameters(OpenJCEPlusProvider provider) {
-        this.provider = provider;
+    public ChaCha20Parameters() {
```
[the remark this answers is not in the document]
KostasTsiounis: I checked and the provider is not used anywhere in this or the other files from which I removed it. I can revert it if you feel uncomfortable with the signature change, but it's just an unused field.

src/main/java/com/ibm/crypto/plus/provider/DHParameters.java
KostasTsiounis: Not sure what you mean here. The class is serializable as long as all of its fields are serializable. Alternatively, we can add a `SuppressWarnings(serial)` or instruct the compiler to ignore such warnings.

src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
KostasTsiounis: Exactly. There are various other fields that are declared as `transient`. I'm not sure that we even ever serialize any of those instances. It's more of a requirement coming from the superclass. Even in the JDK code the serialization warnings are suppressed (https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/0424dfc7c8435f37d18bd32498e5e12fd2966044/src/java.base/share/classes/java/security/interfaces/ECPrivateKey.java#L50).

KostasTsiounis: Same as above. We could switch to `@SuppressWarnings(serial)`.

src/test/java/ibm/jceplus/junit/base/BaseTestAESGCM_IntIV.java
KostasTsiounis: These fields were used because the code accounted for Java <8 versions and needed to instantiate the proper class. But for our purposes, we know exactly what class and constructor we want and can use them directly.

src/main/java/com/ibm/misc/Debug.java
KostasTsiounis: If you check the comment above, this field is supposed to be deprecated. It just seems like we never declared it.

src/test/java/ibm/jceplus/junit/base/BaseTestEdDSASignatureInterop.java
KostasTsiounis: The `com.ibm.crypto.plus.provider.EdDSAPublicKeyImpl` was previously used, which shouldn't be public. So, I the `EdDSAKeyFactory` to get the spec, which is allowed, and then used the inverse process used in the constructor of the `EdDSAPublicKeyImpl` to extract the information required by `BC`'s `Ed25519PublicKeyParameters`, without exposing any unnecessary classes or methods.

src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java
```diff
@@ -103,7 +103,7 @@ public final class AESCCMCipher extends CipherSpi implements AESConstants, CCMCo
```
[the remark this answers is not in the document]
KostasTsiounis: Updated.

src/main/java/com/ibm/crypto/plus/provider/PSSParameters.java
KostasTsiounis: Fixed.

src/main/java/com/ibm/crypto/plus/provider/ock/CCMCipher.java
KostasTsiounis: My bad. Got the code to mirror what we do in Semeru, but forgot to fix the flag. Thanks for catching that.

src/test/java/ibm/jceplus/junit/base/BaseTestAESGCM_IntIV.java
KostasTsiounis: Fixed.

src/main/java/com/ibm/crypto/plus/provider/ChaCha20Parameters.java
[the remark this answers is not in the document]
jasonkatonica: OK if its not in use I am ok removing it.

src/main/java/com/ibm/crypto/plus/provider/DHParameters.java
jasonkatonica: My question here is why did we make this class serializable? I assume we have a reason. If so then we should use a random serialVersionUID produced by something like `serialver` instead of using the value `1L`.

src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
jasonkatonica: I guess i dont see a behavior change here so i am ok with the code as it is.

src/main/java/com/ibm/misc/Debug.java
jasonkatonica: OK i see that now.

src/main/java/com/ibm/crypto/plus/provider/ock/CCMCipher.java
jasonkatonica: Looks good now.

src/main/java/com/ibm/crypto/plus/provider/DHParameters.java
KostasTsiounis: I changed this class, because it is used by `DHPrivateKey` and `DHPublicKey`, which are serializable. The option there were:
- Make the parameters serializable too
- Specify the parameters as `transient` in the keys
- Ignore (i.e. suppress) the serialization warnings

I thought that was the best option, but we can always go for one of the other two.

KostasTsiounis: I updated the `serialVersionUID` using `serialver` as suggested.

jasonkatonica: Updating the `serialVersionUID` looks good to me. This approach to adding Serializable to the class is as good as any of the other approaches you outlined above.

jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/AESKey.java
```diff
@@ -35,7 +34,7 @@ final class AESKey implements SecretKey, Destroyable {
      * @exception InvalidKeyException
      *                if the given key has wrong size
      */
-    public AESKey(byte[] key) throws InvalidKeyException {
+    AESKey(byte[] key) throws InvalidKeyException {
```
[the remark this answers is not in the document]
KostasTsiounis: We removed it, because the constructor shouldn't be called by external classes. I updated those visibilities based on the warnings I was getting, so there might be other visibilities that need to change. I'm not sure whether we should update them all as part of this PR, but let me know which ones you've seen @taoliult. What do you think @jasonkatonica?

src/main/java/com/ibm/crypto/plus/provider/RSAKeyFactory.java
```diff
@@ -188,7 +179,7 @@ public RSAKeyFactory(OpenJCEPlusProvider provider) {
         this.type = KeyType.RSA;
     }
 
-    public RSAKeyFactory(OpenJCEPlusProvider provider, KeyType type) {
+    private RSAKeyFactory(OpenJCEPlusProvider provider, KeyType type) {
```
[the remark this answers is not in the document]
KostasTsiounis: It's used in line 73 by the static `toRSAKey()`.

src/main/java/com/ibm/crypto/plus/provider/AESKey.java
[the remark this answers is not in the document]
taoliult: You mean, all the classes constructors in the `com/ibm/crypto/plus/provider/ `package (Except the OpenJCEPlus/OpenJCEPlusFIPS provider class) shouldn't be called by external classes?

For example, `AESCipher` class, its constructor is still public. And the `EdDSAKeyFactory` class, its constructor changed from `protected` to `public`.

I am totally ok with not update them all in this PR, and if all the classes constructors in the `com/ibm/crypto/plus/provider/ `package shouldn't be called by external classes, then maybe we can have another PR to update them for the consistent?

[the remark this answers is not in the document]
KostasTsiounis: Hmm, I don't remember why I changed the constructor for `EdDSAKeyFactory`. Maybe they should be `protected`. Let me check it out.  

As far as the `AESCipher`, it didn't cause any warnings, so I didn't touch, and there might be other similar issues in the code actually.

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
```diff
@@ -51,7 +51,7 @@ public final class Digest implements Cloneable {
     static private int runtimeContextNum[];
 
     class ConcurrentLinkedQueueLong extends ConcurrentLinkedQueue<Long> {
-        /* empty */
+        private static final long serialVersionUID = 196745693267521676L;
```
JinhangZhang, line 54: Does it necessnary to setup a serialVersionUID for Digest class? I am not sure if Digest object will be serialized and deserialized.

pom.xml
```diff
@@ -230,6 +231,7 @@
                         <arg>--add-exports </arg>
                         <arg>java.base/jdk.internal.logger=openjceplus</arg>
                     </compilerArgs>
+                    <fork>true</fork>
```
JinhangZhang, line 234: Is there any reason why we need to run this compiler plugin in another JVM process?

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
KostasTsiounis: This is not the `Digest` that is serializable, but the inner class. It inherits it from the `ConcurrentLinkedQueue`. We maybe could eventually remove this inner class, but it's not as straightforward.

pom.xml
JinhangZhang: If we want to let some plugins to run in parallel, perhaps we can add one for `maven-surefire-plugin`.

KostasTsiounis: It doesn't have to do with running things in parallel.

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
JinhangZhang: Right, its come from the ConcurrentLinkedQueue.

src/main/java/com/ibm/crypto/plus/provider/AESKey.java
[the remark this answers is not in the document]
KostasTsiounis: I updated the visibility of the constructors of `EdDSAKeyFactory`. Should be fine now.

JinhangZhang:
LGTM

JinhangZhang (approved):

KostasTsiounis:
A few more warnings appeared on the tests, so I addressed them in an additional commit. Mind checking them @jasonkatonica, @taoliult, @JinhangZhang ?

src/test/java/ibm/jceplus/junit/base/BaseTestRSAPSSInterop.java
```diff
@@ -482,7 +482,8 @@ protected void dotestSignature(byte[] content, String algorithmA, String algorit
             algParams.getParameterSpec(PSSParameterSpec.class);
             //System.out.println("parameters=" + algParams.toString());
         } else if (providerA.equals("SunRsaSign")) {
-            sig.setParameter(PSSParameterSpec.DEFAULT);
+            sig.setParameter(new PSSParameterSpec("SHA-1", "MGF1",
```
taoliult, line 485: The PSSParameterSpec.DEFAULT will new a PSSParameterSpec instance which has the below default values:
```
message digest  -- "SHA-1"
mask generation function (mgf) -- "MGF1"
parameters for mgf -- MGF1ParameterSpec.SHA1
SaltLength   -- 20
TrailerField -- 1
```
Why we need to call its constructor instead of using `PSSParameterSpec.DEFAULT` ?

taoliult (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestRSAPSSInterop.java
KostasTsiounis: `PSSParameterSpec.DEFAULT` has been deprecated starting on Java20.

taoliult (approved):
LGTM

jasonkatonica:
> A few more warnings appeared on the tests, so I addressed them in an additional commit. Mind checking them @jasonkatonica, @taoliult, @JinhangZhang ?

These additional updates read ok to me.

JinhangZhang:
> A few more warnings appeared on the tests, so I addressed them in an additional commit. Mind checking them @jasonkatonica, @taoliult, @JinhangZhang ?

LGTM

jasonkatonica:

PR #121, opened 2024-07-03
Update project dependencies

jasonkatonica:
All the project dependencies were updated to the latest versions.

Plugins not yet declared in the dependency section were added. This
allows for easier resolution of dependencies as a whole if the user
desires to pre-download all dependencies needed by the project.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.build.timestamp.format>yyyyMMdd</maven.build.timestamp.format>
-        <jupiter.version>5.10.2</jupiter.version>
+        <jupiter.version>5.10.3</jupiter.version>
     </properties>
         <profiles>
           <profile>
@@ -121,7 +121,7 @@
             <plugin>
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-enforcer-plugin</artifactId>
-              <version>3.3.0</version>
+              <version>3.5.0</version>
               <executions>
                 <execution>
                   <id>enforce-versions</id>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -458,24 +476,33 @@
...
-            <version>3.3.0</version>
+            <version>3.4.2</version>
             <type>maven-plugin</type>
             <exclusions>
                 <exclusion>
                     <groupId>org.codehaus.plexus</groupId>
                     <artifactId>plexus-archiver</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.iq80.snappy</groupId>
```
KostasTsiounis, line 492: I'm guessing this has a CVE and we need a more updated version, right?

pom.xml
```diff
@@ -399,14 +392,39 @@
...
-                <version>5.10.2</version>
+                <version>5.10.3</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
         </dependencies>
     </dependencyManagement>
 
     <dependencies>
+        <dependency>
```
KostasTsiounis, line 403: How come we didn't need to explicitly add the dependencies for those in the past, but we do now?

pom.xml
jasonkatonica: Yes correct.

pom.xml
jasonkatonica: This is not necessarily a requirement. This allows us to declare dependencies up front such that users can pre-download all dependencies in our pom file. I am ok reverting these changes if we think there is a downside to doing this.

KostasTsiounis: Not an issue. Just wondering.

KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

PR #127, opened 2024-07-05
Exclude multithread RSA-related tests and increase thread init timeout

KostasTsiounis:
The multithreaded versions of `RSAPSS` and `RSA_2048` seem to be hanging, so they are temporarily omitted until a fix is available.

In some platforms, like Windows, the initialization of the threads requires more time, so the timeout is increased to account for that.


jasonkatonica (approved):
LGTM

Issue #100, opened 2024-06-13
Key encodings for X25519 and X448 do not match encoding formats of Oracle providers


jasonkatonica:
The test `BaseTestXDHKeyImport` once properly testing specific providers ( OpenJCEPlus and OpenJCEPlusFIPS ) does not match encodings by other providers.

The code should be updated such that encodings created by the OpenJCEPlus and OpenJCEPlusFIPS provider create keys similar to the SunEC provider in each release. There are different formats of such keys explained in a comment here:

https://github.com/IBM/OpenJCEPlus/blob/3c44149386d98d4834da3409744443918d31239a/src/test/java/ibm/jceplus/junit/base/BaseTestXDHKeyImport.java#L159

Our code does not appear to match what Sun is doing and the test should enforce this by producing encodings with Sun and our provider and ensuring not only that they match the expected value but match each other in format.

johnpeck-us-ibm:
We need to make sure that is works with BouncyCastle as well. Since this is where these were tested originally.

jasonkatonica:
Yes that makes sense we investigate that all providers oracle and bouncy castle match, and if not comment this well and decide which encodings we should use.

PR #136, opened 2024-07-11
Fix the encoding formats for X25519 and X448

taoliult:
This PR fixes the XDH Private Key and Public Key encoding formats issues.

According to the PKCS#8 Private-Key Specification, the new format privateKey is an octet string whose contents are the value of the private key. So, adding the octet string before the private key when passing the private key object to its parent class PKCS8Key key object for 17 and after version.

According to Sun old versions, 11 and before, the new XDH format is not supported. So, adding a DER "null" value on the OID sequence only for 11 and before versions.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
@@ -43,6 +43,7 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
     private transient Optional<byte[]> scalar;
     private transient NamedParameterSpec params;
     private CURVE curve;
+    private byte[] k; // The raw key bytes, without OctetString or DER encoded
     BigInteger bi1; // parameter used in FFDHE
     BigInteger bi2; // parameter used in FFDHE
     BigInteger bi3; // parameter used in FFDHE
@@ -55,9 +56,10 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
     private transient XECKey xecKey = null;
 
     private void setFieldsFromXeckey() throws Exception {
-        if (this.key == null) {
-            this.key = extractPrivateKeyFromOCK(xecKey.getPrivateKeyBytes()); // Extract key from GSKit and sets params
-            this.scalar = Optional.of(key);
+        if (k == null) {
+            k = extractPrivateKeyFromOCK(xecKey.getPrivateKeyBytes()); // Extract key from GSKit and sets params
+            setPKCS8KeyByte(k);
+            this.scalar = Optional.of(k);
             this.algid = CurveUtil.getAlgId(this.params.getName());
         }
     }
@@ -97,7 +99,7 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
             // to fit with GSKit and sets params
             int encodingSize = CurveUtil.getDEREncodingSize(curve);
             this.xecKey = XECKey.createPrivateKey(provider.getOCKContext(), alteredEncoded, encodingSize);
-            this.scalar = Optional.of(this.key);
+            this.scalar = Optional.of(k);
         } catch (Exception exception) {
             InvalidKeyException ike = new InvalidKeyException("Failed to create XEC private key");
             provider.setOCKExceptionCause(ike, exception);
@@ -138,9 +140,9 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
         this.provider = provider;
         this.scalar = scalar;
         if (scalar != null)
-            this.key = scalar.get();
+            k = scalar.get();
         try {
-            if (this.key == null) {
+            if (k == null) {
                 int keySize = CurveUtil.getCurveSize(curve);
                 this.xecKey = XECKey.generateKeyPair(provider.getOCKContext(), this.curve.ordinal(), keySize);
             } else {
@@ -149,6 +151,7 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
                 int encodingSize = CurveUtil.getDEREncodingSize(curve);
                 this.xecKey = XECKey.createPrivateKey(provider.getOCKContext(), der, encodingSize);
             }
+            setPKCS8KeyByte(k);
         } catch (Exception exception) {
             InvalidParameterException ike = new InvalidParameterException(
                     "Failed to create XEC private key");
@@ -178,7 +181,7 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
 
         // Adding Key
         DerOutputStream keyOctetString = new DerOutputStream();
-        keyOctetString.putOctetString(key);
+        keyOctetString.putOctetString(k);
         mainSeq.putOctetString(keyOctetString.toByteArray());
 
         // Wrapping up in a sequence
@@ -316,20 +319,21 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
             // XDH private key in SunEC new Java 17 design requires [octet-string[octet-string[key-bytes]]] format,
             // otherwise, it causes interop issue.
             if (isCorrectlyFormedOctetString(keyBytes)) {
-                this.key = derStream.getOctetString(); // We know we are working with the format [octet-string[octet-string[key-bytes]]]
+                k = derStream.getOctetString(); // We know we are working with the format [octet-string[octet-string[key-bytes]]]
             } else {
-                this.key = keyBytes; // Try J11 format [octet-string[key-bytes]]
+                k = keyBytes; // Try J11 format [octet-string[key-bytes]]
             }
         } catch (Exception e) {
             //e.printStackTrace();
-            this.key = keyBytes; // Try J11 format [octet-string[key-bytes]]
+            k = keyBytes; // Try J11 format [octet-string[key-bytes]]
         }
+        setPKCS8KeyByte(k);
         try (DerOutputStream encodedKey = new DerOutputStream()) {
             if (CurveUtil.isFFDHE(this.curve)) {
-                BigInteger octetStringAsBigInt = new BigInteger(this.key);
+                BigInteger octetStringAsBigInt = new BigInteger(k);
                 encodedKey.putInteger(octetStringAsBigInt); // Put in another octet string
             } else {
-                encodedKey.putOctetString(this.key); // Put in another octet string
+                encodedKey.putOctetString(k); // Put in another octet string
             }
             outStream.putOctetString(encodedKey.toByteArray());
         }
@@ -399,7 +403,7 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
         } catch (Exception exception) {
             this.exception = exception;
         }
-        return this.key.clone();
+        return k.clone();
     }
 
     @Override
@@ -491,10 +495,10 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
         bytes.write(oidTmp.toByteArray());
 
         // encode encrypted key
-        if (this.key != null) {
+        if (k != null) {
             // XDH private key in SunEC and new Java 17 design requires [octet-string[octer-string[key-bytes]]] format,
             // otherwise, it causes interop issue. JCK issue 569
-            bytes.putOctetString(new DerValue(DerValue.tag_OctetString, this.key).toByteArray());
+            bytes.putOctetString(new DerValue(DerValue.tag_OctetString, k).toByteArray());
         }
 
         // wrap everything into a SEQUENCE
@@ -521,6 +525,8 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
         }
         if (!destroyed) {
             destroyed = true;
+            if (k != null)
+                Arrays.fill(k, (byte) 0x00);
             if (this.key != null)
                 Arrays.fill(this.key, (byte) 0x00);
             this.xecKey = null;
@@ -544,5 +550,19 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
 
     protected Object writeReplace() throws java.io.ObjectStreamException {
         return new KeyRep(KeyRep.Type.PRIVATE, getAlgorithm(), getFormat(), getEncoded());
+    }
+
+    /**
+     * Set the PKCS8Key key object.
+     * 
+     * @param k The raw key bytes, without OctetString or DER encoded.
+     * @throws IOException 
+     */
+    private void setPKCS8KeyByte(byte[] k) throws IOException {
+        if (Integer.parseInt(provider.getJavaVersionStr()) <= 11) {
+            this.key = k;
+        } else {
+            this.key = new DerValue(DerValue.tag_OctetString, k).toByteArray();
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -57,7 +58,12 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
     private void setFieldsFromXeckey() throws Exception {
         if (this.key == null) {
             this.key = extractPrivateKeyFromOCK(xecKey.getPrivateKeyBytes()); // Extract key from GSKit and sets params
-            this.scalar = Optional.of(key);
+            if (Integer.parseInt(provider.getVersionStr()) <= 11) {
+                super.key = this.key;
+            } else {
+                super.key = new DerValue(DerValue.tag_OctetString, this.key).toByteArray();
```
KostasTsiounis, line 64: I find the existence of a field with the same name, but private, a bit confusing and potentially error prone. Since it is not accessible, maybe we could just use the one variable.

I'm thinking something like:
```
if (Integer.parseInt(provider.getVersionStr()) > 11) {
       this.key = new DerValue(DerValue.tag_OctetString, this.key).toByteArray();
}
```

And I think we're good to set the scalar to `this.key` too.

Same thing goes for other spots too.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -149,6 +155,11 @@ public XDHPrivateKeyImpl(OpenJCEPlusProvider provider, AlgorithmParameterSpec pa
                 int encodingSize = CurveUtil.getDEREncodingSize(curve);
                 this.xecKey = XECKey.createPrivateKey(provider.getOCKContext(), der, encodingSize);
             }
+            if (Integer.parseInt(provider.getVersionStr()) <= 11) {
+                super.key = this.key;
+            } else {
+                super.key = new DerValue(DerValue.tag_OctetString, this.key).toByteArray();
```
KostasTsiounis, line 161: Similar to above.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -523,6 +539,8 @@ public void destroy() throws DestroyFailedException {
             destroyed = true;
             if (this.key != null)
                 Arrays.fill(this.key, (byte) 0x00);
+            if (super.key != null)
+                Arrays.fill(this.key, (byte) 0x00);
```
KostasTsiounis, line 543: I think you mean `Arrays.fill(super.key, (byte) 0x00);`.

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
taoliult: Before my changes, this class doesn't have a `key` variable. So, when assign the value to `this.key`, it actually assign the value to `PKCS8Key` class `key` object. This caused the private key encoding format issue, because after jdk11 and after version, the private key object in `PKCS8Key` class requires the `[octet string]` before it.

So, I added a variable `private byte[] key;`. All the previous operations on the `this.key` will only work for this `XDHPrivateKeyImpl` class private variable `private byte[] key;`. It won't assign the value to its parent class `PKCS8Key` private `key` object. So, it won't affect the current behaviours in `XDHPrivateKeyImpl` class.

But, we do need to set the `PKCS8Key` private `key` object, for example the encoding purpose. So, I am thinking, to use the `super.key` to explicitly set its parent class `PKCS8Key` private `key` object. And because this `super.key` object is in the `PKCS8Key` class, when the jdk version is 17 or after version, we need to add an `[octet string]` before it.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
taoliult: Right, changed it.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
KostasTsiounis: What I don't get is why not assign the `key` object in the `XDHPrivateKeyImpl` class too, after we have fixed the encoding of course. Shouldn't the superclass key also hold the proper encoding in that case too, based always on the Java version?

In other words, why keep `this.key` different than `super.key` since the former one is not reused anywhere anyway. Unless of course I'm missing something in the logic.

KostasTsiounis: If we need both, we can always have distinct names, like `rawKey` and `encodedKey`. I just feel like `this.key` and `super.key` is going to be very confusing to anyone reading it in the future.

taoliult: According to our discussion, changed `this.key` to `k` as the local private variable, same as SunEC. Changed `super.key` to `this.key` as the `key` variable in its parent class `PKCS8Key`, the similar way as what SunEC did.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
taoliult: I did not change according to above, because we still need to assign `this.key = k;` in JDK11 and before version.

jasonkatonica (changes requested):
I had a couple questions on this code.

Looking through the code I was concerned that `this.key` was not set to the same key value as `k` and vice versa at various points in time. I couldn't exactly find a case / path where this would occur but it does seem a bit fragile keeping slightly different copies of the similar values.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -55,9 +56,14 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
...
-        if (this.key == null) {
-            this.key = extractPrivateKeyFromOCK(xecKey.getPrivateKeyBytes()); // Extract key from GSKit and sets params
-            this.scalar = Optional.of(key);
+        if (k == null) {
+            k = extractPrivateKeyFromOCK(xecKey.getPrivateKeyBytes()); // Extract key from GSKit and sets params
+            if (Integer.parseInt(provider.getVersionStr()) <= 11) {
+                this.key = k;
+            } else {
+                this.key = new DerValue(DerValue.tag_OctetString, k).toByteArray();
+            }
```
jasonkatonica, line 65: Can this logic be broken out into a helper function? This same code seems to be used in 3 locations within this change.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
```diff
@@ -43,6 +43,7 @@ final class XDHPrivateKeyImpl extends PKCS8Key implements XECPrivateKey, Seriali
     private Optional<byte[]> scalar;
     private NamedParameterSpec params;
     private CURVE curve;
+    private byte[] k;
```
jasonkatonica, line 46: Perhaps we should add a comment that this is the raw key value to make it obvious this is the key value not the der encoded or octet string encoding of this key value?

jasonkatonica (changes requested):

jasonkatonica:
Also please include a `fixes` link within the above commit and PR text to issue https://github.com/IBM/OpenJCEPlus/issues/100 such that this update gets linked to the issue being fixed.

johnpeck-us-ibm (approved):
Please make sure to run the key creation and interopt tests with SunEC and BC. There was an issue with the encoding because sometime the key created had a leading zero and the use of BigInteger caused that zero to be dropped. This caused an interopt issue with BC. However, this would not always happen only once in a while. So a unit test to do this kind of test  many many times to validate that this issue does not happen.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
taoliult: Done.

src/main/java/com/ibm/crypto/plus/provider/XDHPrivateKeyImpl.java
taoliult: Done.

src/main/java/com/ibm/crypto/plus/provider/XDHPublicKeyImpl.java
```diff
@@ -411,7 +411,9 @@ public byte[] getEncoded() {
                 oidSeq.putOID(this.algid.getOID());
                 if ((oidSubSeq != null)) {
                     oidSeq.write(DerValue.tag_Sequence, oidSubSeq);
-                } else {
+                }
+                
+                if (Integer.parseInt(provider.getVersionStr()) <= 11) {
```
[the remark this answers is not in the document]
taoliult: The provider version comes from `System.getProperty("java.specification.version")` in `OpenJCEPlusProvider` class, link below. When initializing the `OpenJCEPlus` or `OpenJCEPlusFIPS` provider, there is no constructor to specify the provider version, the version will always come from the `java.specification.version`. So, the provider version should always match the Java version.

https://github.com/IBM/OpenJCEPlus/blob/main/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusProvider.java#L25

taoliult:
> I had a couple questions on this code.
> 
> Looking through the code I was concerned that `this.key` was not set to the same key value as `k` and vice versa at various points in time. I couldn't exactly find a case / path where this would occur but it does seem a bit fragile keeping slightly different copies of the similar values.

@jasonkatonica Checked the `XDHPrivateKeyImpl` from `SunEC` codes. From JDK17 and after version, the raw `k` object in `XDHPrivateKeyImpl` is different with the `key` object in `PKCS8Key`(code links below). So, I am thinking, follow the similar way as `SunEC` did. And also, run the `TestAll` tests and all the 3926 tests passed.

https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/src/jdk.crypto.ec/share/classes/sun/security/ec/XDHPrivateKeyImpl.java#L52

https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/openj9/src/jdk.crypto.ec/share/classes/sun/security/ec/XDHPrivateKeyImpl.java#L56

src/test/java/ibm/jceplus/junit/base/BaseTestXDHKeyImport.java
```diff
@@ -185,7 +185,7 @@ public void createKeyPairXDHParamImport(String alg) throws Exception {
         }
 
         NamedParameterSpec paramSpec = new NamedParameterSpec(alg);
-        KeyFactory kf = KeyFactory.getInstance("XDH");
+        KeyFactory kf = KeyFactory.getInstance("XDH", providerName);
```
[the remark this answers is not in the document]
taoliult: Yes, I added another test named `TestXDHInteropBC` for using bouncy castle provider, it still extends the `BaseTestXDHInterop`, just changing to use `BC` for the interop provider. And same as `TestXDHInterop` test, it only added into OpenJCEPlus `TestAll`. 

The test results are:

```
[INFO] Running ibm.jceplus.junit.openjceplus.TestXDHInterop
Testing XDH key generated with provider SunEC using provider OpenJCEPlus
Testing XDH key generated with provider SunEC using provider OpenJCEPlus
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.011 s -- in ibm.jceplus.junit.openjceplus.TestXDHInterop
[INFO] Running ibm.jceplus.junit.openjceplus.TestXDHInteropBC
Testing XDH key generated with provider BC using provider OpenJCEPlus
Testing XDH key generated with provider BC using provider OpenJCEPlus
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.058 s -- in ibm.jceplus.junit.openjceplus.TestXDHInteropBC
```

src/main/java/com/ibm/crypto/plus/provider/XDHPublicKeyImpl.java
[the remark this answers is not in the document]
taoliult: Added a new method `getJavaVersionStr(`) in `OpenJCEPlusProvider`, to return the `java.specification.version` in case the provider version may different with the Java version in the future.

src/main/java/com/ibm/crypto/plus/provider/XDHPublicKeyImpl.java
```diff
@@ -411,7 +411,9 @@ public byte[] getEncoded() {
                 oidSeq.putOID(this.algid.getOID());
                 if ((oidSubSeq != null)) {
                     oidSeq.write(DerValue.tag_Sequence, oidSubSeq);
-                } else {
+                }
```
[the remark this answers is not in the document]
taoliult: According to the previous code, the logic is, if `oidSubSeq == null`, then `oidSeq.putNull();`. And this `oidSubSeq` will only be `null` if `CurveUtil` is not `FFDHE`. If the `CurveUtil` is `FFDHE`, then `oidSubSeq != null`, then it won’t `oidSeq.putNull();`.

But according to the test and also the comment in the codes, this issue is the `XDH` format changes between JDK11 or before and JDK17 or after, no matter the `Curve` is `FFDHE` or not.

So, right now, the current logic is, no matter is `FFDHE` or not, if the JDK version is JDK11 or before, then `oidSeq.putNull();`, if the JDK version is JDK17 or after, then don’t `oidSeq.putNull();`.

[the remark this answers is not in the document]
KostasTsiounis: Yeah, but the difference should be there if the curve is not `FFDHE`, in which case you add `NULL` in `Java <=11` and nothing otherwise. I don't think you want to add `NULL` if the curve is `FFDHE`, which is what will happen now if `Java <=11`, where you will have both the `oidSubSeq` and `NULL`.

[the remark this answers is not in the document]
KostasTsiounis: As I understand it, we would need something like:
```
if ((oidSubSeq != null)) {
   oidSeq.write(DerValue.tag_Sequence, oidSubSeq);
} else if (Integer.parseInt(provider.getVersionStr()) <= 11) {
   oidSeq.putNull();
}
```
Unless, of course, there is something else I'm missing.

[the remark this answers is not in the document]
taoliult: Right, agree, updated the code.

KostasTsiounis (approved):
LGTM

src/test/java/ibm/jceplus/junit/base/BaseTestXDHKeyImport.java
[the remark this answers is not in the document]
taoliult: Check this TestXDHInteropBC test between JDK11 and JDK17, will update the result here shortly.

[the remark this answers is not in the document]
KostasTsiounis: So, this test seems to be using `BC` to generate the keys and then consume the encoded version with `OpenJCEPlusFIPS`. This means that the keys are created and consumed under the same Java version. If we want to do interop between Java versions, we might need to hardcode some encoded keys.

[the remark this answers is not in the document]
jasonkatonica: The import portion of this code was changed in previous updates such that this should work. Importing XDH encoding with and without nested encodings should be allowable. I do agree that ideally we add a few more hardcoded key values to double check this with formats for these encodings spanning Java 11 to 17+.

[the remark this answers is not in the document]
taoliult: Working on it.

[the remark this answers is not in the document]
jasonkatonica: OK this does make sense for interopt testing between the `bcprov-jdk18on` version of bouncy castle and our code running on various Java releases. This aspect of the test does seem like it is covered.

jasonkatonica (approved):
LGTM

PR #146, opened 2024-07-30
Create new mvn profile to aid with bundled SDK test execution

jasonkatonica:
When executing tests included in the `OpenJCEPlus` project a user may wish to make use of the bundled versions of the `OpenJCEPlus` code, not the included `OpenJCEPlus` source code included in this project.

This update provides an alternate `mvn` profile that users can activate using the option `-Dtestenvironment=standalonebundled` when executing tests against an SDK that already has `OpenJCEPlus` bundled.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,7 +1,7 @@
 <!--
 ###############################################################################
 #
-# Copyright IBM Corp. 2023
+# Copyright IBM Corp. 2023, 2024
 #
 # Licensed under the Apache License 2.0 (the "License").  You may not use
 # this file except in compliance with the License.  You can obtain a copy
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -114,6 +114,110 @@
                 <jdk.build.target>22</jdk.build.target>
             </properties>
           </profile>
+          <!--
+          Profile expected to be active most of the time whenever testing what was produced by
+          this project. This includes the providers built and local test artifacts.
+          -->
+          <profile>
+            <id>defaultestenv</id>
```
KostasTsiounis, line 122: Can we change that to something more descriptive?

pom.xml
```diff
@@ -114,6 +114,110 @@
...
+              </plugins>
+            </build>
+          </profile>
+          <!--
+          Profile expected to be active when testing OpenJCEPlus that is contained within 
+          a bundled SDK. A provider built by the project locally is not tested in this case,
+          instead the one within the SDK is tested.
+          -->
+          <profile>
+            <id>standalonetestconfig</id>
```
KostasTsiounis, line 187: Can we change that to something more descriptive?

pom.xml
```diff
@@ -230,6 +334,20 @@
                         <arg>java.base/sun.util.logging=openjceplus</arg>
                         <arg>--add-exports </arg>
                         <arg>java.base/jdk.internal.logger=openjceplus</arg>
+                        <!--
+                        Exports statements below this line are necessary for
+                        compiling only the test source with target `test-compile`.
+                        -->
+                        <arg>--add-exports </arg>
```
KostasTsiounis, line 341: Since those are exactly the same as above, with only the part after `=` changing from `openjceplus` to `ALL-UNNAMED`, could that be a variable set as part of the profile?

KostasTsiounis (changes requested):

pom.xml
jasonkatonica: I updated this id value to be much more descriptive.

pom.xml
jasonkatonica: I updated this id value to be much more descriptive.

pom.xml
jasonkatonica: I moved what I could to the common location and exported to a target location based upon active profile. 

I did however see that a few of these exports are only needed to run tests against the bundled build. These are not there currently. I also enhanced the comment accordingly.

KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

PR #147, opened 2024-07-31
Run all multithreaded tests and print stack trace from all failures

KostasTsiounis:
The multithreaded tests included only an assertion that there weren't any failures as part of each of the individual tests. If any failures would occur, they wouldn't be saved or printed. However, the overall multithreaded test would fail without running the rest of the individual tests.

The execution of the tests is changed to return the potential failures, instead of directly failing. Said failures are printed when all threads conclude. All of the individual tests conclude before the overall multithreaded test fails.

### DIFF
--- a/src/test/java/ibm/jceplus/junit/TestMultithread.java
+++ b/src/test/java/ibm/jceplus/junit/TestMultithread.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2023, 2024
  *
  * Licensed under the Apache License 2.0 (the "License").  You may not use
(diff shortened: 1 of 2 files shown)


jasonkatonica (changes requested):
Seems the copyrights need to be updated in these two files. Other then that I did not have any specific comments.

KostasTsiounis:
> Seems the copyrights need to be updated in these two files. Other then that I did not have any specific comments.

Copyrights updated.

jasonkatonica (approved):
LGTM

johnpeck-us-ibm (approved):

jasonkatonica:

PR #153, opened 2024-07-31
Update the isAssignableFrom checks

taoliult:
According to https://bugs.openjdk.org/browse/JDK-8279800, isAssignableFrom means is-parent-of and you can only cast an object of a child class to that of a parent class.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESKeyFactory.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESKeyFactory.java
@@ -84,7 +84,7 @@ public final class AESKeyFactory extends SecretKeyFactorySpi {
                 && (key.getFormat().equalsIgnoreCase("RAW"))) {
 
             // Check if requested key spec is amongst the valid ones
-            if (SecretKeySpec.class.isAssignableFrom(keySpec)) {
+            if (keySpec.isAssignableFrom(SecretKeySpec.class)) {
                 return new SecretKeySpec(key.getEncoded(), key.getAlgorithm());
             } else {
                 throw new InvalidKeySpecException("Inappropriate key specification");
--- a/src/main/java/com/ibm/crypto/plus/provider/ChaCha20KeyFactory.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ChaCha20KeyFactory.java
@@ -84,7 +84,7 @@ public final class ChaCha20KeyFactory extends SecretKeyFactorySpi {
                 && (key.getFormat().equalsIgnoreCase("RAW"))) {
 
             // Check if requested key spec is amongst the valid ones
-            if (SecretKeySpec.class.isAssignableFrom(keySpec)) {
+            if (keySpec.isAssignableFrom(SecretKeySpec.class)) {
                 return new SecretKeySpec(key.getEncoded(), key.getAlgorithm());
             } else {
                 throw new InvalidKeySpecException("Inappropriate key specification");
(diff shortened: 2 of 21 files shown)


src/main/java/com/ibm/crypto/plus/provider/AESKeyFactory.java
```diff
@@ -84,7 +84,7 @@ protected KeySpec engineGetKeySpec(SecretKey key, Class<?> keySpec)
                 && (key.getFormat().equalsIgnoreCase("RAW"))) {
 
             // Check if requested key spec is amongst the valid ones
-            if (SecretKeySpec.class.isAssignableFrom(keySpec)) {
+            if (keySpec.isAssignableFrom(SecretKeySpec.class)) {
```
KostasTsiounis, line 87: In this scenario where there is no cast, do we want to check if `keySpec` is a parent of `SecretKeySpec`, or the other way around? I would think we want `keySpec` to be some subclass of `SecretKeySpec` or even `instanceOf(SecretKeySpec)`, right?

src/main/java/com/ibm/crypto/plus/provider/ChaCha20KeyFactory.java
```diff
@@ -84,7 +84,7 @@ protected KeySpec engineGetKeySpec(SecretKey key, Class<?> keySpec)
                 && (key.getFormat().equalsIgnoreCase("RAW"))) {
 
             // Check if requested key spec is amongst the valid ones
-            if (SecretKeySpec.class.isAssignableFrom(keySpec)) {
+            if (keySpec.isAssignableFrom(SecretKeySpec.class)) {
```
KostasTsiounis, line 87: Same as above.

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/AESKeyFactory.java
KostasTsiounis: Or maybe I'm wrong, cause whoever gets the result of this method will try to cast it into the `keySpec` class probably. It is a bit confusing. @jasonkatonica What do you think?

taoliult: I have the same concern before when dealing with those without cast. 

1. I check the OpenJDK codes which update the `isAssignableFrom` checks (https://github.com/openjdk/jdk/pull/7037/files#diff-743b02b03ca8fa1bda2310721348024c26e1d796c83a7671b03e9332678a10d4R246). And it did the same.

2. From my understanding (Please correct me if I am wrong). It did not explicitly cast, but it `return new SecretKeySpec(key.getEncoded(), key.getAlgorithm()); `which will, in some kind of implicitly way, to cast as the KeySpec.

KostasTsiounis: `KeySpec` is an interface that is implemented by `SecretKeySpec`, so there is no need for a cast there. What I, also, noticed though is that `SecretKeySpec` has no parents, other than `Object`. So this check practically verifies whether the `keySpec` parameter is the same class as `SecretKeySpec`.

KostasTsiounis (approved):
LGTM

jasonkatonica (changes requested):
Code looks good to me. However we have very minimal coverage in the tests today for the `getParameterSpec` APIs in the various key factory classes. Some additional tests to exercise these lines changed ( and better coverage in general for `getParameterSpec` ) would be great to have.

taoliult:
> Code looks good to me. However we currently have minimal coverage in the tests today for the `getParameterSpec` APIs in the various key factory and parameter classes. Some additional tests to exercise these lines changed ( and better coverage in general for `getParameterSpec` ) would be great to have.

Yes, working on the test case.

jasonkatonica (approved):
LGTM

taoliult:
Squash the commits into 1 commit.

jasonkatonica:

johnpeck-us-ibm (approved):

PR #154, opened 2024-08-07
Update the isAssignableFrom checks

taoliult:
This is a back-port PR from PR https://github.com/IBM/OpenJCEPlus/pull/153

According to https://bugs.openjdk.org/browse/JDK-8279800, isAssignableFrom means is-parent-of and you can only cast an object of a child class to that of a parent class.


jasonkatonica (approved):
LGTM

johnpeck-us-ibm (approved):

PR #169, opened 2024-08-14
Fix the encoding formats for X25519 and X448 (#136)

taoliult:
This is a back-port PR from PR https://github.com/IBM/OpenJCEPlus/pull/136

This commit fixes the XDH Private Key and Public Key encoding formats issues.

According to the PKCS#8 Private-Key Specification, the new format privateKey is an octet string whose contents are the value of the private key. So, adding the octet string before the private key when passing the private key object to PKCS8Key key object for 17 and after version.

According to Sun old versions, 11 and before, the new XDH format is not supported. So, adding a DER "null" value on the OID sequence only for 11 and before versions.


jasonkatonica (approved):
LGTM

jasonkatonica:
ppc64_aix, x86_64_linux, ppc64le_linux, s390x_linux, x86_64_windows

jasonkatonica:
ppc64_aix, x86_64_linux, ppc64le_linux, s390x_linux, x86_64_windows

PR #181, opened 2024-08-24
Skip native compilation when testing bundled SDK

jasonkatonica:
When testing against an `SDK` that already includes `OpenJCEPlus` there
is no reason to compile the native libraries as they are already
included in the runtime. We should skip the native compilation and make
use of the library included in the `SDK` itself.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,6 +196,8 @@
             </activation>
             <properties>
               <test.compilation.export.target>ALL-UNNAMED</test.compilation.export.target>
+              <skip.native.compile>true</skip.native.compile>
+              <build.native.file>None_not_really_used_here</build.native.file>
             </properties>
             <build>
               <plugins>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -196,6 +205,8 @@
             </activation>
             <properties>
               <test.compilation.export.target>ALL-UNNAMED</test.compilation.export.target>
+              <skip.native.compile>true</skip.native.compile>
+              <build.native.file>placeholder_does_not_exist</build.native.file>
```
KostasTsiounis, line 209: I think `placeholder_does_not_exist` is a bit confusing. Maybe something like `None_not_really_used_here`, or anything else that is easier to get?

KostasTsiounis (changes requested):

taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

pom.xml
jasonkatonica: Updated this value to be `None_not_really_used_here` which also seems reasonable and does convey the message that the value is not really used.

KostasTsiounis (approved):
LGTM

PR #183, opened 2024-09-03
Skip native compilation when testing bundled SDK

jasonkatonica:
When testing against an `SDK` that already includes `OpenJCEPlus` there is no reason to compile the native libraries as they are already included in the runtime. We should skip the native compilation and make use of the library included in the `SDK` itself.


KostasTsiounis (approved):
LGTM

PR #189, opened 2024-09-06
Update Werror CFLAG in AIX

KostasTsiounis:
The `-Werror` flag is not supported by the xlc compiler. Instead, the equivalent `-qhalt=w` can be used.

Signed-off-by: Kostas Tsiounis <kostas.tsiounis@ibm.com>

PR #195, opened 2024-09-10
Update Werror CFLAG in AIX

KostasTsiounis:
The `-Werror` flag is not supported by the xlc compiler. Instead, the equivalent `-qhalt=w` can be used.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/189


jasonkatonica (approved):
LGTM

PR #200, opened 2024-09-16
Refactor some signature related tests for Junit5

jasonkatonica:
This update migrates a portion of the tests associated with signatures to Junit5 format. Updates include:
- A new BaseTestJunit5 class is provided that tests can extend in this update and in the future.
- Main methods in individual tests were present inconsistently through the test code and have been removed given that individual tests can be executed directly via mvn surefire or other Junit5 runners.
- Constructors for various tests have also been removed since they are not in use.
- The `TestTLS` was found to impact other tests depending on the order tests are run. If the `OpenJCEPlus` provider is present when running `OpenJCEPlusFIPS` related tests then it is removed, then restored after the test is run. A similar pattern is followed when executing a similar test with `OpenJCEPlus`.
- A few methods were found to be private and should not be declared as public.
- Most tests are currently written to expect a single instance of a given test for each @Test method specified. By default this is not the behavior in Junit5. This update tags various tests with @TestInstance of Lifecycle.PER_CLASS. This tells JUnit to create only one instance of the test class and then reuse it between tests tagged with @Test.
- `static` blocks are often replaced with `@BeforeAll` syntax executed when a test is constructed.
- The higher level test suite classes were modified to also use junit5 syntax ( `@SelectClasses` and `@Suite`).
- Empty comment blocks have been removed as they have no value.
- Mulithreaded test case drivers have been modified to use new Junit5 launchers to launch test classes contained in
`org.junit.platform.launcher package`.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -27,6 +27,7 @@
         <module name="EmptyBlock"/>
         <module name="ImportOrder"/>
     </module>
+    <module name="NewlineAtEndOfFile"/>
     <module name="RegexpHeader">
         <property name="fileExtensions" value="java"/>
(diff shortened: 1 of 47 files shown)


KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

PR #233, opened 2024-10-04
Use openjdk Debug class

### LINKED ISSUE #225: Eliminate OpenJCEPlus Debug class
The Debug class located at https://github.ibm.com/runtimes/OpenJCEPlus/tree/1c0979e7edd1bbad69472ffddb74daee9535d528/src/main/java/com/ibm/misc currently is very similar to the Debug class that exists in the following location https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/openj9/src/java.base/share/classes/sun/security/util/Debug.java is it possible to make use of this class instead of our own?

jasonkatonica:
This update eliminates the use of the last existing class in the `com.ibm.misc` package, which is used for debug traces. Code will for the time being make use of the `sun.security.util.Debug` class instead.

The provider code that currently uses the `com.ibm.misc.Debug` class now makes direct use of the `sun.security.util.Debug` class.

The test code that has more advanced use of the current `Debug` class now contains its own copy of the `Debug` class that once resided in the `com.ibm.misc` package.

Closes #225

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,6 +138,7 @@
                   <configuration>
                     <useModulePath>false</useModulePath>
                     <!-- Could add -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=0.0.0.0:8787 for debug -->
+                    <!-- Could add -Djava.security.auth.debug=all for debug -->
                     <argLine>
                       @{argLine}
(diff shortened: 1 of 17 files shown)


KostasTsiounis (approved):
LGTM

jasonkatonica:

PR #245, opened 2024-10-11
Convert AESCipherInputStream tests to junit5 and fix error messages

jasonkatonica:
This update fixes some error messages being printed to the logs that stated some tests failed when they did not really fail and were running as expected.

Changes include
- Update `junit jupiter` version to `5.11.1`
- Remove `junit-bom` jar dependency that is no longer needed.
- Update `junit-platform-suite` to latest version `1.11.1`
- Convert the `BaseTestAESCipherInputStreamExceptions` class to Junit 5
- Fix incorrect error messages being reported and visible that look as follows since they are not found to be accurate:
    ```
    System.out.println(" Fail: " + ec.getMessage());
    ```
- Convert documentation to javadoc.


KostasTsiounis (approved):
LGTM

taoliult (approved):
LGTM

PR #252, opened 2024-10-11
Update tests to use Junit 5

jasonkatonica:
Tests have been updated to use Junit5. Changes made include:
- Removing the `junit-vintage-engine` dependency. This is no longer in use given all tests now make use of Junit 5.
- Tests are sorted by alphabetical in test suites.
- Tests no longer need to extend `junit.framework.TestCase`.
- Tests are explicitly declared with the `@Test` annotation.
- Test constructors have been removed. Arguments to the base tests are now sent using getters and setters. This is typically only used for setting providers, algorithms, or key sizes in tests.
- Various `main()` methods of tests were removed as they are no longer in use. Tests can be run using various Junit 5 test execution tools such as IDEs, command line tools, and `mvn` surefire which is already in place in this project.
- `BaseTest` was removed as it is no longer in use and represents a Junit 3 test. A new `BaseTestJunit5` test has been created to represent the new base for the test hierarchy.
- The concept of warming up a test has been removed from a few locations. Tests are not expected to act as performance tests.
- Copyrights were updated.
- Various empty comment blocks were removed.
- Some tests were not executing using the expected provider since the incorrect `Utils` class was in use. This was corrected to use the correct class.
- Tests `TestAESGCM_192`, `TestAESGCM_256`, and `TestRSA_4096` were added to the `OpenJCEPlus` and `OpenJCEPlusFIPS` provider test suites.
- Tests `TestECDHInteropBC`, `TestRSATypeCheckEnabled`, `TestAES_192`, and `TestAES_256` were added to the `OpenJCEPlus` provider test suite
- Tests `TestAES_192` and `TestAES_256` were removed from the `TestMultithreadFIPS` since these tests take a long period of time to complete. Coverage functionally is still maintained since they are executed as part of the `OpenJCEPlusFIPS` provider tests.
- Increased timeouts associated with tests `TestMultithreadFIPS` and `TestMultithread`.


johnpeck-us-ibm (approved):

jasonkatonica:
Rebased on latest `main` branch due to conflicts.

jasonkatonica:
Rebased on latest main branch.

jasonkatonica:
Rebased on latest `main` branch.

jasonkatonica:
Rebased on latest code.

src/test/java/ibm/jceplus/junit/base/memstress/BaseTestMemStressRSAPSS2.java
```diff
@@ -89,7 +84,7 @@ public void testRSAPSSSignature() throws Exception {
         long currentUsedMemory = 0;
         long prevUsedMemory = 0;
 
-        KeyPairGenerator keyGen = KeyPairGenerator.getInstance("RSA", providerName);
+        KeyPairGenerator keyGen = KeyPairGenerator.getInstance("RSA", getProviderName());
         keyGen.initialize(this.keysize, new java.security.SecureRandom());
```
KostasTsiounis, line 88: Shouldn't this change to `getKeySize()` too?

src/test/java/ibm/jceplus/junit/base/memstress/BaseTestMemStressXDH.java
```diff
@@ -17,24 +18,18 @@
...
     String curveName = "X25519";
 
-    //"X448";
-    public BaseTestMemStressXDH(String providerName) {
-        super(providerName);
-    }
-
-    public BaseTestMemStressXDH(String providerName, String curveName) {
-        super(providerName);
-        this.curveName = curveName;
```
KostasTsiounis, line 35: Do we have a way to set to a different curve?

src/test/java/ibm/jceplus/junit/openjceplus/TestAESGCM.java
```diff
@@ -8,45 +8,17 @@
...
-    //
-    //
-    public TestAESGCM() {
-        super(Utils.TEST_SUITE_PROVIDER_NAME);
-    }
-
-    //--------------------------------------------------------------------------
-    //
-    //
-    public TestAESGCM(int keySize) throws Exception {
```
KostasTsiounis, line 33: Do we have a way of trying different key sizes or is that not something we want?

KostasTsiounis (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestAESCCM.java
```diff
@@ -49,23 +50,10 @@ public class BaseTestAESCCM extends BaseTest {
...
 
-    public static String provider;
     public static boolean printJunitTrace = false;
 
-
-
-    public BaseTestAESCCM(String providerName) {
-        super(providerName);
-        provider = providerName;
-        printJunitTrace = Boolean
```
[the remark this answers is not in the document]
jasonkatonica: I updated this back to what we had before. The value of `printJunitTrace` is set using a java property. I am not sure we will use this but we can keep that behavior the same for now.

src/test/java/ibm/jceplus/junit/base/BaseTestAESCCM2.java
```diff
@@ -50,22 +51,9 @@ public class BaseTestAESCCM2 extends BaseTest {
...
 
-    public String provider;
-
     public boolean printJunitTrace = false;
 
-
-    public BaseTestAESCCM2(String providerName) {
-        super(providerName);
-        provider = providerName;
-        printJunitTrace = Boolean
```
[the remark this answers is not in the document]
jasonkatonica: I updated this back to what we had before. The value of `printJunitTrace` is set using a java property. I am not sure we will use this but we can keep that behavior the same for now.

src/test/java/ibm/jceplus/junit/base/memstress/BaseTestMemStressRSAPSS2.java
jasonkatonica: Yes agreed. Change this test to use getKeySize() instead such that this is more consistent.

src/test/java/ibm/jceplus/junit/base/memstress/BaseTestMemStressXDH.java
jasonkatonica: Yes we can use getAlgorithm(). I made this update in the latest code.

src/test/java/ibm/jceplus/junit/base/BaseTestAESGCM.java
```diff
@@ -90,20 +90,26 @@ public class BaseTestAESGCM extends BaseTestJunit5 {
 
     protected KeyGenerator aesKeyGen;
     protected SecretKey key;
-    protected AlgorithmParameters params = null;
+    protected AlgorithmParameters params = null;
```
KostasTsiounis, line 93: Can you remove the extra space?

src/test/java/ibm/jceplus/junit/base/BaseTestAESGCM.java
```diff
@@ -90,20 +90,26 @@ public class BaseTestAESGCM extends BaseTestJunit5 {
...
-        if (specifiedKeySize > 0) {
-            aesKeyGen.init(specifiedKeySize);
+        int keySize = -1;
+        try {
+            keySize = getKeySize();
+        } catch (RuntimeException e) {
+            // Ignore exception since keysize was not set.
+            // This is OK since we intend to test the default keysize.
+        }
+        if (keySize > 0) {
```
KostasTsiounis, line 111: This only makes sense if the key is actually set, so should we put it in the try block after getting the key? Just a suggestion. I'm good either way.

KostasTsiounis (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestAESGCM.java
jasonkatonica: Removed this space.

src/test/java/ibm/jceplus/junit/base/BaseTestAESGCM.java
jasonkatonica: In this location the value of `keySize` is set to -1 on line 104. We then attempt to set the value and ignore an exception if its thrown since the `getKeySize()` method will throw an exception if it does not have a positive value.

By the time we get to line 111 the value should be `-1` if not set, or a positive value if set by the test. 

Hopefully this make sense?

KostasTsiounis: I understand how this works now. I was just saying that checking only makes sense if a key size is obtained, thus why I suggested putting in it into the try block and avoiding it altogether if an exception occurs. But again, more of a stylistic comment here.

KostasTsiounis: I was thinking something like: 

```
try {
    int keySize = getKeySize();
    if (keySize > 0) {
        aesKeyGen.init(keySize);
    }
} catch (RuntimeException e) {
    // Ignore exception since keysize was not set.
    // This is OK since we intend to test the default keysize.
}
```

KostasTsiounis (approved):
LGTM

src/test/java/ibm/jceplus/junit/base/BaseTestHmacMD5.java
```diff
@@ -277,22 +246,4 @@ public void testHmacMD5_key7() throws Exception {
         assertTrue("Mac digest did not equal expected", Arrays.equals(digest, digest_7));
 
     }
-
-    //--------------------------------------------------------------------------
-    //
-    //
-    public void warmup() throws Exception {
```
taoliult, line 284: Don't we need this `warmup()` method?

src/test/java/ibm/jceplus/junit/base/Record.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2023, 2024
```
taoliult, line 2: Do we need to update the copyright if we did not modify the file?

src/test/java/ibm/jceplus/junit/base/memstress/BaseTestMemStressAES.java
```diff
@@ -9,87 +9,44 @@
...
-        encodedKey = new byte[(specifiedKeySize > 0 ? specifiedKeySize : 128) / 8];
+        this.specifiedKeySize = getKeySize();
+        encodedKey = new byte[specifiedKeySize / 8];
         r.nextBytes(plainText1024);
 
         //key = new SecretKeySpec(encodedKey, 0, encodedKey.length, "AES");
-        keyGenerator = KeyGenerator.getInstance("AES", "OpenJCEPlus");
-        keyGenerator.init(DEFAULT_KEYSIZE);
+        keyGenerator = KeyGenerator.getInstance("AES", getProviderName());
+        keyGenerator.init(getKeySize());
```
taoliult, line 49: We could use `this.specifiedKeySize` here instead of calling the `getKeySize()` method again.

taoliult (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestAESCopySafe.java
```diff
@@ -32,27 +33,17 @@ public class BaseTestAESCopySafe extends BaseTest {
...
-
-    public BaseTestAESCopySafe(String providerName) {
-        super(providerName);
-        this.specifiedKeySize = 128;
-    }
-
-    public BaseTestAESCopySafe(String providerName, int keySize) throws Exception {
-        super(providerName);
-        this.specifiedKeySize = keySize;
-        Assume.assumeTrue(javax.crypto.Cipher.getMaxAllowedKeyLength("AES") >= keySize);
```
JinhangZhang, line 45: Should we keep the keysize check?

src/test/java/ibm/jceplus/junit/base/BaseTestAES.java
```diff
@@ -27,44 +26,42 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 public class BaseTestAES extends BaseTestCipher {
 
     // 14 bytes: PASSED
-    static final byte[] plainText14 = "12345678123456".getBytes();
```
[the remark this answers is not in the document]
jasonkatonica: I put static back into this set of declarations. I was removing many static declarations for tests that are multithreaded. However since this is a `final` declaration i think this is prob ok to put back in.

src/test/java/ibm/jceplus/junit/base/BaseTestAESCopySafe.java
jasonkatonica: I added this assumption back in. In general this test is not configuring different sized keys in the tests but instead always makes use of a key size of 128. At the time i removed this assumption check since it seemed redundant and stuck out from our other pattern of code here but i added it back in again as it shouldn't hurt anything.

src/test/java/ibm/jceplus/junit/base/BaseTestDHInterop.java
```diff
@@ -19,89 +19,52 @@
...
-
-        compute_dh_key_interop(idString, dhps, providerName, interopProviderName);
-        compute_dh_key_interop(idString, dhps, interopProviderName, providerName);
-
+        String idString = (Integer.valueOf(getKeySize())).toString();
 
+        DHParameterSpec dhps = generateDHParameters(getKeySize());
 
+        compute_dh_key_interop(idString, dhps, getProviderName(), getInteropProviderName());
+        compute_dh_key_interop(idString, dhps, getInteropProviderName(), getProviderName());
```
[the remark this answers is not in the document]
jasonkatonica: Makes sense. I update the test to call these methods only once.

I also noticed while ensuring other tests follow the same pattern that we are doing this same test code twice. The test `testDH_DHSpec()` was eliminated.

src/test/java/ibm/jceplus/junit/base/memstress/BaseTestMemStressAES.java
jasonkatonica: Yes agreed. I used `this.specifiedKeySize` in this location to save one method call.

src/test/java/ibm/jceplus/junit/base/Record.java
jasonkatonica: I reverted the content of this file and removed the copyright change. While searching and replacing i forgot that we had a few tests that were Junit 5 already so this file was unchanged and does not need a copyright change given that the contents are unchanged.

src/test/java/ibm/jceplus/junit/base/BaseTestAES.java
```diff
@@ -74,220 +71,239 @@ public class BaseTestAES extends BaseTestCipher {
...
+    @Test
     public void testAES() throws Exception {
         encryptDecrypt("AES");
     }
 
+    @Test
     public void testAES_CBC_ISO10126Padding() throws Exception {
         try {
-            cp = Cipher.getInstance("AES/CBC/ISO10126Padding", providerName);
+            cp = Cipher.getInstance("AES/CBC/ISO10126Padding", getProviderName());
```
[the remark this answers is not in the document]
jasonkatonica: In general i did find that using the pattern of having getter and setter methods for private fields would be easiest to maintain.

The main incentive to follow this pattern was that when tests are run they can run in any order and are likely to need values associated with a variation such as providers, key sizes, algorithm names, etc. To set these values in another copy of all the fileds in the `BaseTestJunit5` class would require reliably copying all these values ( via a single getter ) into another set of fields for the test run. Doing this in a `@BeforeEach` call would end up potentially making more calls then necessary as it would always be called before each `@Test` even if one variation in a file needs it. Using `@BeforeAll` methods can also work however i have found that if a hierarchy of tests end up having more then one @BeforeAll declared in extended classes then this ended up also causing issues running the test that might not be obvious later on. In general id expect calling getter and setter methods to have very minimal impact on overall performance and has very little downside ( id expect jit to optimize the method call out in most paths ) while makeing the code more understandable ( at least to me :) ).

src/test/java/ibm/jceplus/junit/base/BaseTestHmacMD5.java
jasonkatonica: I dont anticipate using this test as a performance test. I dont believe we need to loop in this test to perform a "warmup".

taoliult (approved):
LGTM

JinhangZhang (approved):
LGTM

PR #263, opened 2024-10-14
Use correct output size for AES on z platform

### LINKED ISSUE #240: Fix AES cipher output size issue for z/OS and LOZ
When the fast command is enabled, the AES cipher does not return the correct output size on z/OS.

The simplified failing test on z/OS and LoZ:

```
import javax.crypto.Cipher;
import javax.crypto.KeyGenerator;
import javax.crypto.SecretKey;
import java.nio.ByteBuffer;

public class TestAesCbc {

        testAesCbc();
    }

    private static void testAesCbc() throws Exception {
        byte[] data = new byte[1500];
        Cipher cipher = Cipher.getInstance("AES/CBC/PKCS5Padding", "OpenJCEPlus");
        KeyGenerator kg = KeyGenerator.getInstance("AES");
        SecretKey key = kg.generateKey();
        cipher.init(Cipher.ENCRYPT_MODE, key, cipher.getParameters());
        ByteBuffer in = ByteBuffer.allocateDirect(data.length);
        ByteBuffer out = ByteBuffer.allocateDirect(cipher.getOutputSize(data.length));
        int updateLen = data.length / 2;
        in.limit(updateLen);
        cipher.update(in, out);
        in.limit(in.capacity());
        cipher.doFinal(in, out);
    }
}
```

The failing JDK test: https://github.com/openjdk/jdk/blob/master/test/jdk/javax/crypto/CipherSpi/ResetByteBuffer.java

**Root Cause:**

The issue happens when the `use_z_fast_command` is enabled for z platforms only.
It uses a different buffer and the `engineGetOutputSize2` method.
However, the AESCipher getOutputSize was not using that method, causing the issue and returning the wrong output size on z platforms.

**Fix:**

The `engineGetOutputSize2` should be renamed to the `getOutputSizeForZ`.
Then in the `engineGetOutputSize` method, the `use_z_fast_command` should be checked.
If it is true, then it should return `getOutputSizeForZ`.
Otherwise, it returns `symmetricCipher.getOutputSize`.

jasonkatonica:
When the fast z command is available the AES cipher does not return the correct output size on platform z.

Platform z uses a different buffer and output size calculation. The method `engineGetOutputSize` when running on platform z should return and take this into account.

A new test was added to exercise byte buffers and various combinations of encryption, decryption, and orders of doFinal and update operations which can be used to recreate the problem.

Closes #240

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESCipher.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESCipher.java
@@ -56,15 +56,9 @@ public final class AESCipher extends CipherSpi implements AESConstants {
         checkCipherInitialized();
 
         try {
-            byte[] output = null;
-            int outputLen = -1;
-            if (use_z_fast_command) {
-                output = new byte[engineGetOutputSize2(inputLen)];
-                outputLen = engineDoFinal(input, inputOffset, inputLen, output, 0);
-            } else {
-                output = new byte[engineGetOutputSize(inputLen)];
-                outputLen = symmetricCipher.doFinal(input, inputOffset, inputLen, output, 0);
-            }
+            byte[] output = new byte[engineGetOutputSize(inputLen)];
+            int outputLen = engineDoFinal(input, inputOffset, inputLen, output, 0);
+
             if (outputLen < output.length) {
                 byte[] out = Arrays.copyOfRange(output, 0, outputLen);
                 if (!encrypting) {
@@ -192,7 +186,11 @@ public final class AESCipher extends CipherSpi implements AESConstants {
     @Override
     protected int engineGetOutputSize(int inputLen) {
         try {
-            return symmetricCipher.getOutputSize(inputLen);
+            if (use_z_fast_command) {
+                return getOutputSizeForZ(inputLen);
+            } else {
+                return symmetricCipher.getOutputSize(inputLen);
+            }
         } catch (Exception e) {
             throw provider.providerException("Unable to get output size", e);
         }
@@ -352,7 +350,7 @@ public final class AESCipher extends CipherSpi implements AESConstants {
             byte[] output = null;
             int outputLen = -1;
             if (use_z_fast_command) {
-                output = new byte[engineGetOutputSize2(inputLen)];
+                output = new byte[getOutputSizeForZ(inputLen)];
                 outputLen = engineUpdate(input, inputOffset, inputLen, output, 0);
             } else {
                 output = new byte[engineGetOutputSize(inputLen)];
@@ -516,7 +514,7 @@ public final class AESCipher extends CipherSpi implements AESConstants {
      * @param inputLen
      * @return
      */
-    private int engineGetOutputSize2(int inputLen) {
+    private int getOutputSizeForZ(int inputLen) {
         int totalLen = Math.addExact(buffered, inputLen);
         if (padding == Padding.NoPadding || !encrypting)
             return totalLen;
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestResetByteBuffer.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestResetByteBuffer.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright IBM Corp. 2024
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution.
+ */
+
+package ibm.jceplus.junit.base;
+
+import java.nio.ByteBuffer;
+import java.util.stream.Stream;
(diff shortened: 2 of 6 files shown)


johnpeck-us-ibm (approved):

JinhangZhang (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/AESCipher.java
```diff
@@ -59,7 +59,7 @@ protected byte[] engineDoFinal(byte[] input, int inputOffset, int inputLen)
             byte[] output = null;
             int outputLen = -1;
             if (use_z_fast_command) {
-                output = new byte[engineGetOutputSize2(inputLen)];
+                output = new byte[getOutputSizeForZ(inputLen)];
```
KostasTsiounis, line 62: Since the `engineGetOutputSize()` now checks `use_z_fast_command`, do we really need the if-else statement here?

src/test/java/ibm/jceplus/junit/base/BaseTestResetByteBuffer.java
```diff
@@ -0,0 +1,136 @@
...
+            initializeEncrypt(direct);
+        } else {
+            initializeDecrypt(direct);
+        }
+
+        // Perform first operation, either an update or dofinal.
+        if (updateFirst == true) {
+            doUpdate();
+        } else {
+            doFinal();
```
KostasTsiounis, line 82: Should we add a `return` here to avoid potentially wrong parameters?

src/test/java/ibm/jceplus/junit/base/BaseTestResetByteBuffer.java
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        // Perform second operation, either an update or dofinal.
+        if (updateSecond == true) {
+            doUpdate();
+        } else {
+            doFinal();
+        }
```
KostasTsiounis, line 90: Should we also include an additional `doFinal()` in case 2 updates were performed?

src/test/java/ibm/jceplus/junit/openjceplus/TestResetByteBuffer.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright IBM Corp. 2024
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution.
+ */
+
+package ibm.jceplus.junit.openjceplus;
```
KostasTsiounis, line 9: Does this test not apply to the FIPS provider too?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/AESCipher.java
jasonkatonica: Yes good point here. I was able to simplify this spot with commit https://github.com/IBM/OpenJCEPlus/pull/263/commits/64d8350ad66b6459c4c9e3629a90b9cf7194c1db since the methods `engineGetOutputSize` and also `engineDoFinal` both handle the logic associated with `use_z_fast_command`.

src/test/java/ibm/jceplus/junit/base/BaseTestResetByteBuffer.java
jasonkatonica: I don't believe that we want to do a return at this spot. A `doFinal()` operation should restore the state of the Cipher to the state it was prior to making the call. It should be initalized and ready for another operation ( either an update or doFinal ) after this doFinal() completes.

src/test/java/ibm/jceplus/junit/base/BaseTestResetByteBuffer.java
jasonkatonica: I don't believe that we want do finalize a double update. This test is just intended to make sure that no exceptions occur with various combinations of `doUpdate` and `doFinal` calls and that the output buffer sizes are calculated correctly.

src/test/java/ibm/jceplus/junit/openjceplus/TestResetByteBuffer.java
jasonkatonica: This did cross my mind that perhaps we wanted to test this with the OpenJCEPlusFIPS provider in addition to just OpenJCEPlus. The code through both these paths should be identical. This being said better safe then sorry so I turned this test on for OpenJCEPlusFIPS with updates in commit https://github.com/IBM/OpenJCEPlus/commit/64d8350ad66b6459c4c9e3629a90b9cf7194c1db

KostasTsiounis (approved):
LGTM

PR #269, opened 2024-10-18
Java 11 XDH private key parsing logic adjustment

jasonkatonica:
XDH encoded private keys can be of two different encoding formats:
1. [octet-string[octet-string[key-bytes]]] on Java 17 and higher.
2. [octet-string[key-bytes]] on Java 11 and lower.

When parsing the Java 11 version of the key encoding we currently incorrectly DER decode the key bytes assuming the Java 17 version of the key encoding format at all times. Occasionally the DER decoder fails to parse the non DER encoding and a failure with the message `Failed to create XEC private key` occurs.

This update attempts to parse the DER key bytes when we have already determined that the key bytes are of the Java 17 and higher key encoding format.


taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

PR #272, opened 2024-10-18
Use correct output size for AES on z platform

### LINKED ISSUE #240: Fix AES cipher output size issue for z/OS and LOZ

jasonkatonica:
When the fast z command is available the AES cipher does not return the correct output size on platform z.

Platform z uses a different buffer and output size calculation. The method `engineGetOutputSize` when running on platform z should return and take this into account.

A new test was added to exercise byte buffers and various combinations of encryption, decryption, and orders of doFinal and update operations which can be used to recreate the problem.

Closes #240

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESCipher.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESCipher.java
@@ -56,15 +56,9 @@ public final class AESCipher extends CipherSpi implements AESConstants {
         checkCipherInitialized();
 
         try {
-            byte[] output = null;
-            int outputLen = -1;
-            if (use_z_fast_command) {
-                output = new byte[engineGetOutputSize2(inputLen)];
-                outputLen = engineDoFinal(input, inputOffset, inputLen, output, 0);
-            } else {
-                output = new byte[engineGetOutputSize(inputLen)];
-                outputLen = symmetricCipher.doFinal(input, inputOffset, inputLen, output, 0);
-            }
+            byte[] output = new byte[engineGetOutputSize(inputLen)];
+            int outputLen = engineDoFinal(input, inputOffset, inputLen, output, 0);
+
             if (outputLen < output.length) {
                 byte[] out = Arrays.copyOfRange(output, 0, outputLen);
                 if (!encrypting) {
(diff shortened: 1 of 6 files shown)


KostasTsiounis (approved):
LGTM

jasonkatonica:
ppc64_aix, x86_64_linux, ppc64le_linux, s390x_linux, x86_64_windows

taoliult (approved):
LGTM

PR #294, opened 2024-10-28
Fix memory leak in Native SymmetricCipher KMC method

### LINKED ISSUE #293: Fix memory leak in the SymmetricCipher
https://github.com/IBM/OpenJCEPlus/blob/0464ecbf69b0b908f77372af37cdce40b550ed9f/src/main/native/SymmetricCipher.c#L455

This pattern in the `SymmetricCipher.c` causes memory leak:

```
inputPointer  = GetPrimitiveArrayCritical() + inputOffset;

CIPHER_XXX(inputPointer, ...);

ReleasePrimitiveArrayCritical(inputPointer);
```

Since the `inputPointer` was added by the `inputOffset`, then the release function can not release the correct memory and causes a memory leak.

That should be like this:
```
inputPointer  = GetPrimitiveArrayCritical();

CIPHER_XXX(inputPointer + inputOffset, ...);

ReleasePrimitiveArrayCritical(inputPointer);
```

jasonkatonica:
The memory address of the allocated buffer within the method `NativeInterface.z_kmc_native` was not correct. This resulted in a memory leak freeing the temporary input and output buffers.

Additional updates were made to ensure that if the `inputPointer` or the `outputPointer` failed to be allocated that memory already allocated was freed if this rare case was to occur.

Fixes #293

### DIFF
--- a/src/main/native/SymmetricCipher.c
+++ b/src/main/native/SymmetricCipher.c
@@ -449,21 +449,36 @@ JNIEXPORT int CIPHER_zKMC_internal(unsigned char* input, unsigned char* output, 
 
 JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_z_1kmc_1native(JNIEnv * env, jclass clazz, jbyteArray input, jint inputOffset, jbyteArray output, jint outputOffset, jlong paramPointer, jint inputLength, jint mode)
 {
-  // Get input and output buffer
-  jboolean isCopy = 0;
+  jboolean isCopy = JNI_FALSE;
   jint len = 0;
-  unsigned char* inputPointer  = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, input,  &isCopy)) + inputOffset;
-  unsigned char* outputPointer = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, output, &isCopy)) + outputOffset;
+  unsigned char* inputPointer  = NULL;
+  unsigned char* outputPointer = NULL;
 
-  if( NULL == outputPointer || NULL == inputPointer ) {
+  // Get input buffer.
+  inputPointer  = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, input,  &isCopy));
+  if (NULL == inputPointer) {
     throwOCKException(env, 0, "NULL from GetPrimitiveArrayCritical!");
-  }
-  else {
-      len = (jint)CIPHER_zKMC_internal(inputPointer, outputPointer, (int)inputLength, (long)paramPointer, (int)mode);
+    return -1;
   }
 
-  if(inputPointer != NULL) (*env)->ReleasePrimitiveArrayCritical(env, input, inputPointer, 0);
-  if(outputPointer != NULL) (*env)->ReleasePrimitiveArrayCritical(env, output, outputPointer, 0);
+  // Get output buffer.
+  outputPointer = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, output, &isCopy));
+  if (NULL == outputPointer) {
+    (*env)->ReleasePrimitiveArrayCritical(env, input, inputPointer, 0);
+    throwOCKException(env, 0, "NULL from GetPrimitiveArrayCritical!");
+    return -1;
+  }
+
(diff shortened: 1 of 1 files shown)


src/main/native/SymmetricCipher.c
```diff
@@ -449,21 +449,32 @@ JNIEXPORT int CIPHER_zKMC_internal(unsigned char* input, unsigned char* output,
...
 {
-  // Get input and output buffer
   jboolean isCopy = 0;
   jint len = 0;
-  unsigned char* inputPointer  = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, input,  &isCopy)) + inputOffset;
-  unsigned char* outputPointer = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, output, &isCopy)) + outputOffset;
 
-  if( NULL == outputPointer || NULL == inputPointer ) {
+  // Get input buffer.
+  unsigned char* inputPointer  = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, input,  &isCopy));
```
KostasTsiounis, line 456: Do we need the `isCopy` as a parameter, since it's just hardcoded a few lines above?

jasonkatonica: I dont mind having `isCopy` as a variable here given that if that were to change this code ( unlikely ) we would change it in one spot for both the input and output buffers. It also makes it a bit more readable knowing that the value represents if a copy should be made or not in the JNI `GetPrimitiveArrayCritical` calls. Do you see any advantage to hardcoding `&0` here?

KostasTsiounis: I don't think we've ever specified this as true in any of our code given that it's easier to pass a copy through Java if needed, rather than having JNI make a copy. Also, the VM might have to make a copy regardless of what you specify if arrays are in different format.

But, in any case, I'm OK with leaving the variable there. But could we at least declared it as `jboolean isCopy = JNI_FALSE` to make it more clear?

jasonkatonica: Yes agreed it is more clear to declare `isCopy` as `JNI_FALSE`. It makes this more readable. Update this using https://github.com/IBM/OpenJCEPlus/pull/294/commits/401a22e22da7b18d16537b840fb63bd16469c5de

KostasTsiounis (approved):
LGTM

src/main/native/SymmetricCipher.c
jasonkatonica: Also agree that typically we create the copy in the Java code not in JNI and that we dont really know for sure if the VM made a copy or not. We are just requesting that no copy is made here.

taoliult (approved):
LGTM

src/main/native/SymmetricCipher.c
```diff
@@ -449,21 +449,34 @@ JNIEXPORT int CIPHER_zKMC_internal(unsigned char* input, unsigned char* output,
...
-  // Get input and output buffer
-  jboolean isCopy = 0;
+  jboolean isCopy = JNI_FALSE;
   jint len = 0;
-  unsigned char* inputPointer  = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, input,  &isCopy)) + inputOffset;
-  unsigned char* outputPointer = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, output, &isCopy)) + outputOffset;
 
-  if( NULL == outputPointer || NULL == inputPointer ) {
+  // Get input buffer.
+  unsigned char* inputPointer  = (unsigned char*)((*env)->GetPrimitiveArrayCritical(env, input,  &isCopy));
```
johnpeck-us-ibm, line 456: All variable definitions should be at the top of the function. Make it easier to find them in a function if the function gets large.

johnpeck-us-ibm (changes requested):

src/main/native/SymmetricCipher.c
jasonkatonica: Yes this makes sense to do. Updated this code with https://github.com/IBM/OpenJCEPlus/pull/294/commits/c47900e797ce4b47f2bd22cc1d0c1ffd9e994868 to declare `inputPointer` and `outputPointer` at the top of the method.

jasonkatonica:
x86_64_linux,ppc64le_linux,s390x_linux,x86_64_windows

PR #300, opened 2024-10-30
Modify SymmetricCipher overlaps buffer checks

### LINKED ISSUE #292: Fix unary oparator in the SymmetricCipher
These two lines have the incorrect unary operator `=`.

That needs to be `+` --> `(outputOffset < (inputOffset + inputLen))`

https://github.com/IBM/OpenJCEPlus/blob/0464ecbf69b0b908f77372af37cdce40b550ed9f/src/main/java/com/ibm/crypto/plus/provider/ock/SymmetricCipher.java#L374

https://github.com/IBM/OpenJCEPlus/blob/0464ecbf69b0b908f77372af37cdce40b550ed9f/src/main/java/com/ibm/crypto/plus/provider/ock/SymmetricCipher.java#L504

jasonkatonica:
The `doFinal` and `update` methods in the SymmetricCipher class have logic to detect if a input buffer and output buffer are overlapping in any sort of way. If they are then a copy of the input is made to a separate location to ensure a safe operation can then take place on the data.

There is a logic problem where the calculation of overlap was incorrect and used a `=` instead of a `+` to calculate if the input and output overlapped. The comment above this code block was correct and remains unmodified.

The test `BaseTestAESGCMCopySafe` was modified to test this condition which it was attempting to do in the past however it was not operating upon input and output buffers that had the same memory address. The test was modified to exercise various overlapping conditions for AES GCM and also AES CBC accordingly. The test was then renamed to `BaseTestAESCopySafe` since we are now testing both GCM and CBC modes.

Fixes #292


jasonkatonica:

KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

PR #305, opened 2024-10-31
Fix memory leak in GCM.c and CCM.c

### LINKED ISSUE #304: Fix memory leak in GCM.c and CCM.c
https://github.com/IBM/OpenJCEPlus/blob/d8041a8c4a123b1f296db9973d4135e6952327aa/src/main/native/GCM.c#L1022

https://github.com/IBM/OpenJCEPlus/blob/d8041a8c4a123b1f296db9973d4135e6952327aa/src/main/native/GCM.c#L1026

https://github.com/IBM/OpenJCEPlus/blob/d8041a8c4a123b1f296db9973d4135e6952327aa/src/main/native/GCM.c#L1336

https://github.com/IBM/OpenJCEPlus/blob/d8041a8c4a123b1f296db9973d4135e6952327aa/src/main/native/GCM.c#L1340

The `inputOffset` and `outputOffset` should not be added to the pointers; otherwise, this causes a memory leak.

Those should be added to the pointers in the function calls

```
zS390((input != NULL) ? (input + inputOffset) : NULL,
          (output != NULL) ? (output + outputOffset) : NULL, ...
```

```
checkTagMismatch((input != NULL) ? (char*) (input + inputOffset) : NULL, ...
```

The CCM.c has the same issue.

https://github.com/IBM/OpenJCEPlus/blob/d8041a8c4a123b1f296db9973d4135e6952327aa/src/main/native/CCM.c#L252

https://github.com/IBM/OpenJCEPlus/blob/d8041a8c4a123b1f296db9973d4135e6952327aa/src/main/native/CCM.c#L256

https://github.com/IBM/OpenJCEPlus/blob/d8041a8c4a123b1f296db9973d4135e6952327aa/src/main/native/CCM.c#L486

https://github.com/IBM/OpenJCEPlus/blob/d8041a8c4a123b1f296db9973d4135e6952327aa/src/main/native/CCM.c#L490

gin-nader:
The `inputOffset` and `outputOffset` should not be added to the pointers; otherwise, this causes a memory leak.

Fixes #304


src/main/native/CCM.c
```diff
@@ -267,12 +265,12 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
         handleIV_CCM(ivLen, keyLen, blockSize, J0Offset, (char*) iv, (char*) (parm_block + keyOffset), (char*) parm_block);
         memcpy(parm_block + counterValueOffset, parm_block + J0Offset + blockSize - 4, 4); // Add Counter Value
 
-        CCMzS390(input, output, aad, &len, &alen, parm_block, &mode);
+        CCMzS390((input != NULL) ? (input + inputOffset) : NULL, (output != NULL) ? (output + outputOffset) : NULL, aad, &len, &alen, parm_block, &mode);
 
-        if(inputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, input, 0);
-        if(outputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, outputJ, output, 0);
+        if(inputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, (input != NULL) ? (input + inputOffset) : NULL, 0);
```
KostasTsiounis, line 270: Why does the `ReleasePrimitiveArrayCritical` have different parameters than the `GetPrimitiveArrayCritical`? Shouldn't we just release `input`?

Also, should we check `if(input != NULL)` instead of `if(inputJ != NULL)`?

src/main/native/CCM.c
```diff
@@ -267,12 +265,12 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
...
+        if(outputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, outputJ, (output != NULL) ? (output + outputOffset) : NULL, 0);
```
KostasTsiounis, line 271: Same as above.

src/main/native/CCM.c
```diff
@@ -267,12 +265,12 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
...
 
-        return checkTagMismatch_CCM((char*) input, len, (char*) parm_block, tagOffset, tagLen);
+        return checkTagMismatch_CCM((input != NULL) ? (char*) (input + inputOffset) : NULL, len, (char*) parm_block, tagOffset, tagLen);
```
KostasTsiounis, line 273: How do we use `input` if it has already been released?

src/main/native/CCM.c
```diff
@@ -501,14 +497,14 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
...
-        CCMzS390(input, output, aad, &len, &alen, parm_block, &mode);
+        CCMzS390((input != NULL) ? (input + inputOffset) : NULL, (output != NULL) ? (output + outputOffset) : NULL, aad, &len, &alen, parm_block, &mode);
 
         // Copy tag
         tag = parameters + ivLen + aadLen + keyLen;
         memcpy(tag, parm_block + 16, tagLen); // Add tag to output
 
-        if(inputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, input, 0);
-        if(outputJ) (*env)->ReleasePrimitiveArrayCritical(env, outputJ, output, 0);
+        if(inputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, (input != NULL) ? (input + inputOffset) : NULL, 0);
```
KostasTsiounis, line 506: Why does the `ReleasePrimitiveArrayCritical` have different parameters than the `GetPrimitiveArrayCritical`? Shouldn't we just release `input`?

Also, should we check `if(input != NULL)` instead of `if(inputJ != NULL)`?

src/main/native/CCM.c
```diff
@@ -501,14 +497,14 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
...
+        if(outputJ) (*env)->ReleasePrimitiveArrayCritical(env, outputJ, (output != NULL) ? (output + outputOffset) : NULL, 0);
```
KostasTsiounis, line 507: Same as above.

src/main/native/GCM.c
```diff
@@ -136,99 +136,99 @@ int GCM_InitForUpdateDecrypt_core(JNIEnv *env, ICC_CTX *ockCtx, ICC_AES_GCM_CTX
...
-	
+
 	static const char *functionName = "NativeInterface.GCM_InitForUpdateDecrypt_core";
 
 	if ( debug ) {
 		gslogFunctionEntry(functionName);
 	}
-	
-  
+
```
KostasTsiounis, line 145: Why 2 empty lines?

src/main/native/GCM.c
```diff
@@ -136,99 +136,99 @@ int GCM_InitForUpdateDecrypt_core(JNIEnv *env, ICC_CTX *ockCtx, ICC_AES_GCM_CTX
...
     	    		gslogMessage ("DETAIL_GCM NI calling ICC_AES_GCM_DecryptUpdate for AAD\n");
     			}
-#endif			    
-			    
+#endif
+
 				rc = ICC_AES_GCM_DecryptUpdate(ockCtx, gcmCtx,aad, aadLen,NULL, 0,NULL, &updateAADlen);
-			   
-			
+
```
KostasTsiounis, line 179: Why 2 empty lines?

src/main/native/GCM.c
```diff
@@ -248,76 +248,76 @@ int GCM_UpdForUpdateDecrypt_core(JNIEnv *env, ICC_CTX *ockCtx, ICC_AES_GCM_CTX *
...
 	static const char *functionName = "NativeInterface.GCM_UpdForUpdateDecrypt_core";
 
 	if ( debug ) {
 		gslogFunctionEntry(functionName);
 	}
 	*updateOutlen = 0;
-	
-   
+
```
KostasTsiounis, line 258: Why 2 empty lines?

src/main/native/GCM.c
```diff
@@ -248,76 +248,76 @@ int GCM_UpdForUpdateDecrypt_core(JNIEnv *env, ICC_CTX *ockCtx, ICC_AES_GCM_CTX *
...
     	    	gslogMessage ("DETAIL_GCM NI checking for data > 0\n");
     		}
-#endif	
-		
+#endif
+
 		if (dataLen > 0) {
-			
-    		
+
```
KostasTsiounis, line 282: Why 2 empty lines?

src/main/native/GCM.c
```diff
@@ -379,10 +379,10 @@ int GCM_InitForUpdateEncrypt_core(JNIEnv* env, ICC_CTX* ockCtx, ICC_AES_GCM_CTX*
     	    gslogMessage ("DETAIL_GCM NI GCM_InitForUpdateEncrypt_core updateOutlen %ld\n", *updateOutlen);
     	   	gslogMessage ("DETAIL_GCM N NI first update calling getOrfreeGCMCtx\n");
     	}
-#endif	
-   
-    
-	
+#endif
+
+
```
KostasTsiounis, line 384: Why 3 empty lines?

src/main/native/GCM.c
```diff
@@ -1037,12 +1035,12 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
 	handleIV(ivLen, keyLen, blockSize, J0Offset, (char*) iv, (char*) (parm_block + keyOffset), (char*) parm_block);
 	memcpy(parm_block + counterValueOffset, parm_block + J0Offset + blockSize - 4, 4); // Add Counter Value
 
-	zS390(input, output, aad, &len, &alen, parm_block, &mode);
-	
-	if(inputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, input, 0);
-	if(outputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, outputJ, output, 0);
+	zS390((input != NULL) ? (input + inputOffset) : NULL, (output != NULL) ? (output + outputOffset) : NULL, aad, &len, &alen, parm_block, &mode);
+
+	if(inputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, (input != NULL) ? (input + inputOffset) : NULL, 0);
```
KostasTsiounis, line 1040: Why does the `ReleasePrimitiveArrayCritical` have different parameters than the `GetPrimitiveArrayCritical`? Shouldn't we just release `input`?

Also, should we check `if(input != NULL)` instead of `if(inputJ != NULL)`?

src/main/native/GCM.c
```diff
@@ -1037,12 +1035,12 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
...
+	if(outputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, outputJ, (output != NULL) ? (output + outputOffset) : NULL, 0);
```
KostasTsiounis, line 1041: Same as above.

src/main/native/GCM.c
```diff
@@ -1037,12 +1035,12 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
...
 
-	return checkTagMismatch((char*) input, len, (char*) parm_block, tagOffset, tagLen);
+	return checkTagMismatch((input != NULL) ? (char*) (input + inputOffset) : NULL, len, (char*) parm_block, tagOffset, tagLen);
```
KostasTsiounis, line 1043: How do we access `input` if it has already been released?

src/main/native/GCM.c
```diff
@@ -1346,19 +1342,19 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
...
-	zS390(input, output, aad, &len, &alen, parm_block, &mode);
+	zS390((input != NULL) ? (input + inputOffset) : NULL, (output != NULL) ? (output + outputOffset) : NULL, aad, &len, &alen, parm_block, &mode);
 
 	// Copy tag
 	tag = parameters + ivLen + aadLen + keyLen;
 	memcpy(tag, parm_block + 16, tagLen); // Add tag to output
 
-	if(inputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, input, 0);
-	if(outputJ) (*env)->ReleasePrimitiveArrayCritical(env, outputJ, output, 0);
+	if(inputJ != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, (input != NULL) ? (input + inputOffset) : NULL, 0);
```
KostasTsiounis, line 1356: Why does the `ReleasePrimitiveArrayCritical` have different parameters than the `GetPrimitiveArrayCritical`? Shouldn't we just release `input`?

Also, should we check `if(input != NULL)` instead of `if(inputJ != NULL)`?

src/main/native/GCM.c
```diff
@@ -1346,19 +1342,19 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
...
+	if(outputJ) (*env)->ReleasePrimitiveArrayCritical(env, outputJ, (output != NULL) ? (output + outputOffset) : NULL, 0);
```
KostasTsiounis, line 1357: Same as above.

KostasTsiounis (changes requested):

src/main/native/CCM.c
jasonkatonica, line 270: For spots like this with if statements that we have touched think we could use brackets to make it explicit what is in the if statement, ie the format :
```
if(inputJ != NULL) {
    (*env)->ReleasePrimitiveArrayCritical(env, inputJ, (input != NULL) ? (input + inputOffset) : NULL, 0);
}
```
Applies also to a few other if statements that were changed that did not use brackets.

jasonkatonica (changes requested):

johnpeck-us-ibm (approved):

src/main/native/CCM.c
gin-nader: I'm not sure what the difference is between the two methods.

There is a check for both `if(input != NULL)` and `if(inputJ != NULL)`.

src/main/native/GCM.c
gin-nader: I'm not sure what the difference is between the two methods.

There is a check for both `if(input != NULL)` and `if(inputJ != NULL)`.

src/main/native/CCM.c
KostasTsiounis: I mean that I think it should be `if(input != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, input, 0);`, since the get and release should be applied to the same arguments, not one with offset.

src/main/native/CCM.c
KostasTsiounis: I mean that I think it should be 
`if(input != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, input, 0);`
, since the get and release should be applied to the same arguments, not one with offset.

src/main/native/GCM.c
KostasTsiounis: I mean that I think it should be 
`if(input != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, input, 0);`
, since the get and release should be applied to the same arguments, not one with offset.

src/main/native/GCM.c
KostasTsiounis: I mean that I think it should be 
`if(input != NULL) (*env)->ReleasePrimitiveArrayCritical(env, inputJ, input, 0);`
, since the get and release should be applied to the same arguments, not one with offset.

src/main/native/CCM.c
gin-nader: Okay, perhaps I'm missing something here then. The `GetPrimitiveArrayCritical` returns `input`, and does not take it as an argument. In the original code, by the time `ReleasePrimitiveArrayCritical` is called, `input` already has the offset applied. If we don't include the offset here, then that would be a functional change to the code.

KostasTsiounis: The offset is applied to read starting from a different part of the array rather than the start. But the full array is loaded and we get a pointer in `input`. So, when releasing that array, the full one should be released. At least to my understanding, the `inputOffset` is just for the reading part. The array, however, should be acquired and released as a whole.

I don't think it will affect anything if `input` is not a copy of the original array, but `GetPrimitiveArrayCritical` cannot always ensure that that's the case. So, if it is a copy we need to free the whole array and not just the part that starts at `inputOffset`.

@jasonkatonica Do you have any further insight on this? Maybe I'm wrong on the use of

jasonkatonica: From what i can tell it does seem like there is another potential leak here in the case where `input` really created a copy when `GetPrimitiveArrayCritical` was called above. I thihnk we should be freeing all memory obtained by `GetPrimitiveArrayCritical` represented by the pointer input if the value is not null. There is other code spinkled through out the other C code that follows a pattern like this instead of trying to free at an offset of the input and output pointers.

src/main/native/CCM.c
jasonkatonica: Yes agreed i think we want to check if input is NULL then free `input` it seems more simple that way.

src/main/native/CCM.c
gin-nader: Good point. This was likely working before but this definitely leaves room for possible memory corruption.

gin-nader:
I believe all comments have been addressed. Please review again.

jasonkatonica (approved):
LGTM

taoliult (approved):
LGTM

JinhangZhang (approved):
LGTM

jasonkatonica:

PR #318, opened 2024-11-04
Allow debug compilation to pass and add DH detail

jasonkatonica:
When turning on all available debug information available with the c library via compilation options in the various makefiles certain code has been found to not compile. This update fixes compilation errors primarily associated with incorrectly named variables and also castings required to call helper methods to print debug messages.

This update also adds additional debug statements to the DH method `DHKEY.generate` since relevant information for other known issues was missing.

### DIFF
--- a/src/main/native/CCM.c
+++ b/src/main/native/CCM.c
@@ -375,8 +375,8 @@ int CCM_encrypt_core(JNIEnv* env, ICC_CTX* ockCtx,
     unsigned char* iv    , int ivLen,
     unsigned char* aad   , int aadLen,
     int tagLen,
-    unsigned char* plain , int plaintextLen,
-    unsigned char* cipher, unsigned long  ciphertextLen) {
+    unsigned char* plainText , int plaintextLen,
+    unsigned char* cipherText, unsigned long  ciphertextLen) {
     int             rc               = ICC_OSSL_SUCCESS;
     static const char * functionName = "NativeInterface.CCM_encrypt_core";
 
@@ -385,20 +385,20 @@ int CCM_encrypt_core(JNIEnv* env, ICC_CTX* ockCtx,
     }
 #ifdef DEBUG_CCM_DATA
     if ( debug ) {
-        gslogMessagePrefix ("DATA_CCM ivNative : ");
-        gslogMessageHex (ivNative, 0, (int) ivLen, 0, 0, NULL);
+        gslogMessagePrefix ("DATA_CCM iv : ");
+        gslogMessageHex ((char*) iv, 0, (int) ivLen, 0, 0, NULL);
 
-        gslogMessagePrefix ("DATA_CCM keyNative : ");
-        gslogMessageHex (keyNative, 0, (int) keyLen, 0, 0, NULL);
+        gslogMessagePrefix ("DATA_CCM key : ");
+        gslogMessageHex ((char*) key, 0, (int) keyLen, 0, 0, NULL);
 
-        gslogMessagePrefix ("DATA_CCM iphertextNative : ");
-        gslogMessageHex (plaintextNative, 0, (int) plaintextLen, 0, 0, NULL);
+        gslogMessagePrefix ("DATA_CCM plainText : ");
+        gslogMessageHex ((char*) plainText, 0, (int) plaintextLen, 0, 0, NULL);
 
-        gslogMessagePrefix ("DATA_CCM aadNative : ");
-        gslogMessageHex (aadNative, 0, (int) aadLen, 0, 0, NULL);
+        gslogMessagePrefix ("DATA_CCM aad : ");
+        gslogMessageHex ((char*) aad, 0, (int) aadLen, 0, 0, NULL);
     }
 #endif
-    rc = ICC_AES_CCM_Encrypt(ockCtx, iv, ivLen, key, keyLen, aad, aadLen, plain, plaintextLen, cipher, &ciphertextLen, tagLen);
+    rc = ICC_AES_CCM_Encrypt(ockCtx, iv, ivLen, key, keyLen, aad, aadLen, plainText, plaintextLen, cipherText, &ciphertextLen, tagLen);
 
     if (rc != ICC_OSSL_SUCCESS) {
         ockCheckStatus(ockCtx);
--- a/src/main/native/ECKey.c
+++ b/src/main/native/ECKey.c
@@ -1129,7 +1129,7 @@ JNIEXPORT jlong JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_XE
     if ( debug ) {
       gslogMessage ("DATA_XEC size=%d", (int) size);
       gslogMessagePrefix ("DATA_EC privateKeyBytes : ");
-      gslogMessageHex (pBytes, 0, size, 0, 0, NULL);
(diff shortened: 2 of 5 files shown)


src/main/native/CCM.c
```diff
@@ -385,17 +385,17 @@ int CCM_encrypt_core(JNIEnv* env, ICC_CTX* ockCtx,
...
+        gslogMessageHex ((char*)iv, 0, (int) ivLen, 0, 0, NULL);
 
-        gslogMessagePrefix ("DATA_CCM keyNative : ");
-        gslogMessageHex (keyNative, 0, (int) keyLen, 0, 0, NULL);
+        gslogMessagePrefix ("DATA_CCM key : ");
+        gslogMessageHex ((char*)key, 0, (int) keyLen, 0, 0, NULL);
 
-        gslogMessagePrefix ("DATA_CCM iphertextNative : ");
-        gslogMessageHex (plaintextNative, 0, (int) plaintextLen, 0, 0, NULL);
+        gslogMessagePrefix ("DATA_CCM plain : ");
```
KostasTsiounis, line 394: Maybe `plainText` would be a better name?

src/main/native/CCM.c
```diff
@@ -385,17 +385,17 @@ int CCM_encrypt_core(JNIEnv* env, ICC_CTX* ockCtx,
     }
 #ifdef DEBUG_CCM_DATA
     if ( debug ) {
-        gslogMessagePrefix ("DATA_CCM ivNative : ");
-        gslogMessageHex (ivNative, 0, (int) ivLen, 0, 0, NULL);
+        gslogMessagePrefix ("DATA_CCM iv : ");
+        gslogMessageHex ((char*)iv, 0, (int) ivLen, 0, 0, NULL);
```
KostasTsiounis, line 389: Should we leave a space after the cast (i.e., `(char *) iv` to make it more readable and be consistent with the one right after?

Same comment goes for all the cast spots.

src/main/native/GCM.c
```diff
@@ -1539,7 +1539,7 @@ JNIEXPORT jint JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_do_
 
 #ifdef DEBUG_GCM_DETAIL
     if ( debug ) {
-        gslogMessage ("NI do_GCM_UpdForUpdateEncrypt cls %x\n", cls);
+        gslogMessage ("NI do_GCM_UpdForUpdateEncrypt thisObj %x\n", thisObj);
```
KostasTsiounis, line 1542: In this case the JNI passes the class, not the object I think. So, the name is a bit confusing. In other JNI code we specify the parameter as `jclass clazz`.

KostasTsiounis (changes requested):

src/main/native/CCM.c
jasonkatonica: Yes a space after the cast is more consistent for this code. Updated to add a space using https://github.com/IBM/OpenJCEPlus/pull/318/commits/b7cf2a703fa95e7b6755333943a19860a02db8cc

src/main/native/CCM.c
jasonkatonica: In the various debugs I would prefer to use the name of the variables instead of using `plainText` and plain wording for the debugs as it makes it easier to search through the code later looking for what was printed in a given debug log. In this case the variable `plainText` does not exist in this method.

src/main/native/GCM.c
jasonkatonica: Looking into this some there are various documents including [this one ](https://docs.oracle.com/en/java/javase/22/docs/specs/jni/design.html) that state something like the following:

---

_The second argument differs depending on whether the native method is static or non-static. The second argument to a non-static native method is a reference to the object. The second argument to a static native method is a reference to its Java class._

---

So I believe that you are right here that the reference is really a reference to a Java class not the object itself given that all our methods in `NativeInterface.java` are all declared as static. 

For consistancy reasons it seems we could just leave this as `thisObj` given that almost all the JNI code in the project uses `thisObj` instead of `clazz`. If we want to make the change for this variable name perhaps we could do this in another PR that changes all variables from `thisObj` to `clazz`.

src/main/native/CCM.c
KostasTsiounis: If you ask me, the variable should also be named `plainText`, as `plain` is not really accurate. Would it make sense to change it, or should we just leave it for another PR?

src/main/native/GCM.c
KostasTsiounis: Sure. We could open another PR to fix those kind of things.

src/main/native/CCM.c
jasonkatonica: OK yes agreed that the variable names `plainText` and `cipherText` are better names then `plain` and `cipher` that is in use in this method now. I updated the names using commit https://github.com/IBM/OpenJCEPlus/pull/318/commits/b0541ed0c25c119cf7c159dfe68393f745687886

KostasTsiounis (approved):
LGTM

jasonkatonica:
x86_64_linux,ppc64le_linux,s390x_linux,x86_64_windows

src/main/native/ECKey.c
```diff
@@ -1129,7 +1129,7 @@ JNIEXPORT jlong JNICALL Java_com_ibm_crypto_plus_provider_ock_NativeInterface_XE
     if ( debug ) {
       gslogMessage ("DATA_XEC size=%d", (int) size);
       gslogMessagePrefix ("DATA_EC privateKeyBytes : ");
-      gslogMessageHex (pBytes, 0, size, 0, 0, NULL);
+      gslogMessageHex ((char*) pBytes, 0, size, 0, 0, NULL);
```
taoliult, line 1132: `pBytes` defined as a `unsigned char *`, do we need/have to cast it to just a `char*` ?

jasonkatonica: Hello, yes the compilation fails if I dont cast to `(char*)`. Our other debug code within `#ifdef` value also casts to `(char*)`.

taoliult: ok, I got it.

taoliult (approved):
LGTM

src/main/native/GCM.c
```diff
@@ -1141,16 +1141,16 @@ int GCM_encrypt_core(JNIEnv* env, ICC_CTX* ockCtx, ICC_AES_GCM_CTX* gcmCtx,
 #ifdef DEBUG_GCM_DATA
     if ( debug ) {
         gslogMessagePrefix ("DATA_GCM ivNative : ");
-        gslogMessageHex (ivNative, 0, (int) ivLen, 0, 0, NULL);
+        gslogMessageHex ((char*) ivNative, 0, (int) ivLen, 0, 0, NULL);
```
JinhangZhang, line 1144: shouldnt we also try to get the message for `(char*)iv` instead of `(char*)ivNative`?

jasonkatonica: On line 1126 `iv` is set to `ivNative` so at this point in the code I think they are the same value. This being said we are also debugging and making use of the variable `ivNative` on line 1167 below so it seems most accurate to print the value of `ivNative` here. Unless there is a different path I am missing where printing the value of `iv` is helpful?

JinhangZhang: makes sense

JinhangZhang (approved):
LGTM

PR #333, opened 2024-11-08
Add FIPS platforms check

taoliult:
Print the error message indicating that OpenJCEPlusFIPS is not supported on the non FIPS platforms, but do not exit.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
@@ -20,6 +20,7 @@ import java.security.Provider;
 import java.security.ProviderException;
 import java.security.PublicKey;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import javax.crypto.SecretKey;
@@ -67,12 +68,49 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
     private static boolean ockInitialized = false;
     private static OCKContext ockContext;
 
+    private static final boolean isPlatformSupported;
+    private static final Map<String, List<String>> supportedPlatforms = new HashMap<>();
+    private static final String osName;
+    private static final String osArch;
+
+    static {
+        supportedPlatforms.put("Arch", List.of("amd64", "ppc64", "s390x"));
+        supportedPlatforms.put("OS", List.of("Linux", "AIX", "Windows"));
+
+        osName = System.getProperty("os.name");
+        osArch = System.getProperty("os.arch");;
+
+        boolean isOsSupported, isArchSupported;
+        // Check whether the OpenJCEPlus FIPS is supported.
+        isOsSupported = false;
+        for (String os: supportedPlatforms.get("OS")) {
+            if (osName.contains(os)) {
+                isOsSupported = true;
+                break;
+            }
+        }
+        isArchSupported = false;
+        for (String arch: supportedPlatforms.get("Arch")) {
+            if (osArch.contains(arch)) {
+                isArchSupported = true;
+                break;
+            }
+        }
+        isPlatformSupported = isOsSupported && isArchSupported;
+    }
+
     @SuppressWarnings({"unchecked", "rawtypes"})
     public OpenJCEPlusFIPS() {
         super("OpenJCEPlusFIPS", info);
         if (debug != null) {
             debug.println("New OpenJCEPlusFIPS instance");
         }
+
+        if (!isPlatformSupported) {
+            throw new UnsupportedOperationException(
+                        "OpenJCEPlusFIPS is not supported on this non FIPS " + osName + " " + osArch + " platform");
+        }
+
         final OpenJCEPlusProvider jce = this;
 
         AccessController.doPrivileged(new java.security.PrivilegedAction() {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,7 +7,7 @@
  */
 
 import groovy.json.JsonOutput;
-import groovy.transform.Field
+import groovy.transform.Field;
 
 @Field boolean PPC64_AIX
 @Field boolean X86_64_LINUX


src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -67,12 +67,37 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
...
+    private static final boolean isPlatformSupported;
+    private static final List<String> supportedPlatforms = List.of("Linux-amd64", "Linux-s390x",
+            "Linux-ppc64le", "Windows-amd64", "AIX-ppc64");
+
+    static {
+        String osName = System.getProperty("os.name");
+        String osArch = System.getProperty("os.arch");;
+        String currentPlatform = osName + '-' + osArch;
+        
+        if (supportedPlatforms.contains(currentPlatform)) {
```
KostasTsiounis, line 79: This could just be `isPlatformSupported = supportedPlatforms.contains(currentPlatform);`

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -67,12 +67,37 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
...
     @SuppressWarnings({"unchecked", "rawtypes"})
     public OpenJCEPlusFIPS() {
         super("OpenJCEPlusFIPS", info);
         if (debug != null) {
             debug.println("New OpenJCEPlusFIPS instance");
         }
+
+        if (!isPlatformSupported) {
+            // Print out the exception but not exit.
+            new ProviderException("OpenJCEPlusFIPS is not supported on this non FIPS platform").printStackTrace();
```
KostasTsiounis, line 95: Shouldn't we throw this exception?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
jasonkatonica: Yes i agree i think we were planning on throwing an exception and letting the JCE framework handle this. Currently that means it will be ignored is my understanding preventing the provider to load "silently".

taoliult: I checked the OpenJDK code, the ProviderCongfig.java. When exceptions occur during provider initialization, if it is the `ProviderExceptionn`, it will cause the process to stop. For other types of exceptions, the stack trace will be printed only if debugging is enabled, and then the process will continue.

I haven't checked the exact how the exceptions are thrown from our OpenJCEPlusFIPS provider, but through testing, I found, no matter the `ProviderException` or another exception, such as `UnsupportedOperationException`, when the exception is thrown, the stack trace is printed, but the process will always exits.

KostasTsiounis: It seems like if it's anything other than a `ProviderException` it just doesn't load the provider and moves on to the next, which is what we want here I think. So, I guess we just need to throw another type of exception there.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
taoliult: Done.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
taoliult: I did, I tried for example `UnsupportedOperationException`, but the process still exits. I am using the mvn build and test for checking if the OpenJCEPlusFIPS can be loaded or not. My understanding is that it will still going to the ProviderConfig to load the provider and get its instance, so the exception should be catched in ProviderConfig. But not sure why the process still exits.

KostasTsiounis: Trying through mvn won't work I think, because it has to be as part of a module that exposes an instance of `java.security.Provider` for `ProviderConfig` to pick it up. That's why you can't see it there.

taoliult: Updated the codes to throw the `UnsupportedOperationException` if the platform is not supported. Without debug, nothing will be printed out and the OpenJCEPlus won't be loaded (Can't be listed when using `Security.getProviders()`). With enable the `jca` debug trace `(-Djava.security.debug=jca)`, the following error will be printed from `ProviderConfig`: 

```
ProviderConfig: Error loading legacy provider com.ibm.crypto.plus.provider.OpenJCEPlusFIPS
java.lang.UnsupportedOperationException: OpenJCEPlusFIPS is not supported on this non FIPS Linux amd64 platform
	at openjceplus/com.ibm.crypto.plus.provider.OpenJCEPlusFIPS.<init>(OpenJCEPlusFIPS.java:110)
	at java.base/java.lang.J9VMInternals.newInstanceImpl(Native Method)
	at java.base/java.lang.Class.newInstance(Class.java:2379)
	at java.base/sun.security.jca.ProviderConfig$ProviderLoader$1.run(ProviderConfig.java:431)
	at java.base/sun.security.jca.ProviderConfig$ProviderLoader$1.run(ProviderConfig.java:428)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:748)
	at java.base/sun.security.jca.ProviderConfig$ProviderLoader.legacyLoad(ProviderConfig.java:428)
	at java.base/sun.security.jca.ProviderConfig$ProviderLoader.load(ProviderConfig.java:393)
	at java.base/sun.security.jca.ProviderConfig$3.run(ProviderConfig.java:270)
	at java.base/sun.security.jca.ProviderConfig$3.run(ProviderConfig.java:264)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:692)
	at java.base/sun.security.jca.ProviderConfig.doLoadProvider(ProviderConfig.java:264)
	at java.base/sun.security.jca.ProviderConfig.getProvider(ProviderConfig.java:242)
	at java.base/sun.security.jca.ProviderList.loadAll(ProviderList.java:342)
	at java.base/sun.security.jca.ProviderList.removeInvalid(ProviderList.java:359)
	at java.base/sun.security.jca.Providers.getFullProviderList(Providers.java:212)
	at java.base/java.security.Security.getProviders(Security.java:462)
	at ListProviderAlgorithms.main(ListProviderAlgorithms.java:9)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:575)
	at jdk.compiler/com.sun.tools.javac.launcher.Main.execute(Main.java:419)
	at jdk.compiler/com.sun.tools.javac.launcher.Main.run(Main.java:192)
	at jdk.compiler/com.sun.tools.javac.launcher.Main.main(Main.java:132)
ProviderConfig: Error loading provider com.ibm.crypto.plus.provider.OpenJCEPlusFIPS
```

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

jasonkatonica (commented):
LGTM

PR #339, opened 2024-11-13
Use non-bundled OCK library in AIX

KostasTsiounis:
If there is a version of `OCK` that is bundled with the utilized `JDK`, it is always preferred over what the current run provides. To avoid that, the bundled version is deleted.

`AIX` is, also, enabled to be in the default platforms that are run.


jasonkatonica (approved):
LGTM

taoliult (approved):
LGTM

PR #344, opened 2024-11-14
Add FIPS platforms check

taoliult:
Print the error message indicating that OpenJCEPlusFIPS is not supported on the non FIPS platforms, but do not exit.


KostasTsiounis (approved):
LGTM

jasonkatonica (approved):
LGTM

PR #348, opened 2024-11-15
Remove 31/32 bit references and unsupported envs

jasonkatonica:
This update removes 32 bit references that are currently not in working or testable state along with references to operating systems and architectures that are not supported by OpenJCEPlus.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
@@ -37,6 +37,7 @@ final class NativeInterface {
     // Default ock core library name
     //
     private static final String OCK_CORE_LIBRARY_NAME = "jgsk8iccs";
+    private static final String JGSKIT_CORE_LIBRARY_NAME = "jgskit";
     private static String osName = null;
     private static String osArch = null;
     private static String JVMFIPSmode = null;
@@ -49,12 +50,12 @@ final class NativeInterface {
                     // ensure we are picking up the expected version within
                     // the JRE.
                     //
-                    preloadOCK("jgsk8iccs", true);
+                    preloadOCK();
                 }
 
                 // Load native code for java-gskit
                 //
-                preloadJGskit("jgskit");
+                preloadJGskit();
                 return null;
             }
         });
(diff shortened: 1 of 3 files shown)


src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
```diff
@@ -131,140 +131,50 @@ static void preloadJGskit(String libraryToLoad) {
...
         File loadFile = null;
-        if (osName.startsWith("Windows")) {
-            if (osArch.equals("x86")) {
-                // win32_x86
-                loadFile = new File(jgskitPath, "libjgskit.dll");
-            } else {
-                // win64_x86
-                loadFile = new File(jgskitPath, "libjgskit_64.dll");
-            }
+        if (osName.startsWith("Windows") && osArch.equals("amd64")) {
```
KostasTsiounis, line 134: The way the code was before, it would get this dll if OS is `Windows` and anything other than `x86`. Is there anything else architecture-wise that we could get here and want to account for?

src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
```diff
@@ -131,140 +131,50 @@ static void preloadJGskit(String libraryToLoad) {
...
-        // sun64_x86: lib<ockCoreLibraryName>_64.so
-        // win32_x86: <ockCoreLibraryName>.dll
-        // win64_x86: <ockCoreLibraryName>_64.dll
+        // AIX: lib<libraryToLoad>_64.so
+        // Linux aarch64: lib<libraryToLoad>.dylib
+        // Linux ppc64le: lib<libraryToLoad>_64.so
+        // Linux s390x: lib<libraryToLoad>_64.so
+        // Linux x86_64: lib<libraryToLoad>_64.so
+        // Mac OS X: lib<libraryToLoad>_64.so
+        // Windows* amd64: <libraryToLoad>_64.dll
```
KostasTsiounis, line 164: Is there a reason for the `*` after `Windows`?

src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
```diff
@@ -131,140 +131,50 @@ static void preloadJGskit(String libraryToLoad) {
...
-            }
-
-            throw new ProviderException(exceptionMessage);
+            throw new ProviderException("Could not load dependent jgskit library for os.name=" + osName
+                        + ", os.arch=" + osArch);
         }
     }
 
-    static void preloadOCK(String libraryToLoad, boolean add64) {
+    static void preloadOCK(String libraryToLoad) {
```
KostasTsiounis, line 149: I'm not even sure why we pass `libraryToLoad` as a parameter. Is the method ever called with a different option?

src/main/native/jgskit.mak
```diff
@@ -23,69 +23,35 @@ ifeq (${PLATFORM},arm-linux64)
...
   OSINCLUDEDIR=aix
-else ifeq (${PLATFORM},ppc-linux32)
-  PLAT=xp
-  CFLAGS+= -m32 -DLINUX
-  LDFLAGS+= -m32
-else ifeq (${PLATFORM},ppc-linux64)
-  PLAT=xp
-  CFLAGS+= -DLINUX
-  LDFLAGS+= -m64
-  IS64SYSTEM=64
```
KostasTsiounis, line 46: Would it make sense to remove the `IS64SYSTEM` everywhere, since they are all 64bit ones now?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
jasonkatonica: At the moment i think all we support are windows for 64 bit ( `amd64` arch ). I dont think we support any other windows environments. If the `osName` was to start with `Windows` and we are running on some other architecture then we would get the error on new line 145 just below. Perhaps I am missing a path here that could lead to an error?

src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
jasonkatonica: The only reason I put `*` in this spot is that the code currently checks for the os name to start with Windows, so this is already represented by the comment where we use `Windows*`

src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
jasonkatonica: Good point. Updated the code here to use a constant since we always pass the same value into this method. The argument was completely removed.

src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
KostasTsiounis: No, nothing specific comes to mind. I was just wondering if there might be one. But I'll resolve this since you can't think of anything either.

src/main/native/jgskit.mak
jasonkatonica: Yes agreed that it makes sense to remove `IS64SYSTEM` as this is no longer needed as everything is 64 bit.

src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
jasonkatonica: Updated with https://github.com/IBM/OpenJCEPlus/pull/348/commits/97153a9c2dcb73771278d1c69605df14dd31a0b4

src/main/native/jgskit.mak
jasonkatonica: Updated with https://github.com/IBM/OpenJCEPlus/pull/348/commits/97153a9c2dcb73771278d1c69605df14dd31a0b4

src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
KostasTsiounis: Oh, so with the `*` you mean that it begins with `Windows`, but there might be other stuff after. Makes sense.

KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

JinhangZhang (approved):
LGTM

PR #352, opened 2024-11-19
Use latest mvn dependency versions

jasonkatonica:
This is a routine update of all the maven dependency versions.

The junit jupiter family of dependency versions now no longer uses a pom wide property to define its version. This is to make it easier to automatically update these fields using the command `mvn versions:use-latest-versions` which automates this. Using a property disrupts this from working.


taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

PR #367, opened 2024-12-03
Avoid memory leak associated with symmetric init

jasonkatonica:
The `NativeInterface.CIPHER_init` method initializes a given context such that it is ready to perform encryption and decryption for a given Cipher.

As part of the initialization method a `ockCipher->cached_context` can be optionally created in addition to the context being used. While allocating the `ockCipher->cached_context` the logic did not account for any previously created contexts that were previously stored in the `ockCipher->cached_context` location. This caused a leak of context objects for each initialization done on objects that already contained a context.

Additional error checking was done for the calls to `ICC_EVP_CIPHER_CTX_new` and `ICC_EVP_CIPHER_CTX_copy` to ensure that the method worked as expected.


KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

JinhangZhang (approved):
LGTM

PR #364, opened 2024-11-27
Use appropriate OID in RSA key generation

KostasTsiounis:
When generating `RSA` keys, the type of key needs to be checked in order to differentiate between `legacy RSA` and `RSAPSS` and select the appropriate `OID`.

Fixes: https://github.com/IBM/OpenJCEPlus/issues/365

Signed-off-by: Kostas Tsiounis <kostas.tsiounis@ibm.com>

PR #369, opened 2024-12-04
Use appropriate OID in RSA key generation

KostasTsiounis:
When generating `RSA` keys, the type of key needs to be checked in order to differentiate between `legacy RSA` and `RSAPSS` and select the appropriate `OID`.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/364


jasonkatonica (approved):
LGTM

PR #372, opened 2024-12-04
Implement calculatePublicKey() in classes that extend PKCS8Key

KostasTsiounis:
`PKCS8Key` implements the `InternalPrivateKey` interface that contains the `calculatePublicKey()` method. If not implemented, an `UnsupportedOperationException` is thrown. This functionality is implemented to support this operation instead of getting an exception.

### DIFF
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestECKeyImportInterop.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestECKeyImportInterop.java
@@ -24,11 +24,51 @@ import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.Arrays;
 import org.junit.jupiter.api.Test;
+import sun.security.util.InternalPrivateKey;
 import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 
 public class BaseTestECKeyImportInterop extends BaseTestJunit5Interop {
 
     static final byte[] origMsg = "this is the original message to be signed".getBytes();
+
+    @Test
+    public void testCreateKeyPairECGenParamImportCalculatePublic() throws Exception {
+        doCreateKeyPairECGenParamImportCalculatePublic(getProviderName(), getInteropProviderName());
+        doCreateKeyPairECGenParamImportCalculatePublic(getInteropProviderName(), getProviderName());
+    }
+
+    private void doCreateKeyPairECGenParamImportCalculatePublic(String generateProviderName,
+            String importProviderName) throws Exception {
+
+        KeyPairGenerator keyPairGen = KeyPairGenerator.getInstance("EC", generateProviderName);
+
+        keyPairGen.initialize(256);
+        KeyPair keyPair = keyPairGen.generateKeyPair();
+        PrivateKey privateKey = keyPair.getPrivate();
+        PublicKey publicKey = keyPair.getPublic();
+
+        // Recreate private key from encoding.
+        byte[] privKeyBytes = privateKey.getEncoded();
+        KeyFactory keyFactory = KeyFactory.getInstance("EC", importProviderName);
+        EncodedKeySpec privateKeySpec = new PKCS8EncodedKeySpec(privKeyBytes);
+        privateKey = keyFactory.generatePrivate(privateKeySpec);
+
+        // Get public key bytes from private.
+        byte[] calculatedPublicKey = ((InternalPrivateKey) privateKey).calculatePublicKey().getEncoded();
+
+        // Get public key bytes from original public key.
+        byte[] publicKeyBytes = publicKey.getEncoded();
+
+        System.out.println("---- Comparing EC public key from KeyPair vs calculated from private key ----");
+        System.out.println("EC public key from Keypair from " + generateProviderName + ": "
+                + BaseUtils.bytesToHex(publicKeyBytes));
+        System.out.println("EC public key from calculatePublicKey() from " + importProviderName + ": "
+                + BaseUtils.bytesToHex(calculatedPublicKey));
+
+        // The original and calculated public keys should be the same
+        assertArrayEquals(calculatedPublicKey, publicKeyBytes);
+    }
 
     @Test
     public void testCreateKeyPairECGenParamImport() throws Exception {
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestXDHInterop.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestXDHInterop.java
@@ -11,6 +11,8 @@ package ibm.jceplus.junit.base;
 import java.security.KeyFactory;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
+import java.security.PrivateKey;
+import java.security.PublicKey;
 import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.EncodedKeySpec;
 import java.security.spec.NamedParameterSpec;
@@ -18,6 +20,8 @@ import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.Base64;
 import org.junit.jupiter.api.Test;
+import sun.security.util.InternalPrivateKey;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 
 public class BaseTestXDHInterop extends BaseTestJunit5Interop {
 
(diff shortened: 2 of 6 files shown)


jasonkatonica (commented):
Code looks correct to me. Should we add either a known answer test or a interop test for both of these new methods?

KostasTsiounis:
> Code looks correct to me. Should we add either a known answer test or a interop test for both of these new methods?

I could add a test where a `KeyPair` is created and then a public key is calculated using this method in the private key. At the end we'll be checking if the calculated public key and the one from the `KeyPair` are the same. Does that make sense?

KostasTsiounis:
I have added tests for this new method with 0040a314e857771e2f6e2d1349dfeaec9fd364f8. Let me know what you think.

johnpeck-us-ibm (changes requested):
Hi Kostas, The RFC for EC keys indicates that the Public Key portion of a private key is optional. So, you will not always be able get the public key from a private key and that should be allowed. The issue is that you throw an exception when you do not get the public key. We should only throw an exception when the public key is present, but we can not calculate the key correctly.

KostasTsiounis:
> Hi Kostas, The RFC for EC keys indicates that the Public Key portion of a private key is optional. So, you will not always be able get the public key from a private key and that should be allowed. The issue is that you throw an exception when you do not get the public key. We should only throw an exception when the public key is present, but we can not calculate the key correctly.

Sure. The public key doesn't have to be part of the encoding. But, to my understanding, the public key can always be calculated based on the private key. Isn't that the case @johnpeck-us-ibm?

I'm not sure though if that happens they way I have it now. I will check it out and make the changes if needed.

KostasTsiounis:
So, I checked into the creation of the public key a bit more. @johnpeck-us-ibm is right that according to the RFC, the private key doesn't have to contain the public key. After checking, I discovered that the `SunEC` implementation does exactly that, so I created the following test to see what happens:
```
// Create keys with the SunEC implementation.
KeyPairGenerator keyPairGen = KeyPairGenerator.getInstance("EC", "SunEC");
keyPairGen.initialize(256);
KeyPair keyPair = keyPairGen.generateKeyPair();
PrivateKey privateKey = keyPair.getPrivate();
PublicKey publicKey = keyPair.getPublic();
// Recreate private key from encoding.
byte[] privKeyBytes = privateKey.getEncoded();
KeyFactory keyFactory = KeyFactory.getInstance("EC", providerName);
EncodedKeySpec privateKeySpec = new PKCS8EncodedKeySpec(privKeyBytes);
privateKey = keyFactory.generatePrivate(privateKeySpec);
// Get public key bytes from private.
byte[] calculatedPublicKey = ((InternalPrivateKey) privateKey).calculatePublicKey().getEncoded();
// Get public key bytes from original public key.
byte[] publicKeyBytes = publicKey.getEncoded();

System.out.println("---- Comparing EC public key from KeyPair vs calculated from private key ----");
System.out.println("EC private key: " + BaseUtils.bytesToHex(privKeyBytes));
System.out.println("EC public key from Keypair: " + BaseUtils.bytesToHex(publicKeyBytes));
System.out.println("EC public key from calculatePublicKey(): " + BaseUtils.bytesToHex(calculatedPublicKey));
// The original and calculated public keys should be the same
Assert.assertArrayEquals(calculatedPublicKey, publicKeyBytes);
```
The test does the following:

1. A keypair is created using the `SunEC` implementation. The private key encoding is obtained. After running the test, an example of an encoding I got would be as follows:

`3041020100301306072a8648ce3d020106082a8648ce3d030107042730250201010420269ccb5f5b846989ee4470837708173737edf90b16f30ab137f06736b1f293d8`, which would be translated into:
```
SEQUENCE {
   INTEGER 0x00 (0 decimal)
   SEQUENCE {
      OBJECTIDENTIFIER 1.2.840.10045.2.1 (ecPublicKey)
      OBJECTIDENTIFIER 1.2.840.10045.3.1.7 (P-256)
   }
   OCTETSTRING 30250201010420269ccb5f5b846989ee4470837708173737edf90b16f30ab137f06736b1f293d8
}
```
Further translating the `OCTETSTRING 30250201010420269ccb5f5b846989ee4470837708173737edf90b16f30ab137f06736b1f293d8` gives us:
```
SEQUENCE {
   INTEGER 0x01 (1 decimal)
   OCTETSTRING 269ccb5f5b846989ee4470837708173737edf90b16f30ab137f06736b1f293d8
}
```
We can see that the public key is not contained in the decoding.

2. I, then, import the encoding through a key factory using the `OpenJCEPlus` implementation to create a private key. 
3. The newly calculated private key is used to produce the public key.
4. The original public key is compared to the calculated key and are found to be the same.

Trying to figure out why this happens, I discovered that when we are trying to create the `ECKey *` from an encoding, we use the `ICC_d2i_ECPrivateKey()` method, which is equivalent to the OpenSSL's `d2i_ECPrivateKey()` method. Looking further into the source code, we can see [here](https://github.com/openssl/openssl/blob/b372b1f76450acdfed1e2301a39810146e28b02c/crypto/ec/ec_asn1.c#L1067) that, if a public key is not present, it is calculated on the spot.

So, I think we are covered in the case where the public key is not part of the private key encoding.

@johnpeck-us-ibm @jasonkatonica Does that make sense to you?

src/test/java/ibm/jceplus/junit/base/BaseTestECKeyImportInterop.java
```diff
@@ -23,13 +23,54 @@
...
+    @Test
+    public void testCreateKeyPairECGenParamImportCalculatePublic() throws Exception {
+        doCreateKeyPairECGenParamImportCalculatePublic(getProviderName(), getInteropProviderName());
+        doCreateKeyPairECGenParamImportCalculatePublic(getInteropProviderName(), getProviderName());
+    }
+
+    public void doCreateKeyPairECGenParamImportCalculatePublic(String generateProviderName,
+            String importProviderName) throws Exception {
+
+        //final String methodName = "testECGenParamImportHardcoded";
```
jasonkatonica, line 44: Could this string be removed that is commented out? I dont see any refs to it that we would potentially use in the future for debugs.

src/test/java/ibm/jceplus/junit/base/BaseTestXDHInterop.java
```diff
@@ -29,6 +33,48 @@ public class BaseTestXDHInterop extends BaseTestJunit5Interop {
     String openJDK_private_X448 = "MEYCAQAwBwYDK2VvBQAEOOJFsgLYxgAIEWuN1FLAGWDzGQRSataAbPLDc1wv5aky4T8hevyWbYdhggc1OCcqQ93gY8rqVTDb";
     // OpenJDK does not currently support FFDHE hence interop testing for FFDHE is not possible
 
+    @Test
+    public void testCreateKeyPairXDHGenParamImportCalculatePublic() throws Exception {
+        if (!"BC".equals(getInteropProviderName())) {
```
jasonkatonica, line 38: Why no BC interop here? Do we get a different answer from BC for the encoded key?

src/test/java/ibm/jceplus/junit/base/BaseTestXDHInterop.java
```diff
@@ -29,6 +33,48 @@ public class BaseTestXDHInterop extends BaseTestJunit5Interop {
...
+            doCreateKeyPairXDHGenParamImportCalculatePublic(getProviderName(), getInteropProviderName());
+            doCreateKeyPairXDHGenParamImportCalculatePublic(getInteropProviderName(), getProviderName());
+        }
+        
+    }
+
+    public void doCreateKeyPairXDHGenParamImportCalculatePublic(String generateProviderName,
+            String importProviderName) throws Exception {
+
+        //final String methodName = "testECGenParamImportHardcoded";
```
jasonkatonica, line 48: Similar question as above any reason to keep this string?

src/test/java/ibm/jceplus/junit/base/BaseTestXDHKeyPairGenerator.java
```diff
@@ -10,11 +10,15 @@
 
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
+import java.security.PrivateKey;
+import java.security.PublicKey;
 import java.security.interfaces.XECPrivateKey;
 import java.security.interfaces.XECPublicKey;
 import java.security.spec.NamedParameterSpec;
+import org.junit.Assert;
```
jasonkatonica, line 18: Please use `org.junit.jupiter.api` instead of the older assert.

src/test/java/ibm/jceplus/junit/base/BaseTestXDHInterop.java
KostasTsiounis: `BouncyCastle` doesn't offer the same APIs, so interop is not available.

src/test/java/ibm/jceplus/junit/base/BaseTestECKeyImportInterop.java
KostasTsiounis: Removed.

src/test/java/ibm/jceplus/junit/base/BaseTestXDHInterop.java
KostasTsiounis: Removed.

src/test/java/ibm/jceplus/junit/base/BaseTestXDHKeyPairGenerator.java
KostasTsiounis: Switched to new API.

src/test/java/ibm/jceplus/junit/base/BaseTestXDHInterop.java
jasonkatonica: OK makes sense

jasonkatonica (approved):
LGTM

JinhangZhang (approved):
LGTM

taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

PR #377, opened 2024-12-06
Fix Digest clone memory leak

jasonkatonica:
Cloning message digests in tight loops has been observed to leak even under conditions that are not expected to leak such as single threaded workload tests.

This update makes use of creating a newly allocated Digest object then populating this object with fields from the original Digest being cloned. Leaks in native memory are no longer observed after this update.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
@@ -205,6 +205,9 @@ public final class Digest implements Cloneable {
         //OCKDebug.Msg(debPrefix, methodName,  "digestAlgo :" + digestAlgo);
     }
 
+    private Digest() {
+    }
+
     static void throwOCKException(int errorCode) throws OCKException {
         //final String methodName = "throwOCKExeption";
         // OCKDebug.Msg(debPrefix, methodName, "throwOCKException errorCode =  " + errorCode);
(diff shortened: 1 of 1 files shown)


src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
```diff
@@ -340,15 +343,28 @@ protected static boolean validId(long id) {
...
+        copy.ockContext = this.ockContext;
         copy.contextFromQueue = false;
+
+        // Allocate a new context for the digestId and copy all state information from our
+        // original context into the copy. 
         try {
             copy.digestId = NativeInterface.DIGEST_copy(
                 this.ockContext.getId(), getId());
-            if (copy.digestId == 0) {
+            if (0 == copy.digestId) {
```
KostasTsiounis, line 367: We use this convention of putting the constant first in the C code. Is that something we want to do in Java code as well?

KostasTsiounis (changes requested):

taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
jasonkatonica: Yes good question. In general I do think its a good idea for us to typically put the constant first. The main argument I have heard over time is that the compile will prevent mistakes such as forgetting the `==` and instead putting a `=` into the if statement. Using a constant ensures that the compile fails if we were to do `if (0 = copy.digestId)` by mistake. Same thinking also applies to the Java code. Id say we should fix these as we go along with changes.

KostasTsiounis (approved):
LGTM

PR #386, opened 2024-12-09
Update README.md

johnpeck-us-ibm:
Added AI Policy to indicate that what to do with AI generated code and where it can be used in this project.


README.md
```diff
@@ -352,4 +352,22 @@ The following contribution guidelines should be followed:
...
+
+   - You MUST validate that the use of the GenAI tool does not produce protected code (i.e. code that is copyrighted or licensed outside of our project's current licenses).
+
+    An example of a tagging would be:
+
+    /* Start code generated by (GenAI tool) */
+
+ 
+
+    /* End code  generated by (GenAI tool) */
```
jasonkatonica, line 373: Ideally this code block should be formatted as a code block using three tics "```" to surround all these lines which should format it.

README.md
```diff
@@ -352,4 +352,22 @@ The following contribution guidelines should be followed:
 
 1. Code should be styled according to the included [style.xml](style.xml) eclipse rules.
 
-1. A pull request should be sent for review only after the github action associated with this repository is automatically executed when a pull request is created.
+2. A pull request should be sent for review only after the github action associated with this repository is automatically executed when a pull request is created.
```
jasonkatonica, line 355: Markdown should allow for using the value `1.` through this list and it will be rendered as different values automatically. Could you please try using `1.` to replace both the values `2.` and `3.`?

jasonkatonica (changes requested):
The commit message and PR description do not match. These should be adjusted so the text in the commit matches the PR description.

The PR description also does not contain the same `Signed-off-by` value as the commit.

PR #416, opened 2025-01-06
Update ECDSASignature to match expected behaviour

KostasTsiounis:
The method `engineSetParameter` that gets an `AlgorithmParameterSpec` as a parameter is implemented, thus overriding the existing version that just throws an `UnsupportedOperationException`.

The `engineSetParameter` that gets `String` and `Object` parameters is updated to match the behaviour of similar implementations from other providers.

`Override` annotations are added to specific methods to enforce checks.

A few more test cases are added to check the new functionality.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/ECDSASignature.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ECDSASignature.java
@@ -10,12 +10,16 @@ package com.ibm.crypto.plus.provider;
 
 import com.ibm.crypto.plus.provider.ock.Signature;
 import java.security.AlgorithmParameters;
+import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.InvalidParameterException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.SignatureException;
 import java.security.SignatureSpi;
+import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.ECParameterSpec;
+import sun.security.util.ECUtil;
 import sun.security.util.ObjectIdentifier;
 
 
@@ -23,6 +27,8 @@ abstract class ECDSASignature extends SignatureSpi {
 
     private OpenJCEPlusProvider provider = null;
     private Signature signature = null;
+    private ECPrivateKey privateKey = null;
+    private ECPublicKey publicKey = null;
 
     ECDSASignature(OpenJCEPlusProvider provider, String ockDigestAlgo) {
         try {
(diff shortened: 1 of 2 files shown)


src/main/java/com/ibm/crypto/plus/provider/ECDSASignature.java
```diff
@@ -109,11 +115,33 @@ protected boolean engineVerify(byte[] sigBytes) throws SignatureException {
...
+        java.security.interfaces.ECKey key = (this.privateKey == null? this.publicKey : this.privateKey);
+        if ((key != null) && !ECUtil.equals(ecparams, key.getParams())) {
+            throw new InvalidAlgorithmParameterException(
+                    "Signature params does not match key params");
+        }
     }
 
     @Deprecated
+    @Override
     protected Object engineGetParameter(String param) throws InvalidParameterException {
```
jasonkatonica, line 145: Seems getting parameters always returns null is this normal behavior for OpenJDK providers also?

KostasTsiounis: This actually throws an `UnsupportedOperationException` in the OpenJDK. I didn't change that behaviour to avoid problems with existing users, but I can update that to mirror the OpenJDK behaviour.

jasonkatonica: I would like to mimic OpenJDK crypto providers in this case. You are correct that a caller today is getting a `null` back from this API then doing nothing with the value ( given that null is not usable in replacement of an actual Object being returned here ). I cant imagine a case where they get the null back, do nothing with it, and their application proceeds. I think it makes most sense to throw an exception alerting callers of this issue.

johnpeck-us-ibm: Maybe I am missing something, but since we are able to set Parameters we need to be able to return them with this API. I do not think null or an exception is acceptable.

KostasTsiounis: The following comment is in the `OpenJDK` code for this method:
```java
// get parameter, not supported. See JCA doc
```

And this is there for the other `getParameter()` method that has `AlgorithmParameters` as return value:
```java
// Always return null even if setParameter is called before.
// According to RFC 3279 2.2.3 and RFC 5758 3.2, no parameters is
// defined for ECDSA AlgorithmIdentifiers.
```

It looks like neither of those is supposed to be working for `ECDSASignature` and they're there just due to interface requirements.

@johnpeck-us-ibm Do you think we should diverge from the `OpenJDK` logic?

KostasTsiounis: I have updated the code to throw an exception similar to the `OpenJDK` code, but let me know if you think we should go our own way on this.

jasonkatonica (approved):
LGTM

JinhangZhang (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/ECDSASignature.java
```diff
@@ -109,13 +115,35 @@ protected boolean engineVerify(byte[] sigBytes) throws SignatureException {
     }
 
     @Deprecated
-    protected void engineSetParameter(String param, Object value) throws InvalidParameterException {
-        throw new InvalidParameterException("No parameter accepted");
+    @Override
+    protected void engineSetParameter(String param, Object value)
```
taoliult, line 119: I saw this `engineSetParameter(String param, Object value)` is previously marked as `@Deprecated`. Any reason why we change it to `@Override`?

KostasTsiounis: It still has the `@Deprecated` tag right above. I just added the `@Override`.

taoliult (approved):
LGTM

PR #423, opened 2025-01-15
Update the DH key size check to include the exponent size check

taoliult:
Fixes https://github.com/IBM/OpenJCEPlus/issues/418

Move the checkKeyLengths() method from DHKeyFactory to DHKeyPairGenerator, rename it to checkKeySize(), and add a check for the exponent size, same as OpenJDK did.

Signed-off-by: Tao Liu <tao.liu@ibm.com>

PR #426, opened 2025-01-16
Update the DH key size check to include the exponent size check

### LINKED ISSUE #418: Change the current DH key length check to consider exp size as well
The `checkKeyLengths` in `DHKeyFactory` should be removed

https://github.com/IBM/OpenJCEPlus/blob/1518b17929c2160faf19e76506b766ddcc99321e/src/main/java/com/ibm/crypto/plus/provider/DHKeyFactory.java#L248C17-L248C32

FIPS static parameters should be moved to the DHKeyPairGenerator

https://github.com/IBM/OpenJCEPlus/blob/1518b17929c2160faf19e76506b766ddcc99321e/src/main/java/com/ibm/crypto/plus/provider/DHKeyFactory.java#L33

The check key length should be added in the DHKeyPairGenerator as below

It also requires to check for the optional expSize as well (0 is passed in case that alg parameter is not specified in the initialize method)
```
static void checkKeySize(int keySize, int minSize, int maxSize, int expSize)
            throws InvalidParameterException {

        if ((keySize < minSize) || (keySize > maxSize) || ((keySize & 0x3F) != 0)) {
            throw new InvalidParameterException(
                    "DH key size must be multiple of 64, and can only range " +
                    "from 512 to 8192 (inclusive). " +
                    "The specific key size " + keySize + " is not supported");
        }

        // optional, could be 0 if not specified
        if ((expSize < 0) || (expSize > keySize)) {
            throw new InvalidParameterException
                    ("Exponent size must be positive and no larger than" +
                    " modulus size");
        }
    }
```

The check without alg parameters
```
if( provider.isFIPS() ) {
        checkKeySize(keySize, MIN_KEYSIZE_FIPS, MAX_KEYSIZE_FIPS ,0);
} else {
        checkKeySize(keySize, MIN_KEYSIZE_NONFIPS, MAX_KEYSIZE_NONFIPS ,0);
}
```

with alg parameter:
```
if( provider.isFIPS() ) {
	checkKeySize(keySize, MIN_KEYSIZE_FIPS, MAX_KEYSIZE_FIPS , ((DHParameterSpec) algParams).getL());
} else {
	checkKeySize(keySize, MIN_KEYSIZE_NONFIPS, MAX_KEYSIZE_NONFIPS , ((DHParameterSpec) algParams).getL());
}
```

taoliult:
This is a back-port PR from PR https://github.com/IBM/OpenJCEPlus/pull/423

Move the checkKeyLengths() method from DHKeyFactory to DHKeyPairGenerator, rename it to checkKeySize(), and add a check for the exponent size, same as OpenJDK did.

Fixes #418

Signed-off-by: Tao Liu <tao.liu@ibm.com>

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/DHKeyFactory.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/DHKeyFactory.java
@@ -10,7 +10,6 @@ package com.ibm.crypto.plus.provider;
 
 import java.io.IOException;
 import java.security.InvalidKeyException;
-import java.security.InvalidParameterException;
 import java.security.Key;
 import java.security.KeyFactorySpi;
 import java.security.PrivateKey;
@@ -30,10 +29,6 @@ import javax.crypto.spec.DHPublicKeySpec;
 public final class DHKeyFactory extends KeyFactorySpi {
 
     private OpenJCEPlusProvider provider;
-    public final static int MIN_KEYSIZE_NONFIPS = 512;
-    public final static int MAX_KEYSIZE_NONFIPS = 8192;
-    public final static int MIN_KEYSIZE_FIPS = 2048;
-    public final static int MAX_KEYSIZE_FIPS = 8192;
 
     static DHKey toDHKey(OpenJCEPlusProvider provider, Key key) throws InvalidKeyException {
         return (DHKey) new DHKeyFactory(provider).engineTranslateKey(key);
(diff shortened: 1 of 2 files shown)


KostasTsiounis (approved):
LGTM

jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/DHKeyFactory.java
```diff
@@ -225,35 +220,4 @@ protected Key engineTranslateKey(Key key) throws InvalidKeyException {
...
-            throws InvalidParameterException {
-
-        if ((keysize < minsize) || (keysize > maxsize) || ((keysize & 0x3F) != 0)) {
-            throw new InvalidParameterException(
-                    "DH key size must be multiple of 64, and can only range "
-                            + "from 512 to 8192 (inclusive). " + "The specific key size " + keysize
-                            + " is not supported");
-        }
-    }
-
```
johnpeck-us-ibm, line 258: Not sure why there are no checks on key sizes in this code. There is the possibility that the keys created here will not be usable by the needed provider with out these.

johnpeck-us-ibm (changes requested):

src/main/java/com/ibm/crypto/plus/provider/DHKeyFactory.java
taoliult: The check key size method named `checkKeyLengths()` moved from this `DHKeyFactory` to `DHKeyPairGenerator`, and renamed as `checkKeySize()`, same as the OpenJDK did. 

Or, you mean, there is no key sizes check for those keys used/created in this DHKeyFactory class?

taoliult: Right, there are no key checks in the `DHKeyFactory` class. The previous `checkKeyLengths()` method in `DHKeyFactory` only be used by the `DHKeyPairGenerator` class. So, it moved from `DHKeyFactory` to `DHKeyPairGenerator` in this PR.

I would suggest, we open another issue to track the discussion and activities for the `"no key checks in the DHKeyFactory class"` issue, and merge this current PR.

taoliult: I opened an issue https://github.ibm.com/runtimes/jit-crypto/issues/653 to track the discussion about if the key size checks are needed in the DHKeyFactory class

johnpeck-us-ibm (approved):

PR #432, opened 2025-01-22
Add support for PBKDF2 algorithms

jasonkatonica:
The following algorithms will be added to the OpenJCEPlusFIPS provider:
- PBKDF2WithHmacSHA224
- PBKDF2WithHmacSHA256
- PBKDF2WithHmacSHA384
- PBKDF2WithHmacSHA512

The following algorithms will be added to the OpenJCEPlus provider:
- PBKDF2WithHmacSHA1
- PBKDF2WithHmacSHA224
- PBKDF2WithHmacSHA256
- PBKDF2WithHmacSHA384
- PBKDF2WithHmacSHA512

Updates required include:
- Adding a new PBKDF2Core class based on OpenJDK.
- Adding a new PBKDF2KeyImpl class based on OpenJDK yet modified to make use of the OCKC library to perform PBKDF2 key derivations.
- New JNI code to call a new native method to derive a key using PBKDF2 from a given password, salt, iteration count, and desired key length.
- Updates in makefiles to allow the new PBKDF.c file to be built into the JNI dll.
- A new set of interoperability test was created to enforce that we get the same results from SunJCE, OpenJCEPlus, OpenJCEPlusFIPS for the KeyFactory methods `generateSecret`, `translateKey`, and `getKeySpec`.
- A new test was introduce to drive various KAT and error paths.


src/test/java/ibm/jceplus/junit/base/BaseTestPBKDF2.java
```diff
@@ -0,0 +1,198 @@
...
+    * a1 d4 25 a1 22 58 33 54 9a db 84 1b 51 c9 b3 17
+    * 6a 27 2b de bb a1 d0 78 47 8f 62 b3 97 f3 3c 8d
+    *
+    * @param algorithm
+    * @throws Exception
+    */
+    @Test
+    public void testPBKDF2KAT() throws Exception {
+        //The KAT published salt value is not acceptable for FIPS, skipt this test.
+        if (this.getProviderName().equalsIgnoreCase("OpenJCEPlusFIPS")) {
```
KostasTsiounis, line 65: Not sure if it's better, but we could do:
```java
@Disabled("The KAT published salt value is not acceptable for FIPS.")
@Override
public void testPBKDF2KAT() throws Exception {}
```

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
```diff
@@ -0,0 +1,272 @@
...
+                    throw new InvalidKeySpecException("Iteration count must be 1000 or higher when using the OpenJCEPlusFIPS provider.");
+                }
+
+                // Password length must be 10 characters or more.
+                if (this.passwd.length < 10) {
+                    throw new InvalidKeySpecException("Password must be 10 characters or higher when using the OpenJCEPlusFIPS provider.");
+                }
+            }
+
+            this.prf = Mac.getInstance(prfAlgo, provider);
```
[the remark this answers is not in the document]
jasonkatonica: This line was changed and now we only save the `prfAlgo` instead of instantiating a MAC without using it according to [another comment here](https://github.com/IBM/OpenJCEPlus/pull/432#discussion_r1928789599) which makes sense to me.

src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
```diff
@@ -0,0 +1,272 @@
...
+
+    /**
+     * Calculates a hash code value for the object.
+     * Objects that are equal will also have the same hashcode.
+     */
+    @Override
+    public int hashCode() {
+        try {
+            return Arrays.hashCode(this.key)
+                    ^ getAlgorithm().toLowerCase(Locale.ENGLISH).hashCode();
```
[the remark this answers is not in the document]
jasonkatonica: This section of code calculates the hash of the object the same way in which [OpenJDK based class](https://github.com/ibmruntimes/openj9-openjdk-jdk24/blob/openj9/src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java#L279) does.  The `^` is a xor operation so that does seem intentional.

src/main/native/PBKDF.c
```diff
@@ -0,0 +1,183 @@
...
+#endif
+        throwOCKException(env, 0,
+                          "Memory allocation failed for key derivation");
+        goto cleanup;
+    }
+
+    // Execute PBKDF2 key derivation
+    rc = ICC_PKCS5_PBKDF2_HMAC(
+        ockCtx, (const char *)passwordNative, passwordLength, saltNative,
+        saltLength, iterations, messageDigest, keyLength, derivedKey);
```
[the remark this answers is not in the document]
jasonkatonica: I changed this code to avoid the extra copy made and was able to make use of the allocated `resultDerivedKey`.

src/main/native/PBKDF.c
```diff
@@ -0,0 +1,183 @@
...
+        }
+#endif
+        throwOCKException(env, 0,
+                          "Memory allocation failed for key derivation");
+        goto cleanup;
+    }
+
+    // Execute PBKDF2 key derivation
+    rc = ICC_PKCS5_PBKDF2_HMAC(
+        ockCtx, (const char *)passwordNative, passwordLength, saltNative,
```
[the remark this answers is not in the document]
jasonkatonica: I changed this code to avoid the extra copy made and was able to make use of the allocated `resultDerivedKey` for the comment below. I also changed how the variable was declared to avoid this cast in this spot calling `ICC_PKCS5_PBKDF2_HMAC`. I did however need to change the cast on the release of `passwordNative` to `(void *)`. Hopefully this reads better now.

src/test/java/ibm/jceplus/junit/base/BaseTestPBKDF2.java
jasonkatonica: This KAT can be run fine when using non fips cryptography. So it still makes sense to enable this test for that case. When running with FIPS cryptography some of the arguments ( password length requirements ) do not meet the standard of 10 characters as documented in security policy.

KostasTsiounis: Yes. I meant have the above code in the `OpenJCEPlusFIPS` equivalent of `TestPBKDF2`, so that the test is only disabled in FIPS mode.

src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
```diff
@@ -0,0 +1,267 @@
...
+    @java.io.Serial
+    private static final long serialVersionUID = -2234868909660948157L;
+
+    private char[] passwd;
+    private byte[] salt;
+    private final int iterCount;
+    private byte[] key;
+
+    // The following fields are not Serializable. See writeReplace method.
+    private transient final String prf;
```
KostasTsiounis, line 51: I don't think a `String` needs to be transient. Also, should we call it `prfAlgo` to avoid confusion?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
jasonkatonica: Agreed seems like this String does not need to be transient now. Please see https://github.com/IBM/OpenJCEPlus/pull/432/commits/67f2097c37619ebaf3aa73288423c4174fdafeba for latest changes in this area.

src/test/java/ibm/jceplus/junit/base/BaseTestPBKDF2.java
jasonkatonica: Yes i think this is prob an easier way to seperate how the FIPS and non FIPS version of this tests runs. Please see https://github.com/IBM/OpenJCEPlus/pull/432/commits/67f2097c37619ebaf3aa73288423c4174fdafeba for latest changes in this area.

KostasTsiounis (approved):
LGTM

taoliult (approved):
LGTM

PR #410, opened 2024-12-19
Eliminate older junit dependency

jasonkatonica:
The older junit dependency is still configured in the pom.xml mvn file as a dependency. This update removes this dependency and migrates all codes to junit jupiter framework. In particular the following changes were made.

1. The `pom.xml` file was updated to remove the junit dependency. It was replaced by just the hamcrest portion of the features that are used in the project.

1. All imports are now being done from the `org.junit.jupiter.api` package instead of `org.junit`.

1. Various assertions required the message string to be last argument to the method instead of first.

Signed-off-by: Jason Katonica <katonica@us.ibm.com>

PR #435, opened 2025-02-03
Eliminate older junit dependency

jasonkatonica:
The older junit dependency is still configured in the pom.xml mvn file as a dependency. This update removes this dependency and migrates all codes to junit jupiter framework. In particular the following changes were made.

1. The `pom.xml` file was updated to remove the junit dependency. It was replaced by just the hamcrest portion of the features that are used in the project.

1. All imports are now being done from the `org.junit.jupiter.api` package instead of `org.junit`.

1. Various assertions required the message string to be last argument to the method instead of first.

Back-ported from: #410

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -548,9 +548,9 @@
             <version>3.5.2</version>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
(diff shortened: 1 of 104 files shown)


KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

PR #438, opened 2025-02-03
Update ECDSASignature to match expected behaviour

KostasTsiounis:
The method `engineSetParameter` that gets an `AlgorithmParameterSpec` as a parameter is implemented, thus overriding the existing version that just throws an `UnsupportedOperationException`.

The `engineSetParameter` that gets `String` and `Object` parameters is updated to match the behaviour of similar implementations from other providers.

`Override` annotations are added to specific methods to enforce checks.

A few more test cases are added to check the new functionality.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/416


jasonkatonica (approved):
LGTM

taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

PR #448, opened 2025-02-07
Remove the use of the SecurityManager

jasonkatonica:
As described in [JEP 486](https://openjdk.org/jeps/486) the SecurityManager is being disabled in Java 24. This update removes all use of the SecurityManager class and privileged actions.


src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
```diff
@@ -39,7 +36,7 @@ public final class Digest implements Cloneable {
     //disable caching mechanism for windows OS
     final static private boolean isWindows = System.getProperty("os.name").startsWith("Windows");
 
-    final static private int numContexts;
+    static private int numContexts;
```
KostasTsiounis, line 39: Any reason we removed the `final` here?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/ock/Digest.java
KostasTsiounis: Or we could declare another variable in the static block and only assign to `numContexts` at the end. I think it's a good idea to keep this `final` in order to avoid potential errors where it is set to something else.

jasonkatonica: > Or we could declare another variable in the static block and only assign to `numContexts` at the end.

Sure sounds reasonable to me Kostas. The code has been recently updated to do this hopefully it reads ok to you.

KostasTsiounis (approved):
LGTM

taoliult (approved):
LGTM

PR #450, opened 2025-02-07
Initialize FastJNIBuffer only when use_z_fast_command is set to true

taoliult:
FastJNIBuffer is initialized every time a SymmetricCipher instance is created, regardless of whether it is used or not. Move the initialization of FastJNIBuffer to occur only when use_z_fast_command is set to true.


johnpeck-us-ibm (approved):

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

PR #469, opened 2025-02-21
Add support for PBKDF2 algorithms

jasonkatonica:
The following algorithms will be added to the OpenJCEPlusFIPS provider:
- PBKDF2WithHmacSHA224
- PBKDF2WithHmacSHA256
- PBKDF2WithHmacSHA384
- PBKDF2WithHmacSHA512

The following algorithms will be added to the OpenJCEPlus provider:
- PBKDF2WithHmacSHA1
- PBKDF2WithHmacSHA224
- PBKDF2WithHmacSHA256
- PBKDF2WithHmacSHA384
- PBKDF2WithHmacSHA512

Updates required include:
- Adding a new PBKDF2Core class based on OpenJDK.
- Adding a new PBKDF2KeyImpl class based on OpenJDK yet modified to make use of the OCKC library to perform PBKDF2 key derivations.
- New JNI code to call a new native method to derive a key using PBKDF2 from a given password, salt, iteration count, and desired key length.
- Updates in makefiles to allow the new PBKDF.c file to be built into the JNI dll.
- A new set of interoperability test was created to enforce that we get the same results from SunJCE, OpenJCEPlus, OpenJCEPlusFIPS for the KeyFactory methods `generateSecret`, `translateKey`, and `getKeySpec`.
- A new test was introduce to drive various KAT and error paths.

Back-ported from: #432

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
@@ -1,0 +1,260 @@
+/*
+ * Copyright IBM Corp. 2025
+ *
+ * Licensed under the Apache License 2.0 (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution.
+ */
+
+package com.ibm.crypto.plus.provider;
+
+import com.ibm.crypto.plus.provider.ock.OCKException;
+import com.ibm.crypto.plus.provider.ock.PBKDF;
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
+import java.io.ObjectStreamException;
+import java.lang.ref.Reference;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.security.KeyRep;
+import java.security.MessageDigest;
+import java.security.spec.InvalidKeySpecException;
+import java.util.Arrays;
+import java.util.Locale;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.PBEKeySpec;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * This class represents a PBE key derived using PBKDF2 defined
+ * in PKCS#5 v2.0. meaning that
+ * 1) the password must consist of characters which will be converted
+ *    to bytes using UTF-8 character encoding.
+ * 2) salt, iteration count, and to be derived key length are supplied
+ *
+ * @author Valerie Peng
+ *
+ * See also same named class from OpenJDK. This class makes use of similar code.
+ */
+final class PBKDF2KeyImpl implements javax.crypto.interfaces.PBEKey {
+
+    private static final long serialVersionUID = -2234868909660948157L;
+
+    private char[] passwd;
+    private byte[] salt;
+    private final int iterCount;
+    private byte[] key;
+    private String prfAlgorithm;
+
+    private static byte[] getPasswordBytes(char[] passwd) {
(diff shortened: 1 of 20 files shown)


src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
```diff
@@ -0,0 +1,260 @@
...
+ *    to bytes using UTF-8 character encoding.
+ * 2) salt, iteration count, and to be derived key length are supplied
+ *
+ * @author Valerie Peng
+ *
+ * See also same named class from OpenJDK. This class makes use of similar code.
+ */
+final class PBKDF2KeyImpl implements javax.crypto.interfaces.PBEKey {
+
+    private static final long serialVersionUID = -2234868909660948157L;
```
KostasTsiounis, line 42: Should we add `@java.io.Serial` like in later versions?

src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
```diff
@@ -0,0 +1,260 @@
...
+
+    /**
+     * Replace the PBE key to be serialized.
+     *
+     * @return the standard KeyRep object to be serialized
+     *
+     * @throws ObjectStreamException if a new object representing
+     * this PBE key could not be created
+     */
+    private Object writeReplace() throws ObjectStreamException {
```
KostasTsiounis, line 217: Should we add `@java.io.Serial` like in later versions?

src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
```diff
@@ -0,0 +1,260 @@
...
+    /**
+     * Restores the state of this object from the stream.
+     * <p>
+     * Deserialization of this class is not supported.
+     *
+     * @param  stream the {@code ObjectInputStream} from which data is read
+     * @throws IOException if an I/O error occurs
+     * @throws ClassNotFoundException if a serialized class cannot be loaded
+     */
+    private void readObject(ObjectInputStream stream) throws IOException, ClassNotFoundException {
```
KostasTsiounis, line 235: Should we add `@java.io.Serial` like in later versions?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
jasonkatonica: Hi Kostas, This syntax `@java.io.Serial` is not available in Java 11. My reading on this annotation is that it should not effect behavior of serialization itself, only help indicate intended use. In this case ensures that all the serializable methods are implemented in a class. There is some more interesting discussion for this at https://stackoverflow.com/questions/63783474/what-is-the-use-of-serial-annotation-as-of-java-14 which seems to match other discussions that i have seen on this topic.

src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
jasonkatonica: See comment at https://github.com/IBM/OpenJCEPlus/pull/469#discussion_r1967636810

src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java
jasonkatonica: See comment at https://github.com/IBM/OpenJCEPlus/pull/469#discussion_r1967636810

taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

PR #473, opened 2025-02-26
Enhancements to AES-GCM Tests and Cipher Initialization

taoliult:
This update addresses multiple issues related to the AES-GCM cipher.

- AES context creation is now validated to prevent returning a NULL context. Using a NULL context in subsequent API calls was previously causing crashes.

- The ICC_AES_GCM_CTX_ctrl API has been removed, as it was deprecated and no longer needed.

- AES-GCM test cases were updated to eliminate unintended use of static variables in multi-threaded scenarios. This issue could lead to unpredictable behavior due to concurrent reads and writes to shared fields.

- BaseTestAESGCMUpdate now uses StringBuilder instead of string concatenation, reducing memory pressure and improving efficiency.

Signed-off-by: Tao Liu <tao.liu@ibm.com>

### DIFF
--- a/src/main/native/GCM.c
+++ b/src/main/native/GCM.c
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023, 2024
+ * Copyright IBM Corp. 2023, 2025
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms provided by IBM in the LICENSE file that accompanied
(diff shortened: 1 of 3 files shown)


KostasTsiounis (approved):
LGTM

jasonkatonica:
Hello could the commit message and PR description be updated to describe a bit more why we are making these changes? Might help in the future. For example something like:

---
This update fixes a wide array of issues found associated with the AES cipher.
- AES context creation is now checked for returning a NULL context. The subsequent use of a NULL context has been seen to cause a crash when used in other APIs.
- The API ICC_AES_GCM_CTX_ctrl was eliminate since this API has been deprecated and is no longer needed.
- Updates were made to AES GCM test cases where static variables were being used unintentionally in multi-threaded test cases. This has potential to cause the test to run with unexpected paths given each thread could be reading / writting to these static shared fields.
- `BaseTestAESGCMUpdate` was modified to use StringBuilder instead of appending to a String. This will cause less pressure on memory and GC and is more efficient.

jasonkatonica (approved):
Code looks good to me. Only recommendations for the PR / commit message.

JinhangZhang (approved):
LGTM

taoliult:
@jasonkatonica Update both the commit message and PR description.

johnpeck-us-ibm (approved):

PR #498, opened 2025-03-04
Enhancements to AES-GCM Tests and Cipher Initialization

taoliult:
This is a back port PR from PR https://github.com/IBM/OpenJCEPlus/pull/473

This update addresses multiple issues related to the AES-GCM cipher.

AES context creation is now validated to prevent returning a NULL context. Using a NULL context in subsequent API calls was previously causing crashes.

The ICC_AES_GCM_CTX_ctrl API has been removed, as it was deprecated and no longer needed.

AES-GCM test cases were updated to eliminate unintended use of static variables in multi-threaded scenarios. This issue could lead to unpredictable behavior due to concurrent reads and writes to shared fields.

BaseTestAESGCMUpdate now uses StringBuilder instead of string concatenation, reducing memory pressure and improving efficiency.

### DIFF
--- a/src/main/native/GCM.c
+++ b/src/main/native/GCM.c
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023, 2024
+ * Copyright IBM Corp. 2023, 2025
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms provided by IBM in the LICENSE file that accompanied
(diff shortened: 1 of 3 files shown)


KostasTsiounis (approved):
LGTM

jasonkatonica (approved):
LGTM

johnpeck-us-ibm (approved):

PR #501, opened 2025-03-05
Remove local copies of NamedCurve and CurveDB

KostasTsiounis:
The exact same classes are offered by the `JDK`, so there is no need for us to maintain our own copies of `NamedCurve` and `CurveDB`.

Some changes are also made in the `ECParameters` class to account for a method from `CurveDB` that is not public, as well as match `OpenJDK` behaviour of `getParameterSpec`.


src/main/java/com/ibm/crypto/plus/provider/ECParameters.java
```diff
@@ -185,7 +190,7 @@ protected <T extends AlgorithmParameterSpec> T engineGetParameterSpec(Class<T> s
 
         if (spec.isAssignableFrom(ECGenParameterSpec.class)) {
             // Ensure the name is the Object ID
-            String name = namedCurve.getObjectId();
+            String name = namedCurve.getNameAndAliases()[0];
```
jasonkatonica, line 193: Similar question here. I assume this method though throws an exception if this is not found.

src/main/java/com/ibm/crypto/plus/provider/ECParameters.java
```diff
@@ -138,7 +140,10 @@ protected void internalInit(AlgorithmParameterSpec paramSpec)
             namedCurve = CurveDB.lookup(name);
         } else if (paramSpec instanceof ECKeySizeParameterSpec) {
             int keySize = ((ECKeySizeParameterSpec) paramSpec).getKeySize();
-            namedCurve = CurveDB.lookup(keySize);
+            String name = ECUtils.getCurvefromSize(keySize);
+            if (name != null) {
+                namedCurve = CurveDB.lookup(name);
+            }
```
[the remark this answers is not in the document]
KostasTsiounis: Following the behaviour from the OpenJDK, the `getCurveFromSize` and `lookup` do not throw an exception, but rather return `null`. This is checked later in line 152 and a more uniform exception is thrown at this point.

src/main/java/com/ibm/crypto/plus/provider/ECParameters.java
KostasTsiounis: This never returns `null` as at least one name is added there in the constructor for each `NamedCurve`.

src/main/java/com/ibm/crypto/plus/provider/ECParameters.java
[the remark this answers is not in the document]
jasonkatonica: OK i see this makes sense given the null check is there and always executed at the end of the method.

jasonkatonica (approved):
LGTM

taoliult (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/ECParameters.java
johnpeck-us-ibm, line 193: This is fine, but if Oracle ever refactors the aliases, we may not get the OID back. Which may cause issues.

johnpeck-us-ibm (approved):

src/main/java/com/ibm/crypto/plus/provider/ECParameters.java
KostasTsiounis: We actually don't want the OID but rather the actual curve name is order to be in sync with what the JDK is doing. An actual JDK test was failing because of this.

johnpeck-us-ibm: Than please change the comment in the code. It is misleading at this point.

KostasTsiounis: Oh, right. I missed that. Changed.

PR #504, opened 2025-03-14
Remove local copies of NamedCurve and CurveDB

KostasTsiounis:
The exact same classes are offered by the `JDK`, so there is no need for us to maintain our own copies of `NamedCurve` and `CurveDB`.

Some changes are also made in the `ECParameters` class to account for a method from `CurveDB` that is not public, as well as match `OpenJDK` behaviour of `getParameterSpec`.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/501


johnpeck-us-ibm (approved):

jasonkatonica (approved):
LGTM

taoliult (approved):
LGTM

PR #508, opened 2025-03-18
Potential multithreading issues related to static variables

taoliult:
This is a back port PR from PR https://github.com/IBM/OpenJCEPlus/pull/485

This commit checks all static variables across Java files to investigate potential multithreading issues and implements the following changes:

AESGCMCipher.java: Update getSecureRandom with a synchronized block to ensure thread safety.

CurveUtil.java, ECNamedCurve.java, OpenJCEPlusProvider.java, RSAKeyPairGenerator.java, and ByteArrayOutputDelay.java: Update some static variables to static final since they remain unchanged after initialization.

SymmetricCipher.java: Update static variables to static final as they do not change after initialization. Additionally, use ConcurrentHashMap to enforce thread safety.

In addition to the changes made in this commit, there are other static variables, as outlined below:

CurveUtil.java: This class contains three static Map variables, which are initialized in a static block. After initialization, these maps are only read, not written or modified. So, they are thread-safe.

NativeInterface.java: The variables osName and osArch are defined as private static String. Regardless of potential race conditions or multiple threads trying to initialize these variables, the values of these static variables remain consistent across the JVM. So, they should be thread-safe.

Digest.java: This class contains two static arrays, which are initialized within a synchronized block. As a result, the initialization is thread-safe, ensuring that the arrays are correctly set up for concurrent access.

AESCCMCipher.java: The static variable private static SecureRandom random = null; is declared but not used in this class.

AESCipher.java: The static variable private static int isHardwareSupport = 0; is declared but not used in this class.

OpenJCEPlus.java and OpenJCEPlusFIPS.java: The static variables ockInitialized and ockContext are initialized within a synchronized method, so they should be fine. The static variable private static Map<String, String> attrs; is declared but not used in this class.

OAEPParameters.java: The static variables OID_MGF1 and OID_PSpecified are initialized in a static block, so they should be fine.

OCKContext.java: The static variable private static String libraryBuildDate = unobtainedValue; is initialized within a synchronized method, so it should be fine.

CurveDB.java: This class contains static maps, but since we removed CurveDB in another PR, no changes are needed for those maps.

CCMCipher.java, GCMCipher.java, and SymmetricCipher.java: The hardware check variables are defined as static. However, on open platforms, hardware support is not checked at all. So, this static variable has no actual impact.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESGCMCipher.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESGCMCipher.java
@@ -780,7 +780,11 @@ public final class AESGCMCipher extends CipherSpi implements AESConstants, GCMCo
         if (firstIV) {
             // SecureRandom random = null;
             if (random == null) {
-                random = provider.getSecureRandom(null);
+                synchronized (AESGCMCipher.class) {
+                    if (random == null) {
+                        random = provider.getSecureRandom(null);
+                    }
+                }
             }
             generatedIVDevField = new byte[GENERATED_IV_DEVICE_FIELD_LENGTH];
             random.nextBytes(generatedIVDevField);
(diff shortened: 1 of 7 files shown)


johnpeck-us-ibm (approved):

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

PR #516, opened 2025-04-03
Tolerate empty message EdDSA verify

jasonkatonica:
The EdDSA signature algorithm always returned the value false during verification in the case that no message was provided to the user. This behavior diverges from other cryptographic providers. This update allows for verification of null messages.

The EdDSA interop test was modified to ensure that sign and verify operations between bouncy castle and OpenJCEPlus providers work including sign and verify options without any message sent to the Signature instance. The test was also converted to a parameterized test and now uses the JCE framework instead of the bouncy castle key factory.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/EdDSASignature.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/EdDSASignature.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2023, 2025
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms provided by IBM in the LICENSE file that accompanied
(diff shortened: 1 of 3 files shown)


src/main/java/com/ibm/crypto/plus/provider/EdDSASignature.java
```diff
@@ -183,12 +183,12 @@ protected boolean engineVerify(byte[] sigBytes) throws SignatureException {
         if (!publicKeyInit) {
             throw new SignatureException("Missing public key");
         }
-        if (message == null) {
-            return false;
-        }
 
         try {
-            byte[] messageBytes = message.toByteArray();
+            byte[] messageBytes = {};
```
KostasTsiounis, line 188: Can't you use `ensureMessageInit()` like in `engineSign`? Just mentioning it for consistency.

KostasTsiounis (changes requested):

taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

JinhangZhang (approved):
LGTM

PR #518, opened 2025-04-03
Update ChaCha20 and ChaCha20Poly1305 cipher to match expected Cipher behaviour

KostasTsiounis:
Certain changes have to be made to match the official documentation regarding `Cipher` instances:
- In newer Java versions (11+), `init()` needs to be called after `doFinal()` when encrypting.
- When performing a `doFinal()` with a short buffer, a call to `init()` is required.

Tests to validate said behaviour are also added or altered accordingly.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/ChaCha20Cipher.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ChaCha20Cipher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023
+ * Copyright IBM Corp. 2023, 2025
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms provided by IBM in the LICENSE file that accompanied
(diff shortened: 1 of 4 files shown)


src/main/java/com/ibm/crypto/plus/provider/ChaCha20Cipher.java
```diff
@@ -456,12 +366,9 @@ private void checkCipherInitialized() throws IllegalStateException {
...
 
-    // Reset class variables after an exception
-    private void resetVarsAfterException() {
-        // force re-initialization presumably with different nonce and key
-        this.initialized = false;
-        this.sbeInLastFinalEncrypt = false;
-        this.initCalledInEncSeq = false;
+    // Reset class variables.
+    private void resetVars() {
+        this.initialized = (!this.encrypting); // force re-initialization only when decrypting
```
taoliult, line 371: Should be `// force re-initialization only when encrypting`?

src/main/java/com/ibm/crypto/plus/provider/ChaCha20Poly1305Cipher.java
```diff
@@ -626,12 +506,10 @@ private void checkCipherInitialized() throws IllegalStateException {
         }
     }
 
-    //Reset class variables after an exception
-    private void resetVarsAfterException() {
-        this.initialized = false; // force re-initialization presumably with different nonce and key
+    // Reset class variables.
+    private void resetVars() {
+        this.initialized = (!this.encrypting); // force re-initialization only when decrypting
```
taoliult, line 511: Also here, should be `// force re-initialization only when encrypting`?

src/main/java/com/ibm/crypto/plus/provider/ChaCha20Cipher.java
KostasTsiounis: Yes. Fixed.

src/main/java/com/ibm/crypto/plus/provider/ChaCha20Poly1305Cipher.java
KostasTsiounis: Yes. Fixed.

johnpeck-us-ibm (approved):

jasonkatonica (commented):
Do we need an additional test update in `BaseTestChaCha20` to cover those cases given we changed both algorithms here?

KostasTsiounis:
> Do we need an additional test update in `BaseTestChaCha20` to cover those cases given we changed both algorithms here?

I added some further tests in `BaseTestChaCha20` to account for retries after a `ShortBufferException` occurs. See 40bd38bef1f8a964e18a0c26dc7394f6a0394fc0.

PR #527, opened 2025-04-09
Tolerate empty message EdDSA

jasonkatonica:
The EdDSA signature algorithm always returned the value false in the case that no message was provided to the user. This behavior diverges from other cryptographic providers. This update allows for verification of null messages.

The EdDSA interop test was modified to ensure that sign and verify operations between bouncy castle and OpenJCEPlus providers work including sign and verify options without any message sent to the Signature instance. The test was also converted to a parameterized test and now uses the JCE framework instead of the bouncy castle key factory.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/516


taoliult (approved):
LGTM

KostasTsiounis (approved):
LGTM

PR #540, opened 2025-04-11
Throw InvalidKeyException when XDH small order

jasonkatonica:
This update matches JDK behavior for XDH key agreement when a small point order is encountered.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
@@ -34,7 +34,7 @@ abstract class XDHKeyAgreement extends KeyAgreementSpi {
     private long genCtx;
     private XECKey ockXecKeyPub = null;
     private XECKey ockXecKeyPriv = null;
-    private byte[] secret = null;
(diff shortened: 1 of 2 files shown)


src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
```diff
@@ -110,20 +110,22 @@ protected Key engineDoPhase(Key key, boolean lastPhase)
             this.secret = XECKey.computeECDHSecret(provider.getOCKContext(), genCtx,
                     ockXecKeyPub.getPKeyId(), ockXecKeyPriv.getPKeyId(), secrectBufferSize);
         } catch (OCKException e) {
+            //Valdate the secret for point has small order
```
KostasTsiounis, line 113: There is a small typo here. "Valdate" instead of "Validate". The wording is a bit off too I think.

jasonkatonica: Updating the spelling and wording here for the comment.

taoliult (approved):
LGTM

PR #544, opened 2025-04-11
Update ChaCha20 and ChaCha20Poly1305 cipher to match expected Cipher behaviour

KostasTsiounis:
Certain changes have to be made to match the official documentation regarding `Cipher` instances:
- In newer Java versions (11+), `init()` needs to be called after `doFinal()` when encrypting.
- When performing a `doFinal()` with a short buffer, a call to `init()` is required.

Tests to validate said behaviour are also added or altered accordingly.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/518


jasonkatonica (approved):
LGTM

PR #535, opened 2025-04-10
Check PSSParameterSpec with different message digest

JinhangZhang:
The mask generation function (MGF) based on a hash algo is recommended to use the same hash function as the hash function fingerprinting the message for RSA-PSS. For example
```
new PSSParameterSpec("SHA-384", "MGF1",
                MGF1ParameterSpec.SHA384, 48,
                PSSParameterSpec.TRAILER_FIELD_BC);
```
However the structures in [PKCS#1v2.1] allow for separate parameterization of the MGF and the message digest. For example
```
new PSSParameterSpec("SHA-384", "MGF1",
                MGF1ParameterSpec.SHA512, 48,
                PSSParameterSpec.TRAILER_FIELD_BC);
```

OpenJCEPlus restricts to use the same hash algo for both MGF and message digest. Adding test cases to ensure that OpenJCEPlus will fail the case if different message digest algos are used for MGF and md.

PR #549, opened 2025-04-16
Check PSSParameterSpec with different message digest

JinhangZhang:
The mask generation function (MGF) based on a hash algo is recommended to use the same hash function as the hash function fingerprinting the message for RSA-PSS. For example
```
new PSSParameterSpec("SHA-384", "MGF1",
                MGF1ParameterSpec.SHA384, 48,
                PSSParameterSpec.TRAILER_FIELD_BC);
```
However the structures in [PKCS#1v2.1] allow for separate parameterization of the MGF and the message digest. For example
```
new PSSParameterSpec("SHA-384", "MGF1",
                MGF1ParameterSpec.SHA512, 48,
                PSSParameterSpec.TRAILER_FIELD_BC);
```

OpenJCEPlus restricts to use the same hash algo for both MGF and message digest. Adding test cases to ensure that OpenJCEPlus will fail the case if different message digest algos are used for MGF and md.

backport from: https://github.com/IBM/OpenJCEPlus/pull/535


johnpeck-us-ibm (approved):

taoliult (approved):
LGTM

jasonkatonica (approved):
LGTM

PR #559, opened 2025-04-18
Optimize SecureRandom usage in various services

jasonkatonica:
`AESCCMCipher` was making use of the default `SecureRandom` algorithm configured on a system. This class should instead make use of the random associated with the provider in use.

`AESCCMCipher`, `AESCipher`, `AESKeyGenerator`, `ChaCha20KeyGenerator`, `DESedeCipher`, `DESedeKeyGenerator`, `ECKeyPairGenerator`, `GCMParameterGenerator`, `HmacKeyGenerator`,
`TlsRsaPremasterSecretGenerator` were modified to
reuse a single source of randomness instead of creating a new instance each time a service is initialized. Test execution times have improved due to this update.

All fields that reference sources of randomness are no longer static. There are concerns that a class may be initialized with a non FIPS provider prior to being used by the FIPS provider. In this case a wrong source of random could potentially be used. The `synchronized` block used in `AESGCMCipher` is no longer needed given this field is no longer static.

Various classes were updated to use a consistent field name, `cryptoRandom`, to hold a sources of randomness. This was done to avoid potential name conflicts with the commonly used `random` argument sent through various APIs. References to this field no longer make use of the `this` keyword given this refactor of the field name now being more unique. A bug was found in both the `ChaCha20Cipher` and `ChaCha20Poly1305Cipher` classes in which the user defined source of randomess was being accidentally being used due to a mix up between `this.random` and the method parameter `random`. The deterministic test was adjusted accordingly for the new behavior.

`CCMParameterGenerator` and `TlsRsaPremasterSecretGenerator` no longer honor a user defined source of randomness similar to our behavior in other services.

The unused field `random` was removed from `EdDSAKeyPairGenerator`.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023, 2024
+ * Copyright IBM Corp. 2023, 2025
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms provided by IBM in the LICENSE file that accompanied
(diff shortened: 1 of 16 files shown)


KostasTsiounis (approved):
LGTM

taoliult (approved):
LGTM

PR #564, opened 2025-04-23
Add check for unsupported EdDSA signature algorithms

taoliult:
The EdDSA signature implementation in OpenJCEPlus only supports the default mode (Ed25519 or Ed448), where the EdDSAParameterSpec context is null and prehash is set to false. This update adds checks in EdDSASignature to ensure Invalid algorithm parameter exception is properly thrown.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/EdDSASignature.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/EdDSASignature.java
@@ -71,7 +71,12 @@ abstract class EdDSASignature extends SignatureSpi {
         // input params and operation are checked.
         // This edDSA singature is using default mode (Ed25519 or Ed448)
         // for edDSAParameterSpec (context = null, prehash = false)
-        if (params instanceof EdDSAParameterSpec) {
+        if (params instanceof EdDSAParameterSpec edParams) {
+            if (edParams.isPrehash() || !edParams.getContext().isEmpty()) {
+                throw new InvalidAlgorithmParameterException(
+                        "The EdDSA signature only supports the default mode (Ed25519 or Ed448),"
+                        + " where the EdDSAParameterSpec context is null and prehash is set to false");
+            }
             if (message != null) {
                 // Sign/Verify is in progress
                 throw new InvalidParameterException(
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestEdDSASignature.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestEdDSASignature.java
@@ -20,6 +20,7 @@ import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.Signature;
 import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.EdDSAParameterSpec;
 import java.security.spec.EdECPoint;
 import java.security.spec.EdECPrivateKeySpec;
 import java.security.spec.EdECPublicKeySpec;
@@ -28,8 +29,11 @@ import java.security.spec.NamedParameterSpec;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.Arrays;
+import java.util.HexFormat;
 import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 public class BaseTestEdDSASignature extends BaseTestJunit5Signature {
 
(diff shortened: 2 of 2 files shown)


src/test/java/ibm/jceplus/junit/base/BaseTestEdDSASignature.java
```diff
@@ -270,6 +283,19 @@ private void runSignTest(String algorithm, AlgorithmParameterSpec params, String
...
+            throws Exception {
+        try {
+            Signature sig = Signature.getInstance(algorithm, getProviderName());
+            sig.setParameter(params);
+            fail("Expected InvalidAlgorithmParameterException for unsupported signature algorithm is NOT thrown");
+        } catch (InvalidAlgorithmParameterException e) {
+            String message = e.getMessage();
+            assertTrue((message != null) && message.contains("The EdDSA signature only supports the default mode"),
+                    "Unexpected exception message: " + message);
+        }
```
jasonkatonica, line 296: Think you could instead use the assertion format as below? This will print the message and perform same logic you have here only using the built in assertions for Junit:
```
        } catch (InvalidAlgorithmParameterException e) {
            assertEquals(The EdDSA signature only supports the default mode", e.getMessage());
        }
```

jasonkatonica (changes requested):

jasonkatonica:
Hi @taoliult Can you please also rebase on the latest version of code present in `main` branch? There are some performance fixes that allow for `s390x` linux to complete without timing out.

src/test/java/ibm/jceplus/junit/base/BaseTestEdDSASignature.java
taoliult: Done, changed to use `assertEquals`.

jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/EdDSASignature.java
```diff
@@ -72,6 +72,12 @@ protected void engineSetParameter(AlgorithmParameterSpec params)
         // This edDSA singature is using default mode (Ed25519 or Ed448)
         // for edDSAParameterSpec (context = null, prehash = false)
         if (params instanceof EdDSAParameterSpec) {
+            EdDSAParameterSpec edParams = (EdDSAParameterSpec) params;
```
KostasTsiounis, line 75: Should we use the `if (params instanceof EdDSAParameterSpec edParams) {` way of casting here?

src/test/java/ibm/jceplus/junit/base/BaseTestEdDSASignature.java
```diff
@@ -233,6 +237,16 @@ public void testRunEdDSAKAT() throws Exception {
...
 
+        // Ed25519ctx
+        byte[] context = HexFormat.of().parseHex("666f6f");
+        runUnsupportedAlgorithmParameterSpecTest("Ed25519", new EdDSAParameterSpec(false, context));
+        // Ed25519ph
+        runUnsupportedAlgorithmParameterSpecTest("Ed25519", new EdDSAParameterSpec(true));
+        // Ed448ph
+        runUnsupportedAlgorithmParameterSpecTest("Ed448", new EdDSAParameterSpec(false, context));
+        runUnsupportedAlgorithmParameterSpecTest("Ed448", new EdDSAParameterSpec(true));
+        runUnsupportedAlgorithmParameterSpecTest("Ed448", new EdDSAParameterSpec(true, context));
```
KostasTsiounis, line 248: Don't we need this use case for `Ed25519` as well?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/EdDSASignature.java
taoliult: We can, but this is introduced from JDK17 and above version, we still need to use this old way in JDK11. Is there any specific reason we have to use the new style? Or would it be OK to keep the code consistent across all versions?

src/test/java/ibm/jceplus/junit/base/BaseTestEdDSASignature.java
taoliult: Good question, for those `EdDSAParameterSpec` combination, I copied the test scenarios from the existing OpenJDK test `TestEdDSA`, from those existing `Ed25519ctx`, `Ed25519ph` and `Ed448ph` tests. For the `Ed25519ctx` and `Ed25519ph` tests in `TestEdDSA`, they don't have those combinations. 

And, in our `EdDSASignature` class, the parameter checks are performed when setting the parameters. No matter it’s `Ed25519` or `Ed448`, they both go through the same `engineSetParameter` method, and that method only checks the parameters. So even if `Ed25519` doesn’t have those combinations, it won't impact the test coverage.

src/main/java/com/ibm/crypto/plus/provider/EdDSASignature.java
KostasTsiounis: We use this convention in all other places where we use the variable after and just change it for 11. By changing it here we'll be consistent with the rest of the codebase. Plus we'll have the new syntax moving forward when 11 is EOS. Not married to the idea, but I think it's more readable. @jasonkatonica what do you think?

taoliult: Ok, agree, let me make this change to keep it consistent with the rest of the codebase—using the new style for version 17 and above, and keep the old style for version 11.

taoliult: Done.

johnpeck-us-ibm (approved):

src/main/java/com/ibm/crypto/plus/provider/EdDSASignature.java
jasonkatonica: Yea agreed that ideally we keep the new style for 17 and higher as we have been doing this in the past.

KostasTsiounis (approved):
LGTM

taoliult:
Backport PRs have been created. Since JDK 11 does not support EdDSA signatures, the backports are only for JDK 24, 21, and 17.

PR #573, opened 2025-04-28
Update KeyAgreement classes to support Generic secret generation

KostasTsiounis:
The `engineGenerateSecret(String)` method in KeyAgreement classes currently only supports `"TlsPremasterSecret"`, but the expectation starting in Java 25 is that `"Generic"` should be accepted as well. This change accomplishes that.

Additional test cases to verify said behaviour are added.


johnpeck-us-ibm (approved):

jasonkatonica (changes requested):
Updates look good except for copyright date changes at the top of modified files.

KostasTsiounis:
> Updates look good except for copyright date changes at the top of modified files.

Updated.

jasonkatonica (approved):
LGTM

taoliult:
The "Generic" should be accepted starting from Java 25, that means there is no need to back port this change?

src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
```diff
@@ -150,9 +150,12 @@ protected SecretKey engineGenerateSecret(String algorithm)
...
             throw new NoSuchAlgorithmException("Algorithm must not be null");
-        if (!(algorithm.equals("TlsPremasterSecret")))
-            throw new NoSuchAlgorithmException("Only supported for algorithm TlsPremasterSecret");
-        return new SecretKeySpec(engineGenerateSecret(), "TlsPremasterSecret");
+        if (!(algorithm.equalsIgnoreCase("TlsPremasterSecret")
+                || algorithm.equalsIgnoreCase("Generic"))) {
+            throw new NoSuchAlgorithmException(
+                    "Unsupported secret key algorithm: " + algorithm);
+        }
+        return new SecretKeySpec(engineGenerateSecret(), algorithm);
```
JinhangZhang, line 158: Since algorithm might not be exactly equals to "TlsPremasterSecret" or "Generic", it may also only contains lower cases such as "tlsPremasterSecret" or "generic". Does SecretKeySpec also allow those conditions?

KostasTsiounis: This code is in sync with the JDK equivalent, so I think the spec (also from the JDK) can work fine with them.

KostasTsiounis:
> The "Generic" should be accepted starting from Java 25, that means there is no need to back port this change?

Exactly. We will only make this change in the `main` branch.

PR #593, opened 2025-05-14
Check for acceptable values in NamedParameterSpec

KostasTsiounis:
When initializing `XDH` or `Ed` instances of keypair generator, key factory and/or key agreement classes using a `NamedParameterSpec` instance, ensure that only appropriate values are accepted or throw an `InvalidAlgorithmParameterException`.

Tests to check that functionality are also added.

Signed-off-by: Kostas Tsiounis

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
@@ -9,7 +9,10 @@
 package com.ibm.crypto.plus.provider;
 
 import java.io.IOException;
+import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidParameterException;
+import java.security.ProviderException;
+import java.security.spec.NamedParameterSpec;
 import java.util.HashMap;
 import java.util.Map;
 import sun.security.util.KnownOIDs;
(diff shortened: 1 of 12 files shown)


taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

src/main/java/com/ibm/crypto/plus/provider/XDHKeyPairGenerator.java
```diff
@@ -74,7 +75,12 @@ private void initXDHKeyPairGenerator(OpenJCEPlusProvider provider, NamedParamete
     @Override
     public void initialize(int keySize, SecureRandom random) {
         CurveUtil.CURVE curve = CurveUtil.getCurveOfSize(keySize);
-        initializeImpl(new NamedParameterSpec(curve.name()));
+        try {
+            initializeImpl(new NamedParameterSpec(curve.name()));
+        } catch (InvalidAlgorithmParameterException iape) {
+            // should never happen
+            throw new ProviderException(iape.getMessage());
```
jasonkatonica, line 82: Can we add a caused by here and pass `iape` ?

src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
```diff
@@ -159,7 +181,20 @@ public static CURVE getCurve(String curveName)
...
+            if (curveName.contains("Ed")) {
+                curve = getEdCurve(spec);
+            } else {
+                curve = getXCurve(spec);
+            }
+            return getAlgId(curve);
+        } catch (InvalidAlgorithmParameterException iape) {
+            // Should never happen, since this is used by the key impls created
+            // from the key generators.
+            throw new ProviderException("getAldId was called with a non-supported curve");
```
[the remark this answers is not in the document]
KostasTsiounis: Added `iape` as cause.

src/main/java/com/ibm/crypto/plus/provider/XDHKeyPairGenerator.java
KostasTsiounis: Switched to constructor with `Throwable` instead of `String`, so that `iape` appears as cause of this.

src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
```diff
@@ -159,7 +181,20 @@ public static CURVE getCurve(String curveName)
      * @throws IOException
      */
     public static AlgorithmId getAlgId(String curveName) throws IOException {
-        return getAlgId(getCurve(curveName));
+        try {
+            CURVE curve;
+            NamedParameterSpec spec = new NamedParameterSpec(curveName);
+            if (curveName.contains("Ed")) {
```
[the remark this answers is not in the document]
KostasTsiounis: I added to `curveName.toUpperCase()` and also added a specific `else if` for `X` curves and an `else` that throws an exception, even though it should never happen.

src/main/java/com/ibm/crypto/plus/provider/CurveUtil.java
```diff
@@ -117,25 +120,44 @@ public static CurveUtil.CURVE getCurve(ObjectIdentifier oid, Integer size)
...
+    }
+
+    /**
+     * Returns the XEC curve type based on the provided named parameter spec.
+     *
+     * @param   params
+     * @return  CURVE
+     * @throws  InvalidAlgorithmParameterException
+     */
+    public static CURVE getXCurve(NamedParameterSpec params)
```
JinhangZhang, line 151: Does it necessary to separate XEC and Ed curves into two functions? It occurs to me that one function is enough

KostasTsiounis: I think it is necessary, because this functionality is used by `Ed*` classes and `XDH*` classes and only the appropriate curves should be allowed in each case. Not sure how this can be ensured using just one method. Do you have something in mind?

JinhangZhang: I dont see too much different between the original `getCurve` function and your change, rather than 

1. changing the parameters.
2. separate Ed* + XDH*. 

What I meant is:
```
public static CURVE getCurve(NamedParameterSpec params)
        throws InvalidAlgorithmParameterException {
        String curveName = params.getName();
        if (curveName == null)
            throw new InvalidAlgorithmParameterException("Invalid AlgorithmParameterSpec: " + params);
        switch (curveName.toUpperCase()) {
            case "X25519":
                return CURVE.X25519;
            case "X448":
                return CURVE.X448;
            case "ED25519":
                return CURVE.Ed25519;
            case "ED448":
                return CURVE.Ed448;
            case "FFDHE2048":
                return CurveUtil.CURVE.FFDHE2048;
            case "FFDHE3072":
                return CurveUtil.CURVE.FFDHE3072;
            case "FFDHE4096":
                return CurveUtil.CURVE.FFDHE4096;
            case "FFDHE6144":
                return CurveUtil.CURVE.FFDHE6144;
            case "FFDHE8192":
                return CurveUtil.CURVE.FFDHE8192;
            default:
                throw new InvalidAlgorithmParameterException("Invalid AlgorithmParameterSpec: " + params);
        }
    }
```
It seems put Ed* and XDH* into one single method is enough, it will return based on the value of NamedParameterSpec.

jasonkatonica (approved):
LGTM

JinhangZhang (approved):
LGTM

PR #604, opened 2025-05-16
Accept spec using a lowercase x448 curve name

taoliult:
In the XDHPublicKeyImpl class, when generating the value of p, the code checks whether the curve is X448 using the condition "this.params.getName().equals("X448”)". However, the “params.getName()” returns x448 (lowercase), the check fails, and the code falls into the “else” block, which is for the X25519 curve.

Changing from “this.params.getName()” to “this.curve.name()” will fix the issue. Inside the “CurveUtil.getCurve()” method, the curve name is changed to uppercase before comparison, and return the uppercase. This will guarantee that “this.curve.name()” always return uppercase “X448".

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/XDHKeyFactory.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/XDHKeyFactory.java
@@ -64,7 +64,7 @@ class XDHKeyFactory extends KeyFactorySpi {
                 }
 
                 //Validate algs match for key and keyfactory
-                if (this.alg != null && !(params.getName().equals(this.alg))) {
+                if (this.alg != null && !(params.getName().equalsIgnoreCase(this.alg))) {
                     throw new InvalidKeySpecException("Parameters must be " + this.alg);
                 }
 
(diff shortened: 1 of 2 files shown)


KostasTsiounis:
Can I suggest that we change the title to explain more the what instead of the why? Something like "Accept spec with lowercase curve name". Also, the description would make more sense in the issue, not the actual PR, since it describes the problem you are trying to solve instead of what you are doing with this change.

taoliult:
> Can I suggest that we change the title to explain more the what instead of the why? Something like "Accept spec with lowercase curve name". Also, the description would make more sense in the issue, not the actual PR, since it describes the problem you are trying to solve instead of what you are doing with this change.

Updated, please check.

KostasTsiounis (commented):
Should we add a test to verify this behaviour?

JinhangZhang (approved):
LGTM

taoliult:
> Should we add a test to verify this behaviour?

Yes, I will open another issue for adding the test case.

PR #606, opened 2025-05-21
Accept spec using a lowercase x448 curve name

taoliult:
In the XDHPublicKeyImpl class, when generating the value of p, the code checks whether the curve is X448 using the condition "this.params.getName().equals("X448”)". However, the “params.getName()” returns x448 (lowercase), the check fails, and the code falls into the “else” block, which is for the X25519 curve.

Changing from “this.params.getName()” to “this.curve.name()” will fix the issue. Inside the “CurveUtil.getCurve()” method, the curve name is changed to uppercase before comparison, and return the uppercase. This will guarantee that
“this.curve.name()” always return uppercase “X448".


KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

jasonkatonica (approved):
LGTM

PR #621, opened 2025-06-02
React to JCE framework changes for key values

jasonkatonica:
This update migrates to use the new value in the JCE framework that is intended on being used for holding the key value.


KostasTsiounis:
Shouldn't we have a separate PR with the pipeline changes so that we can revert them once an official 25 build comes out?

johnpeck-us-ibm (approved):

taoliult:
So, the change from `key` to `privKeyMaterial`, I assume it only applies starting with Java 25?

taoliult (approved):
LGTM

PR #622, opened 2025-06-02
Add test case for accepting spec with lowercase X448 curve name

taoliult:
This test validates the fix for handling lowercase X448 curve names by comparing the public key generated via a key pair generator with the one reconstructed from a public key spec using a key factory. This ensures the correctness of the X448 path in the key factory implementation.

### DIFF
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestXDH.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestXDH.java
@@ -20,6 +20,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.NoSuchProviderException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
+import java.security.interfaces.XECPublicKey;
 import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.KeySpec;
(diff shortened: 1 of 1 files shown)


KostasTsiounis:
I'm not sure how this test a spec with a lowercase curve name. The generator is initialized with a spec with uppercase and then all subsequent specs are retrieved from the key, which should be the same. Is there another step I am missing here?

src/test/java/ibm/jceplus/junit/base/BaseTestXDH.java
```diff
@@ -40,16 +41,22 @@ public class BaseTestXDH extends BaseTestJunit5 {
...
         NamedParameterSpec nps = new NamedParameterSpec(curveName);
         compute_xdh_key(curveName, nps);
+        runKeyFactoryTest(curveName, nps);
     }
 
     @Test
     public void testXDH_X448() throws Exception {
+        System.out.println(
+                "\n\n\n\n************************** Starting testXDH_X448 ************************");
         String curveName = "X448";
```
johnpeck-us-ibm, line 56: Where is the lower case name?

johnpeck-us-ibm (approved):

KostasTsiounis:
I also think we're not covering all the code changes we made in the previous PR. There is a change in the path for `generatePrivate()`, another one in the `getKeySpec()` where the factory has been initialized with `X25519` or `X448` specifically instead of just `XDH` and another one through `translateKey()`. Should we cover those paths too or do we think we don't need them?

taoliult:
> I also think we're not covering all the code changes we made in the previous PR. There is a change in the path for `generatePrivate()`, another one in the `getKeySpec()` where the factory has been initialized with `X25519` or `X448` specifically instead of just `XDH` and another one through `translateKey()`. Should we cover those paths too or do we think we don't need them?

The change replaces `equals` with `equalsIgnoreCase`. Without this update, a mismatch ( lowercase "x" not matching uppercase "X" ) would result in an exception. With the update, such cases no longer throw an exception. I don't think we need dedicated test cases for those exception scenarios. The test case I added only verifies the change from `params.getName()` to `curve.name()`.

jasonkatonica (approved):
LGTM

PR #677, opened 2025-06-30
Allow OpenJCEPlus to run in developer mode

jasonkatonica:
Users may wish to develop on various platforms in which the OpenJCEPlusFIPS provider currently does not function. This update allows for execution on such platforms while issuing a warning to users when loading the provider.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
@@ -30,6 +30,8 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
     // Field serialVersionUID per tag [SERIALIZATION] in DesignNotes.txt
     private static final long serialVersionUID = 929669768004683845L;
 
+    private static final boolean printFipsDeveloperModeWarning = Boolean.parseBoolean(System.getProperty("openjceplus.fips.devmodewarn", "true"));
+
     private static final String info = "OpenJCEPlusFIPS Provider implements the following:\n" +
 
             "Algorithm parameter                : AES, DiffieHellman, DSA, EC, GCM, OAEP, RSAPSS\n"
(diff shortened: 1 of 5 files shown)


johnpeck-us-ibm (approved):

KostasTsiounis (approved):
LGTM

jasonkatonica:
> Do we need a flag to enable/disable this developer mode, or will it always be available?

On all platforms where the new method `isFIPSCertifiedPlatform` returns false developer mode will be turned on automatically. Additionally a few warning messages to stdout will be issued in this case.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -103,9 +103,11 @@ public OpenJCEPlusFIPS() {
             debug.println("New OpenJCEPlusFIPS instance");
         }
 
-        if (!isPlatformSupported) {
-            throw new UnsupportedOperationException(
-                        "OpenJCEPlusFIPS is not supported on this non FIPS " + osName + " " + osArch + " platform");
+        if (!isFIPSCertifiedPlatform) {
+            System.out.println("WARNING: OpenJCEPlusFIPS is running in developer mode. Non production workload assumed. This environment is not certified for FIPS 140-3: " + osName + ":" + osArch);
```
taoliult, line 107: Do we need to remove this System.out.println(), since the same message will be printed when debug is enabled below?

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -799,6 +801,13 @@ private synchronized void initializeContext() {
 
         try {
             boolean useFIPSMode = true;
+            if (!isFIPSCertifiedPlatform) {
+                System.out.println("WARNING: OpenJCEPlusFIPS is about to load non FIPS 140-3 library!");
```
taoliult, line 805: Same here, do we still need this System.out.println()?

PR #678, opened 2025-07-01
Update doc for install and PQC

jasonkatonica:
- Documented how to install a JCE provider using Security.insertProviderAt
- Enhanced static configuration instructions via java.security file
- Added select notes for PQC algorithms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Build Status:
 `OpenJCEPlus` and `OpenJCEPlusFIPS` providers are currently supported on the following architectures and operating system combinations as reported by `mvn --version` in the values `OS name` and `arch`:
 | OS name                 | arch        |
 | ----------------------- | ----------- |
+| linux                   | aarch64     |
 | linux                   | amd64       |
 | linux                   | s390x       |
 | linux                   | ppc64le     |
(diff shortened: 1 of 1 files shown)


johnpeck-us-ibm (approved):

README.md
```diff
@@ -186,15 +187,37 @@ mvn '-Dock.library.path=$PROJECT_HOME/OCK/' test -Dtest=TestClassname
...
+        public class ProviderInsert {
+            public static void main(String[] args) {
+                // Insert the provider at the top, position 1
+                Security.insertProviderAt(new com.ibm.crypto.plus.provider.OpenJCEPlus(), 1);
+            }
+        }
+        ```
 
-1. Execute your application specifying the `openjceplus.jar`, the location of the OCK library, and the location of the jgskit library as follows.
+1. If you're using a Java runtime environment that already includes OpenJCEPlus, such as Semeru, you can skip the following step.
```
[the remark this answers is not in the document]
jasonkatonica: In general the documented mentioned here is specific to using OpenJCEPlus/OpenJCEPlusFIPS regardless of Java runtime environments. I only mentioned the use of exports here given that Semeru configuration and usages via the restricted security mode mechanisms and the FIPS 140-3 profile is different and expected to be documented by Semeru. I dont believe it is the goal of this project to document restricted security mode profile and its usages within Semeru to automatically configure OpenJCEPlus.

In terms of specifying OpenJCEPlus Jar file and JGSKIT library if somebody was doing development of their own versions of these files I would still expect some action to be required to use their own in Semeru to override what is bundled there.

JinhangZhang (approved):
LGTM

README.md
```diff
@@ -186,15 +187,37 @@ mvn '-Dock.library.path=$PROJECT_HOME/OCK/' test -Dtest=TestClassname
...
+
+    - Approach 2: Programmatic Installation Using `insertProviderAt`
+        To register a JCE provider programmatically, use the `java.security.Security.insertProviderAt` method. This method allows you to specify the provider and its preference order directly in your Java code.
+
+        ```java
+        import java.security.Security;
+        public class ProviderInsert {
+            public static void main(String[] args) {
+                // Insert the provider at the top, position 1
+                Security.insertProviderAt(new com.ibm.crypto.plus.provider.OpenJCEPlus(), 1);
```
KostasTsiounis, line 213: Do we need to point to the alternative for `OpenJCEPlusFIPS`?

KostasTsiounis: Also, should we replace `1` with `XX` from the previous approach?

README.md
KostasTsiounis, line 218: Should we mention the need for `--patch-module`?

KostasTsiounis (changes requested):

README.md
jasonkatonica: Made an update to add what i think we would need at a minimum for semeru using `--patch-module` to pick up a set of built maven artifacts.

README.md
jasonkatonica: Changed this section to use XX in the code even though that will not run.

PR #681, opened 2025-07-03
Remove com.ibm.crypto.provider.restrictRSAExponent

taoliult:
This commit removes the system property com.ibm.crypto.provider.restrictRSAExponent which is false by default, along with all related code.


src/main/java/com/ibm/crypto/plus/provider/RSAKeyFactory.java
```diff
@@ -109,14 +96,6 @@ static void checkKeyLengths(int modulusLen, BigInteger exponent, int minModulusL
         if (modulusLen > maxLen) {
             throw new InvalidKeyException("RSA keys must be no longer than " + maxLen + " bits");
         }
-
-        // If a RSAPublicKey, make sure the exponent isn't too big.
-        if (restrictExpLen && (exponent != null) && (modulusLen > MAX_MODLEN_RESTRICT_EXP)
```
JinhangZhang, line 114: Does this check unnecessary?

taoliult: By default, this property is set to false, and throughout our codebase and test cases, we never explicitly set it to true. So, this check is effectively never used. I believe it's safe to remove both this property and its corresponding check.

JinhangZhang (approved):
LGTM

johnpeck-us-ibm (approved):

PR #682, opened 2025-07-03
Update hashCode() and equals() in keys to match OpenJDK's version

KostasTsiounis:
Several keys, like `AESKey`, `ChaCha20Key`, `DESedeKey`, `DSAPrivateKey`, `ECPrivateKey` and `EdDSAPrivateKey`, are updated so that the result of their `hashCode()` and `equals()` methods matches the one produced by the `OpenJDK` equivalent.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESKey.java
@@ -12,6 +12,7 @@ import java.security.InvalidKeyException;
 import java.security.KeyRep;
 import java.util.Arrays;
 import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
 import javax.security.auth.DestroyFailedException;
 
 /**
@@ -72,11 +73,8 @@ final class AESKey implements SecretKey {
     public int hashCode() {
         checkDestroyed();
 
-        int retval = 0;
-        for (int i = 1; i < this.key.length; i++) {
-            retval += this.key[i] * i;
-        }
-        return (retval ^= "aes".hashCode());
+        SecretKeySpec keySpec = new SecretKeySpec(this.key, getAlgorithm());
+        return keySpec.hashCode();
     }
 
     @Override
--- a/src/main/java/com/ibm/crypto/plus/provider/ChaCha20Key.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ChaCha20Key.java
@@ -12,6 +12,7 @@ import java.security.InvalidKeyException;
 import java.security.KeyRep;
 import java.util.Arrays;
 import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
 import javax.security.auth.DestroyFailedException;
 
 /**
@@ -73,11 +74,8 @@ final class ChaCha20Key implements SecretKey, ChaCha20Constants {
     public int hashCode() {
         checkDestroyed();
 
-        int retval = 0;
-        for (int i = 1; i < this.key.length; i++) {
-            retval += this.key[i] * i;
-        }
-        return (retval ^= "ChaCha20".hashCode());
+        SecretKeySpec keySpec = new SecretKeySpec(this.key, getAlgorithm());
+        return keySpec.hashCode();
     }
 
     @Override
--- a/src/main/java/com/ibm/crypto/plus/provider/DESedeKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/DESedeKey.java
@@ -13,6 +13,7 @@ import java.security.KeyRep;
 import java.util.Arrays;
 import javax.crypto.SecretKey;
 import javax.crypto.spec.DESedeKeySpec;
+import javax.crypto.spec.SecretKeySpec;
 import javax.security.auth.DestroyFailedException;
 import javax.security.auth.Destroyable;
 
@@ -75,11 +76,8 @@ final class DESedeKey implements SecretKey, Destroyable {
     public int hashCode() {
         checkDestroyed();
 
-        int retval = 0;
-        for (int i = 1; i < this.key.length; i++) {
-            retval += this.key[i] * i;
-        }
-        return (retval ^= "desede".hashCode());
+        SecretKeySpec keySpec = new SecretKeySpec(this.key, getAlgorithm());
+        return keySpec.hashCode();
     }
 
     @Override
--- a/src/main/java/com/ibm/crypto/plus/provider/DSAPrivateKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/DSAPrivateKey.java
@@ -14,7 +14,6 @@ import java.io.Serializable;
 import java.math.BigInteger;
 import java.security.AlgorithmParameters;
 import java.security.InvalidKeyException;
-import java.security.Key;
 import java.security.interfaces.DSAParams;
 import java.security.spec.DSAParameterSpec;
 import java.security.spec.InvalidParameterSpecException;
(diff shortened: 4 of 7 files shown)


src/main/java/com/ibm/crypto/plus/provider/EdDSAKeyFactory.java
```diff
@@ -75,7 +74,7 @@ protected Key engineTranslateKey(Key key) throws InvalidKeyException {
                 privKey = (com.ibm.crypto.plus.provider.EdDSAPrivateKeyImpl) key;
             } else {
                 try {
-                    privKey = new EdDSAPrivateKeyImpl(provider, params, Optional.of(privateKeyBytes));
+                    privKey = new EdDSAPrivateKeyImpl(provider, params, privateKeyBytes);
```
jasonkatonica, line 77: Why make this change? `Optional.of` is usually used for null checking a value I woudl assume here. So in this spot we do check for `null` on line 71 just above so seems ok to remove. Similar comment on on line 174 below in that we know the bytes are a value at that point. Just curious why remove these and what behavior change is intended? ( Similar question with other `Optional` removals.

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
```diff
@@ -177,12 +182,11 @@ private byte[] processEncodedPrivateKey(byte[] encoded) throws IOException {
         // Read, convert, then write private key
         this.privKeyMaterial = inputValue[2].getOctetString(); // Get octet string
         //Need to remove seq tag from key
-        this.privKeyMaterial = Arrays.copyOfRange(this.privKeyMaterial, 2, this.privKeyMaterial.length);
-        this.h = Optional.of(this.privKeyMaterial);
+        //this.privKeyMaterial = Arrays.copyOfRange(this.privKeyMaterial, 2, this.privKeyMaterial.length);
```
jasonkatonica, line 185: Perhaps could just remove this line instead of commenting it out.

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
```diff
@@ -244,7 +248,7 @@ private byte[] extractPrivateKeyFromOCK(byte[] privateKeyBytes) throws IOExcepti
         byte[] privData = null;
         if (inputValue.length > 2) {
             privData = inputValue[2].getOctetString();
-            privData = new DerInputStream(privData).getOctetString();
+            //privData = new DerInputStream(privData).getOctetString();
```
jasonkatonica, line 251: Similar comment here we can remove the line instead of commenting it out.

src/main/java/com/ibm/crypto/plus/provider/AESKey.java
```diff
@@ -72,11 +73,8 @@ public synchronized byte[] getEncoded() {
...
         checkDestroyed();
 
-        int retval = 0;
-        for (int i = 1; i < this.key.length; i++) {
-            retval += this.key[i] * i;
-        }
-        return (retval ^= "aes".hashCode());
+        SecretKeySpec keySpec = new SecretKeySpec(this.key, getAlgorithm());
+        return keySpec.hashCode();
     }
```
johnpeck-us-ibm, line 78: Note: SecretKeySpec.hashcode is basically do the same as the old code.

src/main/java/com/ibm/crypto/plus/provider/DESedeKey.java
```diff
@@ -75,11 +76,8 @@ public synchronized byte[] getEncoded() {
     public int hashCode() {
         checkDestroyed();
 
-        int retval = 0;
-        for (int i = 1; i < this.key.length; i++) {
-            retval += this.key[i] * i;
-        }
-        return (retval ^= "desede".hashCode());
+        SecretKeySpec keySpec = new SecretKeySpec(this.key, getAlgorithm());
+        return keySpec.hashCode();
```
johnpeck-us-ibm, line 80: Note: SecretKeySpec.hashcode is basically do the same as the old code.

src/main/java/com/ibm/crypto/plus/provider/DSAPrivateKey.java
```diff
@@ -249,38 +248,6 @@ private void checkDestroyed() {
...
-                }
-            }
-        } catch (Exception e1) {
-            //Should never get here
-            //System.out.println("Object = Exception - " + e1.toString());
-        }
-        return false;
-    }
-
     public static boolean equals(DSAParams spec1, DSAParams spec2) {
```
johnpeck-us-ibm, line 251: The added equals is not the same as the removed equals.  With the above one removed the equals in PKCS8Key will be used and that is not the same comparison. I am worried that there will be a case  where the use of PKCS8Key's equals(object) will not give the desired result.

src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
```diff
@@ -681,36 +680,4 @@ private void checkDestroyed() {
...
-                    return true;
-                }
-            }
-        } catch (Exception e1) {
-            //Should never get here
-            //System.out.println("Object = Exception - " + e1.toString());
-        }
-        return false;
-    }
-
```
johnpeck-us-ibm, line 715: With the above one removed the equals in PKCS8Key will be used and that is not the same comparison. I am worried that there will be a case where the use of PKCS8Key's equals(object) will not give the desired result.

jasonkatonica (approved):

src/main/java/com/ibm/crypto/plus/provider/EdDSAKeyFactory.java
johnpeck-us-ibm: I am okay with this change. However, this seems to be a lot of changes unrelated to the issue being worked.

src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
```diff
@@ -295,7 +294,7 @@ public Optional<byte[]> getBytes() {
         } catch (Exception exception) {
             this.exception = exception;
         }
-        return this.h;
+        return Optional.of(this.h);
```
johnpeck-us-ibm, line 297: See above.

johnpeck-us-ibm (changes requested):

src/main/java/com/ibm/crypto/plus/provider/AESKey.java
KostasTsiounis: It only does the same in versions 21 and earlier. The idea is that we can back port the same code to all versions and pick up the appropriate implementation from the JDK instead of keep updating these parts.

src/main/java/com/ibm/crypto/plus/provider/ChaCha20Key.java
```diff
@@ -73,11 +74,8 @@ public synchronized byte[] getEncoded() {
...
         checkDestroyed();
 
-        int retval = 0;
-        for (int i = 1; i < this.key.length; i++) {
-            retval += this.key[i] * i;
-        }
-        return (retval ^= "ChaCha20".hashCode());
+        SecretKeySpec keySpec = new SecretKeySpec(this.key, getAlgorithm());
+        return keySpec.hashCode();
     }
```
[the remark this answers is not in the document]
KostasTsiounis: See https://github.com/IBM/OpenJCEPlus/pull/682#discussion_r2190885449

src/main/java/com/ibm/crypto/plus/provider/DSAPrivateKey.java
KostasTsiounis: Having this equals is causing interoperability issues with the Sun providers. I am not sure as to why we need a separate `equals()` given that the encodings have been updated to be correct per version.

src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
KostasTsiounis: Same as previous.

src/main/java/com/ibm/crypto/plus/provider/DSAPrivateKey.java
KostasTsiounis: The use case  is that we use the method `equals()` to compare two objects created by different providers (e.g., generate with `OpenJCEPlus` and get through `KeyFactory` from `Sun`). The method calculates the encodings at runtime using the object values rather than use a pre-saved encoding. It uses the internal `getEncodedInternal()` method (see https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/3f587164ce3ffee516de98ed8ca7d8719b0b6613/src/java.base/share/classes/sun/security/pkcs/PKCS8Key.java#L295) that will return the same result for the same key regardless of who created the object, given that we save the same data as expected.

Given that we also extend the `PKCS8Key` class, our code will have the same behaviour as the Sun one. But if we override `equals()` we have different logic that doesn't interoperate with `Sun` since the parameters are created in a slightly different way that is not specifically defined (as opposed to the RFCs for the encodings).

I don't believe there is a compatibility issue here, since the method recalculates the encoding from the individual information rather than using existing encodings. So, it doesn't matter if someone has already saved an encoding. That encoding will be parsed, an object will be created and the encoding needed for `equals()` will be recalculated and will have the same format, provided that everyone extends the `PKCS8Key` class.

As far as existing product code, if someone is using `equals()` to compare 2 keys and it works for them, it should work the exact same way, as the 2 keys will now use the `PKCS8Key`'s `equals()` that will produce their encodings and verify that they are the same.

I can't think of any other scenario, but maybe I'm missing something.

src/main/java/com/ibm/crypto/plus/provider/AESKey.java
KostasTsiounis: Yes, that is correct. In versions below 21, this logic doesn't change at all. Using `SecretKeySpec` just makes the code consistent over different versions.

PR #723, opened 2025-07-16
Fix key compatibility problem in XDHKeyAgreement

dev-koan:
The KeyAgreement test was failing because the key passed in the parameter was an instance of `Sun`.  Resolved by using `XDHKeyFactory` for private and public keys passed as parameters in `engineInit()` and `engineDoPhase()`. Added a check to throw `InvalidKeyException` when the key is not an instance of  `XECPrivateKey` or `XECPublicKey` as required. 

Consistency is also ensured between the two function with respect to the order that the checks are taking place.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
@@ -15,12 +15,12 @@ import com.ibm.crypto.plus.provider.ock.XECKey;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.Key;
-import java.security.KeyFactory;
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
+import java.security.interfaces.XECPrivateKey;
+import java.security.interfaces.XECPublicKey;
 import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.NamedParameterSpec;
-import java.security.spec.XECPrivateKeySpec;
 import javax.crypto.KeyAgreementSpi;
 import javax.crypto.SecretKey;
 import javax.crypto.ShortBufferException;
@@ -65,11 +65,22 @@ abstract class XDHKeyAgreement extends KeyAgreementSpi {
     @Override
     protected Key engineDoPhase(Key key, boolean lastPhase)
             throws InvalidKeyException, IllegalStateException {
-        if (!(key instanceof XDHPublicKeyImpl))
-            throw new InvalidKeyException("Key is not an XDHPublicKeyImpl");
+        if (!(key instanceof XECPublicKey)) {
+            throw new InvalidKeyException("Unsupported key type");
+        }
         if (ockXecKeyPriv == null)
             throw new IllegalStateException(
                     "object is not initialized correctly (private key is not received)");
+
+        if (!(key instanceof XDHPublicKeyImpl)) {
+            try {
+                key = (XDHPublicKeyImpl) XDHKeyFactory.toXECKey(this.provider, this.alg, key);
+            } catch (ClassCastException cce) {
+                throw new InvalidKeyException("Translated key is not an instance of XDHPublicKeyImpl", cce);
+            } catch (Exception exception) {
+                throw new InvalidKeyException("Unable to translate key", exception);
+            }
+        }
 
         XDHPublicKeyImpl xdhPublicKeyImpl = (XDHPublicKeyImpl) key;
 
@@ -218,6 +229,20 @@ abstract class XDHKeyAgreement extends KeyAgreementSpi {
     @Override
     protected void engineInit(Key key, AlgorithmParameterSpec params, SecureRandom random)
             throws InvalidKeyException, InvalidAlgorithmParameterException {
+        
+        if (!(key instanceof XECPrivateKey)) {
+            throw new InvalidKeyException("Unsupported key type");
+        }
+
+        if (!(key instanceof XDHPrivateKeyImpl)) {
+            try {
+                key = (XDHPrivateKeyImpl) XDHKeyFactory.toXECKey(this.provider, this.alg, key);
+            } catch (ClassCastException cce) {
+                throw new InvalidKeyException("Translated key is not an instance of XDHPrivateKeyImpl", cce);
+            } catch (Exception exception) {
+                throw new InvalidKeyException("Unable to translate key", exception);
+            }
+        }
 
         // Check if parameter is a valid NamedParameterSpec instance
         if (params != null) {
@@ -228,17 +253,6 @@ abstract class XDHKeyAgreement extends KeyAgreementSpi {
                 }
             } else {
                 throw new InvalidAlgorithmParameterException("Invalid Parameters: " + params);
-            }
-        }
-
-        if (!(key instanceof XDHPrivateKeyImpl)) {
-            try {
-                KeyFactory kf = KeyFactory.getInstance(this.alg);
-                XECPrivateKeySpec spec = kf.getKeySpec(key, XECPrivateKeySpec.class);
-                key = kf.generatePrivate(spec);
-            } catch (Exception exception) {
-                // should not happen
-                throw new InvalidKeyException("KeyFactory is not working as expected");
             }
         }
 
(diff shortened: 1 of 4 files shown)


dev-koan:
> Is there an issue to look at for this? This seems like a very normal issue for ANY key that comes from a different provider than the one that key is being used with. I do no think this issue is only for these keys.

Yes, there is an issue that the PR addresses (https://github.ibm.com/runtimes/jit-crypto/issues/839).
It is a normal problem that exists when the key comes from a different provider.

dev-koan:
> Okay so the issue is that the key that was passed into the OpenJCEPlus provider was created by the SUN provider.
> 
> I doubt the test was meant to cross test providers. Meaning that a key was created by one provider and than passed to another provider to verify compatibility and function. This sounds like a test case issue or a test set up issue. Please check the test case and determine why the key was created by the SUN provider and not the OpenJCEPlus provider. Since, this test was for XDHKeyAgreement I would have expected ALL keys to be created by the same provider or at least all operations including key creation to be done separately on each side.

I looked into the test case and you are correct, the test was not meant to cross test providers. When it gets key pair ([link](https://github.ibm.com/runtimes/JCK11-unzipped/blob/main/JCK-runtime-11a/src/tests/api/javax_crypto/KeyAgreement/KeyAgreementBase.java#L325)), the function `checkKPGSupportAndGetKeyPair` doesn't have a parameter for the provider, so it must be using `Sun` as default. But when an instance of `KeyAgreement` is created, the provider is specified causing a mismatch. 
This can either be resolved by changing the test case (but it is a JCK test) or the provider code.

johnpeck-us-ibm:
Are you sure this is a JCK test? If it is try adding OpenJCEPlus first in the provider list and rerun the test.

johnpeck-us-ibm:
I have talked with the Java 8 JS team and that caused me to wonder why the issue is only for XDH and not DH or ECDH etc. The code in question seems to be able to test ALL Key Agreement algs.

johnpeck-us-ibm:
Interesting but I guess not surprising since the issues you show indicate that the failure is in the Native code version of SunEC. The actual code for this version of the SunEC provider does not seem to have been open sourced. So we can not determine what the SunEC code was looking for in order for this to work.

Now, if you put OpenJCEPlus last in the list and add code to do the translate that will only fix that scenario.

The real thing we want to do is to put OpenJCEPlus first in the list and have the JCK work. However, doing that will cause the at least the Native SunEC provider to have issues with the keys from OpenJCEPlus.

Therefore, I think the best thing to do for this issue is to go back to Oracle to have them remove the test for now and to correct it in the future.

@jasonkatonica Lets discuss this when you return from vacation.

johnpeck-us-ibm (changes requested):
Code is fine.
However, we need an interop test case to be created to verify that this code is working as expected.

Also, I want to voice comcern, Doing this can set an expectation that all operations in OpenJCEPlus have a similar feature. So, if we do this here. We should probably do this for ALL operations: Key Agreement, signnatures, Ciphers, etc.

KostasTsiounis:
> It is also strange that all platforms are failing EXCEPT for AIX which is not expected.

This might be failing due to the interop problem on the Sun side. The `NativeXDHKeyPairGenerator`, which lacks the ability to handle non-Sun keys at the moment, is used by all platforms except `AIX`.

src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
```diff
@@ -73,12 +73,23 @@ abstract class XDHKeyAgreement extends KeyAgreementSpi {
...
+        }
         if (ockXecKeyPriv == null)
             throw new IllegalStateException(
                     "object is not initialized correctly (private key is not received)");
 
+        if (!(key instanceof XDHPublicKeyImpl)) {
+            try {
+                key = (XDHPublicKeyImpl) XDHKeyFactory.toXECKey(this.provider, this.alg, key);
+            } catch (ClassCastException cce) {
+                throw new InvalidKeyException("Key is not an instance of XDHPublicKeyImpl", cce);
```
KostasTsiounis, line 87: It is not clear in this exception message which key is not an instance of `XDHPublicKeyImpl`. It sounds as if the user provided key has an issue, when it's the one produced by the factory.

src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
```diff
@@ -226,6 +237,20 @@ protected void engineInit(Key key, SecureRandom random) throws InvalidKeyExcepti
...
+        
+        if (!(key instanceof XECPrivateKey)) {
+            throw new InvalidKeyException("Unsupported key type");
+        }
+
+        if (!(key instanceof XDHPrivateKeyImpl)) {
+            try {
+                key = (XDHPrivateKeyImpl) XDHKeyFactory.toXECKey(this.provider, this.alg, key);
+            } catch (ClassCastException cce) {
+                throw new InvalidKeyException("Key is not an instance of XDHPrivateKeyImpl", cce);
```
KostasTsiounis, line 249: Same comment as above.

src/test/java/ibm/jceplus/junit/base/BaseTestXDHKeyAgreementInterop.java
```diff
@@ -0,0 +1,49 @@
...
+        KeyPair kp2 = kpg2.generateKeyPair();
+
+        ka1.init(kp1.getPrivate());
+        ka1.doPhase(kp2.getPublic(), true);
+        
+        ka2.init(kp2.getPrivate());
+        ka2.doPhase(kp1.getPublic(), true);
+
+        byte[] ss1 = ka1.generateSecret();
+        byte[] ss2 = ka2.generateSecret();
```
KostasTsiounis, line 45: I would add some more combinations. In the ones you have now the KeyAgreement instance and private key are always coming from the same provider. I think we should try the inverse too where the private key comes from the other provider.

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
johnpeck-us-ibm: There is no KeyFactory involved here. In this code it has not idea what has or has not been done to the key. It is just making sure that the passed in key is what it expects.

src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
```diff
@@ -73,12 +73,23 @@ abstract class XDHKeyAgreement extends KeyAgreementSpi {
...
             throw new IllegalStateException(
                     "object is not initialized correctly (private key is not received)");
 
+        if (!(key instanceof XDHPublicKeyImpl)) {
+            try {
+                key = (XDHPublicKeyImpl) XDHKeyFactory.toXECKey(this.provider, this.alg, key);
+            } catch (ClassCastException cce) {
+                throw new InvalidKeyException("Key is not an instance of XDHPublicKeyImpl", cce);
+            } catch (Exception exception) {
+                throw new InvalidKeyException("KeyFactory is not working as expected", exception);
```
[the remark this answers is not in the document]
dev-koan: Updated.

src/main/java/com/ibm/crypto/plus/provider/XDHKeyAgreement.java
```diff
@@ -226,6 +237,20 @@ protected void engineInit(Key key, SecureRandom random) throws InvalidKeyExcepti
...
+            throw new InvalidKeyException("Unsupported key type");
+        }
+
+        if (!(key instanceof XDHPrivateKeyImpl)) {
+            try {
+                key = (XDHPrivateKeyImpl) XDHKeyFactory.toXECKey(this.provider, this.alg, key);
+            } catch (ClassCastException cce) {
+                throw new InvalidKeyException("Key is not an instance of XDHPrivateKeyImpl", cce);
+            } catch (Exception exception) {
+                throw new InvalidKeyException("KeyFactory is not working as expected", exception);
```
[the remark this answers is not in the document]
dev-koan: Updated.

jasonkatonica (approved):
LGTM

johnpeck-us-ibm (approved):

PR #745, opened 2025-09-03
Throw exception in case of incorrect EC key size

KostasTsiounis:
If the user provides an EC key size that doesn't correspond to one of the available curves, throw an exception instead of defaulting to 256.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/737


jasonkatonica (approved):
LGTM

johnpeck-us-ibm (approved):

PR #764, opened 2025-09-05
Don't use agent for top level jenkins job

### LINKED ISSUE #763: Jenkins Pipeline uses "agent any" and holds it during parallel execution
I noticed the Jenkins jobs running are landing on nodes that are reserved for specific workloads. This is because the syntax used is `agent any` here
https://github.com/IBM/OpenJCEPlus/blob/45fb329a673b0b04c9ce86c700d5f9878255f8c2/Jenkinsfile#L649C5-L649C14

Furthermore, the parallel part of the build is kicked off before that first node is released so it continues to hold the executor on that node but no real work is being done on it.

https://github.com/IBM/OpenJCEPlus/blob/45fb329a673b0b04c9ce86c700d5f9878255f8c2/Jenkinsfile#L698

I only have experience with scripted pipeline syntax so I'm not entirely sure if/how you would code it in declarative syntax. If it was scripted, it would be something like this

```
node(labels) {
    // setup parallel stuff
}
parallel mapForParallel
```

I believe it is possible to run `agent none` which might work in this case because there doesn't seem to be any real work happening before the parallel is kicked off. Another option could be using a [stage agent](https://www.jenkins.io/doc/book/pipeline/syntax/#stage-level-agents).

AdamBrousseau:
Fixes #763

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -646,7 +646,7 @@ pipeline {
             Overall build timeout (HOURS)')
     }
 
-    agent any
+    agent none
     stages {
         stage('Build And Test OpenJCEPlus') {
             steps {


AdamBrousseau:
image (3)
Looks good to me.

jasonkatonica:
@AdamBrousseau Could you close this PR and instead refer to the same change using https://github.com/IBM/OpenJCEPlus/pull/767

PR #794, opened 2025-09-18
Make use of API to fetch Java 25

jasonkatonica:
Releases of Java 25 are now available and this update will make use
of the latest Java 25 ga build.


jasonkatonica:
Updated to use GA version of temurin in github actions instead of early access.

Issue #814, opened 2025-09-29
Restoring PQCPublicKey object from encoding fails to allow signature verification


jasonkatonica:
When the following test is run on Java 25 everything works as expected:
```
import java.security.*;
import java.security.spec.*;
import java.util.HexFormat;

/**
 * ML-DSA Digital Signature Demo.
 */
public class MLDSASignatureDemo {

    private static final String ML_DSA_ALGORITHM = "ML-DSA-87";
    private static final String PROVIDER = "OpenJCEPlus";
    private static final byte[] MESSAGE = "thequicklazycatjumpedoverthemoonifitwantedto".getBytes();
    private static final boolean DEBUG_BYTES = true;

    /**
     * The encoding for the public key produced by Bob and consumed by Alice.
     */
    private static byte[] publicKeyBytes = null;

    /**
     * The signature produced by Bob and Consumed by Alice.
     */
    private static byte[] signature = null;

    public static void main(String[] args) throws Exception {
        Provider myProvider = (Provider) Class.forName("com.ibm.crypto.plus.provider.OpenJCEPlus")
                .getDeclaredConstructor().newInstance();
        Security.insertProviderAt(myProvider, 1);

        System.out.println("=".repeat(50));
        System.out.println(ML_DSA_ALGORITHM + " PQC Demo");
        System.out.println("Message length: " + MESSAGE.length + " bytes");
        System.out.println("=".repeat(50));

        BobProcessing();
        // Transmission occured here of fields "publicKeyBytes" and "signature"
        AliceProcessing();
    }

    private static void BobProcessing() throws Exception {
        // Generate ML-DSA key pair
        KeyPairGenerator keyPairGenerator = KeyPairGenerator.getInstance(ML_DSA_ALGORITHM,
                PROVIDER);
        KeyPair keyPair = keyPairGenerator.generateKeyPair();
        PrivateKey privateKey = keyPair.getPrivate();
        PublicKey originalPublicKey = keyPair.getPublic();
        System.out.println("✓ Key pair generated successfully");
        printKeyInfo(privateKey, originalPublicKey);

        // Sign the MESSAGE. Save the signature in private field above so we can send to Alice.
        Signature signer = Signature.getInstance(ML_DSA_ALGORITHM, PROVIDER);
        signer.initSign(privateKey);
        signer.update(MESSAGE);
        signature = signer.sign();
        System.out.println("✓ Data signed successfully");
[...]

PR #815, opened 2025-09-29
Always store human readable alg name in PQC keys

jasonkatonica:
The algorithm name at times stores the OID instead of the human readable name when running on releases other then Java 25. This update uses our internal maps to always convert the OID to the human readable name within the PQC key objects.

Fixes [#814](https://github.com/IBM/OpenJCEPlus/issues/814)

### DIFF
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestPQCSignature.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestPQCSignature.java
@@ -8,13 +8,17 @@
 
 package ibm.jceplus.junit.base;
 
+import java.security.KeyFactory;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.X509EncodedKeySpec;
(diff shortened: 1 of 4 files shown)


src/test/java/ibm/jceplus/junit/base/BaseTestPQCSignature.java
```diff
@@ -27,12 +32,32 @@ public void testPQCKeySignature(String Algorithm) throws Exception {
...
-            throw new Exception(e.getCause() +" - "+Algorithm, e);
+        KeyPair keyPair = generateKeyPair(Algorithm);
+        doSignVerify(Algorithm, origMsg, keyPair.getPrivate(), keyPair.getPublic());
+    }
+
+    @ParameterizedTest
+    @CsvSource({"ML_DSA_44","ML-DSA-65","ML_DSA_87"})
+    public void testPQCKeySignatureEncodings(String Algorithm) throws Exception {
+
+        if (getProviderName().equals("OpenJCEPlusFIPS")) {
```
KostasTsiounis, line 43: Do we need this check?

jasonkatonica: Sure once we have this in the FIPS provider then this test should also execute without the if statement for FIPS 140-3 ideally.

PR #819, opened 2025-10-01
Always store human readable alg name in PQC keys

jasonkatonica:
The algorithm name at times stores the OID instead of the human readable name when running on releases other then Java 25. This update uses our internal maps to always convert the OID to the human readable name within the PQC key objects.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/815


KostasTsiounis (approved):
LGTM

PR #822, opened 2025-10-02
Add JMH performance benchmarks

jasonkatonica:
JMH performance updates are added to aid in comparing the SDK we are running with vs what is being delivered in OpenJCEPlus provider.

A new Jenkins performce pipeline is provided to ensure that we have the capability when desired to performance test a given update present in a PR.

A new common groovy library was created for logic that is common between the existing Jenkins pipeline that runs the builds and tests and the new performance specific Jenkins pipeline.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/671

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@
 # Allow tracked changes for configuration file used by the
 # clang-format tool for formatting C code.
 !.clang-format
+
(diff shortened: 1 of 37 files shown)


KostasTsiounis (approved):
LGTM

PR #826, opened 2025-10-03
Disable a few EC curves in computing shared secret

JinhangZhang:
According to the policy, curve size less than 192 should not be allowed in computing shared secret in ECDH keyagreement in FIPS 140-3 mode.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/ECDHKeyAgreement.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ECDHKeyAgreement.java
@@ -26,6 +26,7 @@ import javax.crypto.KeyAgreementSpi;
 import javax.crypto.SecretKey;
 import javax.crypto.ShortBufferException;
 import javax.crypto.spec.SecretKeySpec;
+import sun.security.util.ObjectIdentifier;
 
 public final class ECDHKeyAgreement extends KeyAgreementSpi { // implements
                                                               // AlgorithmStatus
@@ -44,6 +45,7 @@ public final class ECDHKeyAgreement extends KeyAgreementSpi { // implements
     private ECPublicKey ecPublicKey = null;
     private ECPrivateKey ecPrivateKey = null;
     private int secretLen;
+    private static boolean disableSmallCurve = Boolean.parseBoolean(System.getProperty("openjceplus.disableSmallerECKeySizeForSharedKeyComputing", "true"));
 
     public ECDHKeyAgreement(OpenJCEPlusProvider provider) {
         // System.out.println ("In ECDHKeyAgreement");
@@ -162,6 +164,29 @@ public final class ECDHKeyAgreement extends KeyAgreementSpi { // implements
     protected byte[] engineGenerateSecret() throws IllegalStateException {
         if (generateSecret == false) {
             throw new IllegalStateException("Wrong state");
+        }
+
+        if (provider.isFIPS() && disableSmallCurve) {
+            ECNamedCurve ecPubKeyNamedCurve = ECParameters.getNamedCurve(this.ecPublicKey.getParams());
+            ECNamedCurve ecPriKeyNamedCurve = ECParameters.getNamedCurve(this.ecPrivateKey.getParams());
+            if (ecPubKeyNamedCurve == null || ecPriKeyNamedCurve == null) {
+                throw new IllegalStateException("Curve name is not recognized or not supported");
+            }
+            ObjectIdentifier oidPubKey = ECNamedCurve.getOIDFromName(ecPubKeyNamedCurve.getName());
+            ObjectIdentifier oidPriKey = ECNamedCurve.getOIDFromName(ecPriKeyNamedCurve.getName());
+
+            // check if curve is FIPS allowance and if it is allowed for ECDH shared secret computation
+            if (!((oidPubKey.toString()).equals("1.2.840.10045.3.1.7") // secp256r1 [NIST P-256, X9.62 prime256v1]
+                || (oidPubKey.toString()).equals("1.3.132.0.33") // secp224r1 [NIST P-224]
+                || (oidPubKey.toString()).equals("1.3.132.0.34") // secp384r1 [NIST P-384]
+                || (oidPubKey.toString()).equals("1.3.132.0.35")) ||
+                !((oidPriKey.toString()).equals("1.2.840.10045.3.1.7") // secp256r1 [NIST P-256, X9.62 prime256v1]
+                || (oidPriKey.toString()).equals("1.3.132.0.33") // secp224r1 [NIST P-224]
+                || (oidPriKey.toString()).equals("1.3.132.0.34") // secp384r1 [NIST P-384]
+                || (oidPriKey.toString()).equals("1.3.132.0.35")) // secp521r1 [NIST P-521]
+            ) {
+                throw new IllegalStateException(ecPubKeyNamedCurve.getName() + " curve is not supported in FIPS for calculating the shared secret");
+            }
         }
 
         // Reset the key agreement here (in case anything goes wrong)
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestECDH.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestECDH.java
@@ -44,13 +44,13 @@ import static org.junit.jupiter.api.Assertions.fail;
 public class BaseTestECDH extends BaseTestJunit5 {
 
     static final byte[] origMsg = "this is the original message to be signed".getBytes();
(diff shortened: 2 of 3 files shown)


jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/ECDHKeyAgreement.java
```diff
@@ -164,6 +166,31 @@ protected byte[] engineGenerateSecret() throws IllegalStateException {
...
+            if (!ECNamedCurve.isFIPS(oidPriKey.toString())) {
+                throw new IllegalStateException(ecPriKeyNamedCurve.getName() + " curve is not supported in FIPS");
+            }
+
+            if (!((oidPubKey.toString()).equals("1.2.840.10045.3.1.7") // secp256r1 [NIST P-256, X9.62 prime256v1]
+                || (oidPubKey.toString()).equals("1.3.132.0.33") // secp224r1 [NIST P-224]
+                || (oidPubKey.toString()).equals("1.3.132.0.34") // secp384r1 [NIST P-384]
+                || (oidPubKey.toString()).equals("1.3.132.0.35")) // secp521r1 [NIST P-521]
+            ) {
+                throw new IllegalStateException(ecPriKeyNamedCurve.getName() + " curve is not supported in FIPS for calculating the shared secret");
```
KostasTsiounis: Why do we need the previous checks if we have this, given that this class only calculates shared secrets?

src/main/java/com/ibm/crypto/plus/provider/ECDHKeyAgreement.java
```diff
@@ -164,6 +166,31 @@ protected byte[] engineGenerateSecret() throws IllegalStateException {
             throw new IllegalStateException("Wrong state");
         }
 
+        if (provider.isFIPS() && disableSmallCurve) {
+            ECNamedCurve ecPubKeyNamedCurve = ECParameters.getNamedCurve(this.ecPublicKey.getParams());
+            ECNamedCurve ecPriKeyNamedCurve = ECParameters.getNamedCurve(this.ecPrivateKey.getParams());
+            ObjectIdentifier oidPubKey = null;
```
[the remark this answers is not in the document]
JinhangZhang: updated

src/main/java/com/ibm/crypto/plus/provider/ECDHKeyAgreement.java
```diff
@@ -164,6 +166,31 @@ protected byte[] engineGenerateSecret() throws IllegalStateException {
...
+            ObjectIdentifier oidPriKey = null;
+            oidPubKey = ECNamedCurve.getOIDFromName(ecPubKeyNamedCurve.getName());
+            oidPriKey = ECNamedCurve.getOIDFromName(ecPriKeyNamedCurve.getName());
+
+            // check if curve is FIPS allowance and if it is allowed for ECDH shared secret computation.
+            if (!ECNamedCurve.isFIPS(oidPubKey.toString())) {
+                throw new IllegalStateException(ecPubKeyNamedCurve.getName() + " curve is not supported in FIPS");
```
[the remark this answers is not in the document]
JinhangZhang: updated

src/main/java/com/ibm/crypto/plus/provider/ECDHKeyAgreement.java
```diff
@@ -164,6 +166,33 @@ protected byte[] engineGenerateSecret() throws IllegalStateException {
             throw new IllegalStateException("Wrong state");
         }
 
+        if (provider.isFIPS() && disableSmallCurve) {
+            ECNamedCurve ecPubKeyNamedCurve = ECParameters.getNamedCurve(this.ecPublicKey.getParams());
+            ECNamedCurve ecPriKeyNamedCurve = ECParameters.getNamedCurve(this.ecPrivateKey.getParams());
+            ObjectIdentifier oidPubKey = ECNamedCurve.getOIDFromName(ecPubKeyNamedCurve.getName());
+            ObjectIdentifier oidPriKey = ECNamedCurve.getOIDFromName(ecPriKeyNamedCurve.getName());
```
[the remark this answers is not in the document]
JinhangZhang: added

src/main/java/com/ibm/crypto/plus/provider/ECDHKeyAgreement.java
KostasTsiounis: > The restriction is only applied for FIPS 140-3 mode

What I mean is that we check the OIDs with `ECNamedCurve.isFIPS()` and then we check again that the OIDs are one of 4 allowed. Aren't those 4 a subset of what the `ECNamedCurve.isFIPS()` would allow? And if yes, aren't the initial checks unnecessary? Unless I'm missing something in what we are trying to check.

JinhangZhang: Some of the fips approved EC curves from ECNamedCurve.isFIPS() are less than 256, e.g. `secp224r1 [NIST P-224]` and `secp192r1 [NIST P-192, X9.62 prime192v1]`. Both of them can not be used in shared secret computation. Therefore, I have to hardcoded those 4 curves in the list.

src/main/java/com/ibm/crypto/plus/provider/ECDHKeyAgreement.java
```diff
@@ -164,6 +166,35 @@ protected byte[] engineGenerateSecret() throws IllegalStateException {
...
+        if (provider.isFIPS() && disableSmallCurve) {
+            ECNamedCurve ecPubKeyNamedCurve = ECParameters.getNamedCurve(this.ecPublicKey.getParams());
+            ECNamedCurve ecPriKeyNamedCurve = ECParameters.getNamedCurve(this.ecPrivateKey.getParams());
+
+            if (ecPubKeyNamedCurve == null || ecPriKeyNamedCurve == null) {
+                throw new IllegalStateException("Curve name is not recognized or not supported");
+            }
+
+            ObjectIdentifier oidPubKey = ECNamedCurve.getOIDFromName(ecPubKeyNamedCurve.getName());
+            ObjectIdentifier oidPriKey = ECNamedCurve.getOIDFromName(ecPriKeyNamedCurve.getName());
```
johnpeck-us-ibm: The other check is to make sure that the curves are the same and the OIDs are the same. They should not be be different.

JinhangZhang: I agree, either public or private, which one you think could be more reasonable should be left here?

johnpeck-us-ibm: The curve name and OID should mean the same thing. So., since the curve name is first I would use that to do the comparison to validate the keys are compatible.

johnpeck-us-ibm (changes requested):

src/main/java/com/ibm/crypto/plus/provider/ECDHKeyAgreement.java
JinhangZhang: It seems that ecPubKeyNamedCurve.getName() will return `secp224r1` but will return `NIST P-256`/`NIST P-384`/`NIST P-521` for others, open another issue to track this different behaviour. In this case, will use the OID instead

src/test/java/ibm/jceplus/junit/base/BaseTestECDH.java
```diff
@@ -70,7 +70,7 @@ synchronized static void generateParams(String provider_name) {
         try {
 
             //String provider_name = "OpenJCEPlus";
-            String curveName_192k1 = "secp192k1";
+            String curveName_192k1 = "secp192r1";
```
KostasTsiounis: I think we should change the variable name given that we changed the curve.

src/test/java/ibm/jceplus/junit/base/BaseTestECDH.java
```diff
@@ -333,7 +346,7 @@ void compute_ecdh_key_with_global_key(String idString, AlgorithmParameterSpec al
         final String methodName = "compute_ecdh_key_with_global_key" + "_" + idString;
         KeyPair keyPairA = null, keyPairB = null;
         switch (idString) {
-            case "secp192k1":
+            case "secp192r1":
                 keyPairA = keyPairA_192k1;
```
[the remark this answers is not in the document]
JinhangZhang: removed

KostasTsiounis (approved):
LGTM

jasonkatonica (changes requested):

jasonkatonica (changes requested):

jasonkatonica (approved):
LGTM

PR #834, opened 2025-10-07
Refactor ECPrivateKey and allow getting public key

KostasTsiounis:
The `ECPrivateKey` class is refactored to consolidate a single point of performing encoding operations.

Functionality is added to allow saving the public key if provided with the private key encoding.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
@@ -23,8 +23,8 @@ import sun.security.pkcs.PKCS8Key;
 import sun.security.util.DerInputStream;
 import sun.security.util.DerOutputStream;
 import sun.security.util.DerValue;
-import sun.security.util.ObjectIdentifier;
 import sun.security.x509.AlgorithmId;
+import sun.security.x509.X509Key;
 
 final class ECPrivateKey extends PKCS8Key implements java.security.interfaces.ECPrivateKey {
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
```diff
@@ -216,323 +168,76 @@ final class ECPrivateKey extends PKCS8Key implements java.security.interfaces.EC
...
+     * The native library requires the parameters to be part of the encoding.
+     * If they existing encoding doesn't have them, add them. If it does
+     * contain them, check against the existing parameters.
+     *
+     * @return  the new encoding required by the native library
+     * @throws IOException
+     */
+    private byte[] createEncodedPrivateKeyWithParams() throws IOException {
+        DerInputStream privKeyBytesEncodedStream = new DerInputStream(this.privKeyMaterial);
+        DerValue[] inputDerValue = privKeyBytesEncodedStream.getSequence(4);
```
JinhangZhang: we have one additional value from the encoded?

src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
```diff
@@ -150,61 +103,60 @@ final class ECPrivateKey extends PKCS8Key implements java.security.interfaces.EC
         this.provider = provider;
 
         try {
-            parseKeyBits();
-        } catch (IOException e) {
-            throw new InvalidKeyException("parseKeyBits: " + e.getMessage());
-        }
+            // Set parameters.
```
johnpeck-us-ibm: Looks like the indention is off here.

KostasTsiounis: The comment is in the same indentation as the code that follows. It applies up until line 112 and not the whole block, so I can't get it outside of the `try`. Is that what you mean?

PR #847, opened 2025-10-14
Fixes for PQC test cases and Fix Serialization issues with keys.

johnpeck-us-ibm:
Needed to remove the dependency of JDK17 in the testcases. Also, there is an issue with serialization. When KeyRep is used to serialize the key the provider order is used to de serialize the key which can cause the wrong key to be created. So, added a new KeyRep class that contains the provider which will allow the correct key to be created.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/680

Signed-off-by: John Peck <[140550562+johnpeck-us-ibm@users.noreply.github.com](mailto:140550562+johnpeck-us-ibm@users.noreply.github.com)>

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESKey.java
@@ -9,7 +9,6 @@
 package com.ibm.crypto.plus.provider;
 
 import java.security.InvalidKeyException;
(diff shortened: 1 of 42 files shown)


JinhangZhang (approved):
LGTM

KostasTsiounis (approved):
LGTM

PR #853, opened 2025-10-16
Allow mvn test execution for entire windows family

jasonkatonica:
While we typically execute builds and testing for openjceplus on windows 2022 we would like to remove the restriction so they can be run on other machines also.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/844


KostasTsiounis (approved):
LGTM

taoliult (approved):

PR #864, opened 2025-10-17
Fix pipeline branch names for artifactory archival

jasonkatonica:
The branch name was always resolving to a null value causing builds with the same build number to upload on top of each other in various branches. This update seems to restore the correct branch name by using a different environment variable set by Jenkins checkout plugin.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/858


KostasTsiounis (approved):
LGTM

PR #873, opened 2025-10-21
Modify archive dir for build and perf pipelines

jasonkatonica:
The build and test pipeline and the performance pipeline currently upload to the same location in artifactory. If both the branch name and build number match they can overlay each others result.

This update adds another folder to differentiate more cleanly between the two pipeline results in artifactory.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/868


KostasTsiounis (approved):
LGTM

taoliult (approved):

PR #890, opened 2025-10-29
Implement cleaner to replace deprecated finalizers

sabrinajlee:
The cleaner is the replacement for the deprecated finalize() method for cleaning objects before garbage collection.

Each class that was previously using finalize() will have an anonymous function that cleans up the necessary resources. The number of cleaner threads can be set by a property with the default value being 2. The object and its anonymous function are registered to one of the cleaner threads in round robin order.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/845


jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

PR #895, opened 2025-10-30
Support OpenXL compiler for AIX

jasonkatonica:
This update migrates to the OpenXL compiler. This update is expected
only for Java 25 and higher.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -209,15 +209,16 @@ def run(platform) {
             // Some OSes have some further specific requirements.
             if (software == "aix") {
                 // Java 25 requires C++17.1 runtime. Otherwise crashes occur.
-                nodeTags += "&&sw.tool.c++runtime.17_1"
+                nodeTags = "hw.arch.${node_hardware}&&sw.os.aix.7_2&&sw.tool.c++runtime.17_1&&ci.role.build"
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -209,15 +209,16 @@ def run(platform) {
             // Some OSes have some further specific requirements.
             if (software == "aix") {
                 // Java 25 requires C++17.1 runtime. Otherwise crashes occur.
-                nodeTags += "&&sw.tool.c++runtime.17_1"
-            }
+                nodeTags = "sw.os.aix.7_2&&sw.tool.c++runtime.17_1&&ci.role.build"
```
KostasTsiounis: We don't have a hardware tag in this. Not sure if it will make any difference, but maybe just for completeness.

jasonkatonica: I could add this back in if we like however `sw.os.aix.7_2` does imply `hw.arch.ppc64` I can add it back in just so its similar to our other machine lists if that makes sense.

KostasTsiounis: Yeah, I thought of the same thing. Not necessary, but maybe good for consistency.

JinhangZhang (approved):
LGTM

PR #900, opened 2025-10-31
Support for PBES2, PBEKeyfactory, PBEKey and PBEParameters

dev-koan:
This update adds support for PBES2, PBEKeyFactory, PBEKey, and PBEParameters for PBE algorithms with Hmac and AES for OpenJCEPlus provider.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/833


dev-koan:
FYI -> Added encodePassword in PBEKey.java as this function is not implemented in PBEUtil.java in jdk21 and back.
-> Modified hashcode() to return the same value as jdk21 which differs from the newer jdk versions.
-> Added code in `PBES2Core.java` as `PBEUtils.java` doesn't exists in jdk17 and less.

README.md
```diff
@@ -276,170 +276,200 @@ The following properties can be used to configure application behavior at runtim
 
 The following algorithms are registered by the OpenJCEPlus and OpenJCEPlusFIPS providers.
 
-| Algorithm Type            | Algorithm Name             | OpenJCEPlusFIPS | OpenJCEPlus  | Notes        |
```
jasonkatonica: Can this table match the column widths of main branch. I do acknowledge that the `512/224`,`512/256` have been removed but the rest of the table i think has different formatting. Keeping the table as close as possible to other branches makes it easier to maintain.

dev-koan: Updated.

src/main/java/com/ibm/crypto/plus/provider/PBES2Core.java
```diff
@@ -0,0 +1,278 @@
...
+        } catch (InvalidAlgorithmParameterException ie) {
+            throw new InvalidKeyException("requires PBE parameters", ie);
+        }
+    }
+
+    protected void engineInit(int opmode, Key key,
+                              AlgorithmParameterSpec params,
+                              SecureRandom random)
+        throws InvalidKeyException, InvalidAlgorithmParameterException {
+        PBEKeySpec pbeSpec = pbes2Params.getPBEKeySpec(blkSize, keyLength,
```
jasonkatonica: There is a whitespace difference here on line 128 vs what is in `main` branch.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/PBES2Parameters.java
```diff
@@ -0,0 +1,464 @@
...
+            salt = specified.getOctetString();
+        } else {
+            // the 'otherSource' ASN.1 CHOICE for 'salt' is not supported
+            throw new IOException("PBE parameter parsing error: "
+                + "not an ASN.1 OCTET STRING tag");
+        }
+        iCount = pBKDF2_params.data.getInteger();
+
+        // keyLength INTEGER (1..MAX) OPTIONAL,
+        DerValue prf = null;
```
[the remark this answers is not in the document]
dev-koan: The `getOptional()` used in main is not supported in 17 and back hence i took this particular code block from jdk17.
The equivalent code of `getOptional()` can be added but i think it is a safer bet to take this block from openjdk

PR #905, opened 2025-10-31
Implement cleaner in key algorithms

sabrinajlee:
Removes the deprecated finalize() methods from the key classes and implements a runnable function to handle native memory cleanup in their place.

Related to: https://github.com/IBM/OpenJCEPlus/pull/845

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESKey.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023, 2024
+ * Copyright IBM Corp. 2023, 2025
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms provided by IBM in the LICENSE file that accompanied
@@ -40,10 +40,15 @@ final class AESKey implements SecretKey {
         if ((key == null) || !AESUtils.isKeySizeValid(key.length)) {
             throw new InvalidKeyException("Wrong key size");
         }
+        if (provider == null) {
+            throw new IllegalArgumentException("provider is null");
+        }
 
         this.key = new byte[key.length];
         this.provider = provider;
         System.arraycopy(key, 0, this.key, 0, key.length);
+
+        this.provider.registerCleanable(this, cleanOCKResources(this.key));
     }
 
     @Override
@@ -151,21 +156,18 @@ final class AESKey implements SecretKey {
         }
     }
 
-    /**
-     * This function zeroizes the key so that it isn't in memory when GC is
-     * done.
-     */
-    @Override
-    protected void finalize() throws Throwable {
-        try {
-            synchronized (this) {
-                if (this.key != null) {
-                    Arrays.fill(this.key, (byte) 0x00);
-                    this.key = null;
+    private Runnable cleanOCKResources(byte[] key) {
+        return() -> {
+            try {
+                if (key != null) {
+                    Arrays.fill(key, (byte) 0x00);
+                }
+            } catch (Exception e){
+                if (OpenJCEPlusProvider.getDebug() != null) {
+                    OpenJCEPlusProvider.getDebug().println("An error occurred while cleaning : " + e.getMessage());
+                    e.printStackTrace();
                 }
             }
-        } finally {
-            super.finalize();
-        }
+        };
     }
 }
--- a/src/main/java/com/ibm/crypto/plus/provider/ChaCha20Key.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ChaCha20Key.java
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2023, 2024
+ * Copyright IBM Corp. 2023, 2025
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms provided by IBM in the LICENSE file that accompanied
@@ -41,9 +41,16 @@ final class ChaCha20Key implements SecretKey, ChaCha20Constants {
         if ((key == null) || (key.length != ChaCha20_KEY_SIZE)) {
             throw new InvalidKeyException("Wrong key size");
         }
+        
+        if (provider == null) {
+            throw new IllegalArgumentException("provider is null");
+        }
 
         this.key = new byte[key.length];
+        this.provider = provider;
         System.arraycopy(key, 0, this.key, 0, key.length);
+
+        this.provider.registerCleanable(this, cleanOCKResources(this.key));
     }
 
     @Override
@@ -151,21 +158,18 @@ final class ChaCha20Key implements SecretKey, ChaCha20Constants {
         }
     }
 
-    /**
-     * This function zeroizes the key so that it isn't in memory when GC is
-     * done.
-     */
-    @Override
-    protected void finalize() throws Throwable {
-        try {
-            synchronized (this) {
-                if (this.key != null) {
-                    Arrays.fill(this.key, (byte) 0x00);
-                    this.key = null;
+    private Runnable cleanOCKResources(byte[] key) {
+        return() -> {
+            try {
+                if (key != null) {
+                    Arrays.fill(key, (byte) 0x00);
+                }
+            } catch (Exception e){
+                if (OpenJCEPlusProvider.getDebug() != null) {
+                    OpenJCEPlusProvider.getDebug().println("An error occurred while cleaning : " + e.getMessage());
+                    e.printStackTrace();
                 }
             }
-        } finally {
-            super.finalize();
-        }
+        };
     }
 }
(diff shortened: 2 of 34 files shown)


src/main/java/com/ibm/crypto/plus/provider/AESKey.java
```diff
@@ -40,10 +40,15 @@ final class AESKey implements SecretKey {
...
         }
+        if (provider == null) {
+            throw new IllegalArgumentException("provider is null");
+        }
 
         this.key = new byte[key.length];
         this.provider = provider;
         System.arraycopy(key, 0, this.key, 0, key.length);
+
+        this.provider.registerCleanable(this, cleanResources(key));
```
jasonkatonica: Shouldn't this be `this.key` similar to the previous definition of `finalize()` ? Seems we are registering the cleaning of the wrong key variable.

src/main/java/com/ibm/crypto/plus/provider/ChaCha20Key.java
```diff
@@ -41,9 +41,16 @@ final class ChaCha20Key implements SecretKey, ChaCha20Constants {
...
+        
+        if (provider == null) {
+            throw new IllegalArgumentException("provider is null");
+        }
 
         this.key = new byte[key.length];
+        this.provider = provider;
         System.arraycopy(key, 0, this.key, 0, key.length);
+
+        this.provider.registerCleanable(this, cleanResources(key));
```
jasonkatonica: Shouldn't this be `this.key` similar to the previous definition of `finalize()` ? Seems we are registering the cleaning of the wrong key variable.

src/main/java/com/ibm/crypto/plus/provider/DESedeKey.java
```diff
@@ -40,12 +40,18 @@ final class DESedeKey implements SecretKey, Destroyable {
...
+        }
 
         this.key = new byte[DESedeKeySpec.DES_EDE_KEY_LEN];
+        this.provider = provider;
         System.arraycopy(key, 0, this.key, 0, DESedeKeySpec.DES_EDE_KEY_LEN);
         DESedeKeyGenerator.setParityBit(key, 0);
         DESedeKeyGenerator.setParityBit(key, 8);
         DESedeKeyGenerator.setParityBit(key, 16);
+
+        this.provider.registerCleanable(this, cleanResources(key));
```
jasonkatonica: Shouldn't this be `this.key` similar to the previous definition of `finalize()` ? Seems we are registering the cleaning of the wrong key variable.

src/main/java/com/ibm/crypto/plus/provider/ock/DHKey.java
```diff
@@ -91,18 +103,25 @@ public static DHKey createPublicKey(OCKContext ockContext, byte[] publicKeyBytes
...
+            byte[] publicKeyBytes, OpenJCEPlusProvider provider) {
         this.ockContext = ockContext;
         this.dhKeyId = dhKeyId;
         this.pkeyId = 0;
         this.parameters = parameters;
         this.privateKeyBytes = privateKeyBytes;
         this.publicKeyBytes = publicKeyBytes;
+        this.provider = provider;
+
+        this.provider.registerCleanable(this, cleanOCKResources(privateKeyBytes, dhKeyId, pkeyId, ockContext));
```
jasonkatonica: I believe we have to wrap all primative types in a wrapper class such as `pkeyID` and `dhKeyId` here? Searching through the code at least `this.pkeyId` is set after the `registerCleanable` in this spot and hence would not be cleaning what we think its cleaning? This comment applies to other areas of the code in this PR with similar pattern.

src/main/java/com/ibm/crypto/plus/provider/ock/PQCKey.java
```diff
@@ -194,4 +196,16 @@ public String toString() {
...
+
+    private Runnable cleanOCKResources(byte[] privateKeyBytes, long pkeyId, OCKContext ockContext){
+        return() -> {
+            if ((privateKeyBytes != null) && (privateKeyBytes != unobtainedKeyBytes)) {
+                Arrays.fill(privateKeyBytes, (byte) 0x00);
+            }
+
+            if (pkeyId != 0) {
+                NativeInterface.MLKEY_delete(ockContext.getId(), pkeyId);
+            }
```
jasonkatonica: Are we planning on following similar pattern as Message Digest where we do a try catch and a debug message?

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/ock/DHKey.java
```diff
@@ -28,7 +30,22 @@ public final class DHKey implements AsymmetricKey {
     private static final String badIdMsg1 = "Public or Private Key Identifier is not valid";
     private static final String debPrefix = "DHKey";
 
-    public static DHKey generateKeyPair(OCKContext ockContext, byte[] parameters)
+    public class LongWrapper {
```
KostasTsiounis: I think we should extract this class as it's used by multiple keys and might be used by more in the future.

src/main/java/com/ibm/crypto/plus/provider/ock/DHKey.java
```diff
@@ -227,32 +261,34 @@ private synchronized void obtainParameters() throws OCKException {
...
         // OCKDebug.Msg (debPrefix, methodName, "Id :" + id);
         return (id != 0L);
     }
+
+
+
+    private Runnable cleanOCKResources(byte[] privateKeyBytes, long dhKeyId, LongWrapper pkeyId, OCKContext ockContext) {
+        return () -> {
+            if ((privateKeyBytes != null) && (privateKeyBytes != unobtainedKeyBytes)) {
+                Arrays.fill(privateKeyBytes, (byte) 0x00);
```
KostasTsiounis: In `AESKey` this is in the `try` block as well. Should we move it here too? Same comment for other keys.

KostasTsiounis: Just putting it in the `try` block isn't enough since you are only catching `OCKException` there. So, we either need another block or to catch `Exception` in general there.

jasonkatonica:
Some style checks are failing as follows:
```
Error:  /home/runner/work/OpenJCEPlus/OpenJCEPlus/src/main/java/com/ibm/crypto/plus/provider/AESKey.java:17:1: Extra separation in import group before 'com.ibm.crypto.plus.provider.ock.OCKException' [ImportOrder]
Error:  /home/runner/work/OpenJCEPlus/OpenJCEPlus/src/main/java/com/ibm/crypto/plus/provider/AESKey.java:17:1: Wrong order for 'com.ibm.crypto.plus.provider.ock.OCKException' import. [ImportOrder]
Error:  /home/runner/work/OpenJCEPlus/OpenJCEPlus/src/main/java/com/ibm/crypto/plus/provider/AESKey.java:17:8: Unused import - com.ibm.crypto.plus.provider.ock.OCKException. [UnusedImports]
Error:  /home/runner/work/OpenJCEPlus/OpenJCEPlus/src/main/java/com/ibm/crypto/plus/provider/AESKey.java:165:17: 'if' child has incorrect indentation level 16, expected level should be 20. [Indentation]
```

src/main/java/com/ibm/crypto/plus/provider/ock/DHKey.java
sabrinajlee: Having a general catch for `Exception` seems like the better option since we probably expect to handle any exception the same way here.

src/main/java/com/ibm/crypto/plus/provider/PrimitiveWrapper.java
```diff
@@ -0,0 +1,41 @@
...
+ * under the terms provided by IBM in the LICENSE file that accompanied
+ * this code, including the "Classpath" Exception described therein.
+ */
+
+package com.ibm.crypto.plus.provider;
+
+public class PrimitiveWrapper {
+    public static class LongWrapper { 
+        long value;
+        public LongWrapper(long value) {
```
KostasTsiounis: I don't think you need to add wrapper after the type. You could just name in `Long` and then use it with `PrimitiveWrapper.Long`.

src/main/java/com/ibm/crypto/plus/provider/PrimitiveWrapper.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright IBM Corp. 2025
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms provided by IBM in the LICENSE file that accompanied
+ * this code, including the "Classpath" Exception described therein.
+ */
+
+package com.ibm.crypto.plus.provider;
+
```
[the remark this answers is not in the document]
sabrinajlee: Looks like it didn't pick up the change, the comment appears at line 12: 
```/**
 * This class is used by some algorithms that use the cleaner to clean up`
 * native resources. Primitive type variables are passed by value instead
 * of reference as parameters. There are some cases where a primitive member
 * variable is modified after registering the instance to the cleaner, in 
 * which case the cleaner may not have the updated value of the variable.
 * To handle this scenario, one of the inner classes of PrimitiveWrapper 
 * is used to allow the passing of a primitive variable by reference.
 */

Issue #920, opened 2025-11-05
Failures while writting DESede keys using keytool


jasonkatonica:
While writting a key using keytool the following failure occurs:
```
02:47:03  keytool error: java.security.KeyStoreException: Cannot invoke "com.ibm.crypto.plus.provider.OpenJCEPlusProvider.getName()" because "this.provider" is null
02:47:03  
02:47:03  java.security.KeyStoreException: Cannot invoke "com.ibm.crypto.plus.provider.OpenJCEPlusProvider.getName()" because "this.provider" is null
02:47:03  	at java.base/com.sun.crypto.provider.JceKeyStore.engineSetKeyEntry(JceKeyStore.java:294)
02:47:03  	at java.base/java.security.KeyStore.setKeyEntry(KeyStore.java:1194)
02:47:03  	at java.base/sun.security.tools.keytool.Main.doGenSecretKey(Main.java:1890)
02:47:03  	at java.base/sun.security.tools.keytool.Main.doCommands(Main.java:1191)
02:47:03  	at java.base/sun.security.tools.keytool.Main.run(Main.java:419)
02:47:03  	at java.base/sun.security.tools.keytool.Main.main(Main.java:400)
02:47:03  	at KeyToolTest.test(KeyToolTest.java:177)
02:47:03  	at KeyToolTest.test(KeyToolTest.java:145)
02:47:03  	at KeyToolTest.testOK(KeyToolTest.java:200)
02:47:03  	at KeyToolTest.testAll(KeyToolTest.java:401)
02:47:03  	at KeyToolTest.main(KeyToolTest.java:1842)
02:47:03  	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
02:47:03  	at java.base/java.lang.reflect.Method.invoke(Method.java:571)
02:47:03  	at com.sun.javatest.regtest.agent.MainWrapper$MainTask.run(MainWrapper.java:138)
02:47:03  	at java.base/java.lang.Thread.run(Thread.java:1485)
02:47:03  Caused by: java.lang.NullPointerException: Cannot invoke "com.ibm.crypto.plus.provider.OpenJCEPlusProvider.getName()" because "this.provider" is null
02:47:03  	at openjceplus/com.ibm.crypto.plus.provider.DESedeKey.writeReplace(DESedeKey.java:124)
[...]

PR #924, opened 2025-11-06
Set private provider field based upon constructor

jasonkatonica:
The DESedeKey and ChaCha20Key classes do allow for a provider to be used on construction of the key, however the provider is not saved to the private field of the class. This can cause NPE to be visable during serialization operations. This update simply saves the provider to the private field for later use.

Fixes: https://github.com/IBM/OpenJCEPlus/issues/920

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/921


KostasTsiounis (approved):
LGTM

taoliult (approved):

PR #938, opened 2025-11-07
Restrict EC curves in computing shared secret

JinhangZhang:
According to the policy, curve size less than 192 should not be allowed in computing shared secret in ECDH
keyagreement in FIPS 140-3 mode. This PR proposes an option to enable/disable this behaviour.

Backported-from: https://github.com/IBM/OpenJCEPlus/pull/826

Signed-off-by: JinhangZhang <Jinhang.Zhang@ibm.com>


KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

PR #941, opened 2025-11-07
Implement cleaner in key algorithms

sabrinajlee:
Removes the deprecated finalize() methods from the key classes and implements a runnable function to handle native memory cleanup in their place.

Related to: https://github.com/IBM/OpenJCEPlus/pull/845

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/905


src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -126,7 +126,7 @@ final class PQCPrivateKey extends PKCS8Key {
             }
 
             this.pqcKey = PQCKey.createPrivateKey(provider.getOCKContext(), 
-                                   this.name, this.key);
```
jasonkatonica: Leading whitespace seems different here then on 25.

PR #974, opened 2025-11-18
Fix AESGCMCipher key cleanup action

sabrinajlee:
The cleanup method in AESGCMCipher was not correctly zeroing out the key array due to an incorrect null reference passed as parameter. Putting a wrapper around the key variable fixes this issue.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/970


KostasTsiounis (approved):
LGTM

PR #991, opened 2025-11-20
Pass constructor parameters in Service initialization

KostasTsiounis:
Current code ignores parameters that need to be passed to the constructor of a service that needs to be initialized.

This change is rectifying this and ensuring parameters are appropriately forwarded to the service's constructor.

A test to verify this is added using KDF services, which is currently the only one that seems to take advantage of this.

Backported-from: https://github.com/IBM/OpenJCEPlus/pull/971


jasonkatonica (approved):
LGTM

PR #997, opened 2025-11-26
Support for PBES1 using native acceleration

dev-koan:
This update adds support for PBES1, PBEKeyFactory, and PBEParameters for PBES1 algorithms using native acceleration for OpenJCEPlus provider. PBE cipher tests have also been updated for better test coverage.


src/main/java/com/ibm/crypto/plus/provider/PBEParameters.java
```diff
@@ -0,0 +1,148 @@
...
+
+    public static final class SHA1AndDESede extends PBEParameters {
+        public SHA1AndDESede() throws NoSuchAlgorithmException {
+            super("PBEWithSHA1AndDESede");
+        }
+    }
+
+    public static final class SHA1AndRC2_40 extends PBEParameters {
+        public SHA1AndRC2_40() throws NoSuchAlgorithmException {
+            super("PBEWithSHA1RC2_40");
```
jasonkatonica: Do these names need to match the algorithm name? If so seems like this should be `PBEWithSHA1AndRC2_40` ?

src/main/native/PBE.c
```diff
@@ -0,0 +1,176 @@
...
+    if (NULL != text) {
+        (*env)->ReleasePrimitiveArrayCritical(env, algorithm, (void*)text, 0);
+        text = NULL;
+    }
+    if (NULL != saltNative) {
+        (*env)->ReleasePrimitiveArrayCritical(env, salt, saltNative, 0);
+        saltNative = NULL;
+    }
+    if (NULL != passwordNative) {
+        (*env)->ReleasePrimitiveArrayCritical(env, password,
```
jasonkatonica: I wasnt sure in this spot if we have a copy of the password or not allocated above with `GetPrimitiveArrayCritical`. From what i can tell this API may or may not be a copy so not sure we can zero out sensitive data ( password ) here.

PR #998, opened 2025-12-02
Silence compilation warnings

sabrinajlee:
This fix silences warnings that appear in necessary code paths.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -502,6 +502,7 @@
                     <compilerArgs>
                         <arg>-XDignore.symbol.file</arg>
                         <arg>-Xlint:all</arg>
+                        <arg>-Xlint:-processing</arg>
                         <arg>--add-exports </arg>
                         <arg>java.base/sun.security.internal.spec=openjceplus</arg>
                         <arg>--add-exports </arg>
--- a/src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java
@@ -170,6 +170,7 @@ final class NativeInterface {
         }
     }
 
+    @SuppressWarnings("restricted")
     private static boolean loadIfExists(File libraryFile) {
         String libraryName = libraryFile.getAbsolutePath();
 
--- a/src/main/java/ibm/security/internal/spec/RawKeySpec.java
+++ b/src/main/java/ibm/security/internal/spec/RawKeySpec.java
@@ -25,7 +25,8 @@ public class RawKeySpec implements KeySpec {
     /**
      * @param key contains the key as a byte array
      */
-
+    
+    @SuppressWarnings("this-escape")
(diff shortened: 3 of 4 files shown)


pom.xml
```diff
@@ -502,6 +502,7 @@
                     <compilerArgs>
                         <arg>-XDignore.symbol.file</arg>
                         <arg>-Xlint:all</arg>
+                        <arg>-Xlint:-processing</arg>
```
jasonkatonica: Not sure exactly what this is surpressing but using a universal option to supress additional warnings here might not be ideal. Perhaps we can make a different change or more narrowly supress the warning.

src/main/java/ibm/security/internal/spec/RawKeySpec.java
```diff
@@ -25,7 +25,8 @@ public class RawKeySpec implements KeySpec {
     /**
      * @param key contains the key as a byte array
      */
-
+    
+    @SuppressWarnings("this-escape")
     public RawKeySpec(byte[] key) {
         keyBytes = key.clone();
         cleaner.register(this, cleanOCKResources(keyBytes));
```
jasonkatonica: I guess there is a legitimate concern here that the runnable would finish before we construct `this`? Not sure if we have any other ideas on what might work here.

sabrinajlee: Here's the exact warning:
```[WARNING] /OpenJCEPlus/src/main/java/ibm/security/internal/spec/RawKeySpec.java:[30,25] [this-escape] possible 'this' escape before subclass is fully initialized```

File: pom.xml (hunk 1 above)
Comment (jasonkatonica) [reply pom.xml]: OK seems like these are all test related processing options for junit. I wonder if we could change the `pom.xml` file such that this option is only specified when the tests are compiled. Looking around I dont see any better options so far then what you have done here, besides trying to turn off processing for only the test code.

Comment (jasonkatonica) [reply pom.xml]: After second thought I am unsure its worth splitting logic in the pom.xml in order to target the test compilation vs the other classes in the src directory. For now we can keep it how you have it here.

File: src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java [hunk 3]
```diff
@@ -170,6 +170,7 @@ static void preloadOCK() {
         }
     }
 
+    @SuppressWarnings("restricted")
```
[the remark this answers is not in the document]
sabrinajlee: `Runtime.getRuntime().load(libraryName);` is still causing this restricted error, 

```[WARNING] /Users/sabrinalee/OpenJCEPlusTests/OpenJCEPlus/src/main/java/com/ibm/crypto/plus/provider/ock/NativeInterface.java:[182,36] [restricted] Runtime.load(String) is a restricted method.```

Not sure we can get around silencing this one.

PR #1004, opened 2025-12-05
Validate output buffer size

dev-koan:
This update validates the output buffer size. If the buffer is too small a ShortBufferException is thrown with the required buffer length.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESCipher.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESCipher.java
@@ -86,6 +86,12 @@ public final class AESCipher extends CipherSpi implements AESConstants {
         checkCipherInitialized();
 
         try {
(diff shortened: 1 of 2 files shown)


src/main/java/com/ibm/crypto/plus/provider/AESCipher.java
```diff
@@ -86,6 +86,11 @@ protected int engineDoFinal(byte[] input, int inputOffset, int inputLen, byte[]
         checkCipherInitialized();
 
         try {
+            if ((output == null) || ((output.length - outputOffset) < engineGetOutputSize(inputLen))) {
```
KostasTsiounis: You can save the result from `engineGetOutputSize()` so that you don't call it twice.

KostasTsiounis (changes requested):

PR #1037, opened 2025-12-11
Update junit version and use tree view runner

jasonkatonica:
As mentioned at the bottom of the maven surefire plugin a treeview is available to format the output of our Junit tests as they are executed.

This is setup to suppress any output that is not part of a test that has failed.

The treeview is able to allow easier view of tests that are actually run during a given execution along with any tests that have failed or are being skipped entirely. This will make it much easier to view what tests are really running to avoid accidental tests being skipped or not run at all.

Tests that are parameterized will print their parameters in the log.

Junit was updated to the latest Junit 5 version available.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1029

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -244,6 +244,13 @@
                   <groupId>org.apache.maven.plugins</groupId>
                   <artifactId>maven-surefire-plugin</artifactId>
                   <version>3.5.3</version>
+                  <dependencies>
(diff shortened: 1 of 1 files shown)


KostasTsiounis (approved):
LGTM

PR #1042, opened 2025-12-11
Add Check for PQCPrivateKey in Decapsulator

johnpeck-us-ibm:
Added code to check the private key passed in and
added tests to check the code for invalid keys

Signed-off-by: John Peck <140550562+johnpeck-us-ibm@users.noreply.github.com>

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
@@ -12,11 +12,16 @@ import com.ibm.crypto.plus.provider.ock.OCKException;
 import com.ibm.crypto.plus.provider.ock.OJPKEM;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
+import java.security.KeyFactory;
 import java.security.PrivateKey;
 import java.security.ProviderException;
 import java.security.PublicKey;
 import java.security.SecureRandom;
 import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.EncodedKeySpec;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.X509EncodedKeySpec;
+import java.util.Arrays;
 import javax.crypto.DecapsulateException;
 import javax.crypto.KEM;
 import javax.crypto.KEMSpi;
@@ -49,7 +54,6 @@ public class MLKEMImpl implements KEMSpi {
         return size;
     }
 
-
     /*
      * spec - The AlgorithmParameterSpec is not used and should be null. If not null
      * it will be ignored.
@@ -60,14 +64,28 @@ public class MLKEMImpl implements KEMSpi {
     public KEMSpi.EncapsulatorSpi engineNewEncapsulator(PublicKey publicKey,
             AlgorithmParameterSpec spec, SecureRandom secureRandom)
             throws InvalidAlgorithmParameterException, InvalidKeyException {
-        if (publicKey == null || !(publicKey instanceof PQCPublicKey) ) {
-            throw new InvalidKeyException("unsupported key");
+        
+        PublicKey pubKey = publicKey;
+        if (pubKey == null) {
+            throw new InvalidKeyException("Key is null.");
+        }
+
+        if (!(pubKey instanceof PQCPublicKey)) {
+            // Try and convert this key to a usage PQCPublicKey
+            try {
+                KeyFactory kf = KeyFactory.getInstance(this.alg, this.provider.getName());
+                EncodedKeySpec publicKeySpec = new X509EncodedKeySpec(publicKey.getEncoded());
+                pubKey = kf.generatePublic(publicKeySpec);
+       
+            } catch (Exception e) {
+                throw new InvalidKeyException("unsupported key", e);
+            }
         }
 
         if (spec != null) {
             throw new InvalidAlgorithmParameterException("no spec needed");
         }
-        return new MLKEMEncapsulator(publicKey, spec, null);
+        return new MLKEMEncapsulator(pubKey, spec, null);
     }
 
     class MLKEMEncapsulator implements KEMSpi.EncapsulatorSpi {
@@ -129,14 +147,32 @@ public class MLKEMImpl implements KEMSpi {
             AlgorithmParameterSpec spec)
             throws InvalidAlgorithmParameterException, InvalidKeyException {
  
-        if (privateKey == null) {
-            throw new InvalidKeyException("unsupported key");
+        PrivateKey privKey = privateKey;
+
+        if (privKey == null) {
+            throw new InvalidKeyException("Key is null.");
+        }
+
+        if (!(privKey instanceof PQCPrivateKey)) {
+            // Try and convert this key to a usage PQCPrivateKey
+            byte[] encoding = null;
+            try {
+                KeyFactory kf = KeyFactory.getInstance(this.alg, this.provider.getName());
+                encoding = privateKey.getEncoded();
+                PKCS8EncodedKeySpec privateKeySpec = new PKCS8EncodedKeySpec(encoding);
+                privKey = kf.generatePrivate(privateKeySpec);     
+            } catch (Exception e) {
+                throw new InvalidKeyException("unsupported key", e);
+            } finally {
+                Arrays.fill(encoding, (byte) 0);
+            }
+
         }
 
         if (spec != null) {
             throw new InvalidAlgorithmParameterException("no spec needed");
         }
-        return new MLKEMDecapsulator(privateKey, null);
+        return new MLKEMDecapsulator(privKey, null);
     }
 
     /*
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
@@ -8,6 +8,7 @@
 
 package ibm.jceplus.junit.base;
 
+import java.security.InvalidKeyException;
 import java.security.KeyFactory;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
@@ -20,6 +21,7 @@ import javax.crypto.SecretKey;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
 public class BaseTestKEM extends BaseTestJunit5 {
@@ -35,8 +37,6 @@ public class BaseTestKEM extends BaseTestJunit5 {
         KEM kem = KEM.getInstance(Algorithm, getProviderName());
 
         KeyPair pqcKeyPair = generateKeyPair(Algorithm);
-        pqcKeyPair.getPublic();
-        pqcKeyPair.getPrivate();
 
         KEM.Encapsulator encr = kem.newEncapsulator(pqcKeyPair.getPublic());
         KEM.Encapsulated enc = encr.encapsulate(0, 32, "AES");
@@ -56,8 +56,6 @@ public class BaseTestKEM extends BaseTestJunit5 {
         KEM kem = KEM.getInstance(Algorithm, getProviderName());
 
         KeyPair pqcKeyPair = generateKeyPair(Algorithm);
-        pqcKeyPair.getPublic();
-        pqcKeyPair.getPrivate();
 
         KEM.Encapsulator encr = kem.newEncapsulator(pqcKeyPair.getPublic());
         KEM.Encapsulated enc = encr.encapsulate();
@@ -78,8 +76,6 @@ public class BaseTestKEM extends BaseTestJunit5 {
         KEM kem = KEM.getInstance(Algorithm, getProviderName());
 
         KeyPair pqcKeyPair = generateKeyPair(Algorithm);
-        pqcKeyPair.getPublic();
-        pqcKeyPair.getPrivate();
 
         KEM.Encapsulator encr = kem.newEncapsulator(pqcKeyPair.getPublic());
         for (int i =0; i < 4; i++) {
@@ -161,8 +157,6 @@ public class BaseTestKEM extends BaseTestJunit5 {
         KEM kem = KEM.getInstance(Algorithm, getProviderName());
 
         KeyPair pqcKeyPair = generateKeyPair(Algorithm);
-        pqcKeyPair.getPublic();
-        pqcKeyPair.getPrivate();
 
         KEM.Encapsulator encr = kem.newEncapsulator(pqcKeyPair.getPublic());
         KEM.Encapsulated enc = encr.encapsulate(0, 16, "AES");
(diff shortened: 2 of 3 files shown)


src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
```diff
@@ -129,7 +129,7 @@ public KEMSpi.DecapsulatorSpi engineNewDecapsulator(PrivateKey privateKey,
             AlgorithmParameterSpec spec)
             throws InvalidAlgorithmParameterException, InvalidKeyException {
  
-        if (privateKey == null) {
+        if (privateKey == null || !(privateKey instanceof PQCPrivateKey)) {
```
jasonkatonica: I do understand the nature of this part of the update to ensure we have a PQCPrivateKey. Could we in the future, perhaps another update support private key material from OpenJDK here converting the key to our format?

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
```diff
@@ -175,6 +176,49 @@ public void testKEMSmallerSecret(String Algorithm) throws Exception {
...
+
+        KeyPair pqcKeyPair = generateKeyPair("RSA");
+        pqcKeyPair.getPublic();
+        pqcKeyPair.getPrivate();
+
+        try {
+            KEM.Encapsulator encr = kem.newEncapsulator(pqcKeyPair.getPublic());
+            fail("testKEMKeys failed RSA public key worked.");
+        } catch (InvalidKeyException ike) {
+            // continue with test
```
jasonkatonica: Asserting the error message here and other catch statements of "unsupported key" might be more specific.

jasonkatonica (commented):
Some tests are also failing with the following unused variables:
```
Error:  /home/runner/work/OpenJCEPlus/OpenJCEPlus/src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java:190:16: Unused local variable 'encr'. [UnusedLocalVariable]
Error:  /home/runner/work/OpenJCEPlus/OpenJCEPlus/src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java:197:16: Unused local variable 'decr'. [UnusedLocalVariable]
Error:  /home/runner/work/OpenJCEPlus/OpenJCEPlus/src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java:208:16: Unused local variable 'encr'. [UnusedLocalVariable]
Error:  /home/runner/work/OpenJCEPlus/OpenJCEPlus/src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java:215:16: Unused local variable 'decr'. [UnusedLocalVariable]
```

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
jasonkatonica: Yes if we have a means to do this automatically here then that would be ideal. I would assume that the conversion will also make a check very similar to this such that if it was not an appropriate key ( Such as RSA in the test below ) it would also fail given the key is not appropriate to be converted.

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
johnpeck-us-ibm: See updates to messages in code. They were changes to be specific to each case.

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
```diff
@@ -175,6 +168,60 @@ public void testKEMSmallerSecret(String Algorithm) throws Exception {
...
+
+        // Test null keys
+        PublicKey pub = null;
+        PrivateKey priv = null;
+
+        try {
+            KEM.Encapsulator encr = kem.newEncapsulator(pub);
+            if (removeWarning) {
+                System.out.println((Object) encr);
+            }
```
jasonkatonica: No need to print each `encr` here unless you have a reason?

To remove the unused variable we could replace:
```
EM.Encapsulator encr = kem.newEncapsulator(pub);
            if (removeWarning) {
                System.out.println((Object) encr);
            }
```
with:
```
kem.newEncapsulator(pub);
```
Similar question on the other spots we print the decapsulator.

johnpeck-us-ibm: I fixed this by using .toString() with out printing anything.

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
johnpeck-us-ibm: Added.

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
jasonkatonica: Yes this does fix the warning but we can remove all the code and just not declare `EM.Encapsulator encr` at all. In this case we wont get the warning and we can still make a call to the method under test with just `kem.newEncapsulator(pub);`

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
jasonkatonica: Great in the long run this prob would be better.

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
```diff
@@ -175,6 +169,51 @@ public void testKEMSmallerSecret(String Algorithm) throws Exception {
...
+    @CsvSource({"ML-KEM", "ML-KEM-512", "ML_KEM_768", "ML_KEM_1024"})
+    public void testKEMKeys(String Algorithm) throws Exception {
+
+        KEM kem = KEM.getInstance(Algorithm, getProviderName());
+
+        KeyPair pqcKeyPair = generateKeyPair("RSA");
+
+        try {
+            KEM.Encapsulator encr = kem.newEncapsulator(pqcKeyPair.getPublic());
+            encr.toString();
```
jasonkatonica: I dont think we need to execute `.toString()` on `encr`. We also dont need encr. I think we can change these lines:
```
KEM.Encapsulator encr = kem.newEncapsulator(pqcKeyPair.getPublic());
encr.toString();
```
to just:
```
kem.newEncapsulator(pqcKeyPair.getPublic());
```
The important part here is we call the `newEncapsulator` method to raise the exception we are driving.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
```diff
@@ -74,6 +74,38 @@ public void testPQCKeyGenKEM_PlusToInterop() throws Exception {
...
+        if (enc == null){
+            System.out.println("enc = null");
+            assertTrue(false, "KEMPlusCreatesInteropGet failed no enc.");
+        }
+        SecretKey keyE = enc.key();
+
+        KEM.Decapsulator decr = kemInterop.newDecapsulator(privateKey);
+        SecretKey keyD = decr.decapsulate(enc.encapsulation(), 0, 32, "AES");
+
+        assertTrue(Arrays.equals(keyE.getEncoded(), keyD.getEncoded()), "Secrets do NOT match");
```
jasonkatonica: Junit5 assert array can print what the arrays are in the case of failure. So instead of:
```
assertTrue(Arrays.equals(keyE.getEncoded(), keyD.getEncoded()), "Secrets do NOT match");
```
This might be more informative:
```
assertArrayEquals(keyE.getEncoded(), keyD.getEncoded(), "Secrets do NOT match");
```

jasonkatonica (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
johnpeck-us-ibm: I am doing this to get rid of a compile warning. I need to make use of encr and decr to get rid of the compiler warning. So, doing the toString makes the compiler happy. I probably code use KEM.Encapsulator_ instead of KEM.Encapsulator encr but that would only work on 25 and 26.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
johnpeck-us-ibm: Updated All

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
jasonkatonica: I switched to your branch and made [the above update ](https://github.com/IBM/OpenJCEPlus/pull/1042/files#r2615314875)in various places including these lines and did not see any sort of compilation warning. The checkstyle failure from earlier was also removed. Could you post more on the compiler error? Printing shoudl not be required from the compiler.

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
```diff
@@ -60,14 +64,28 @@ private int getEncapsulationLength() {
...
 
+        if (!(pubKey instanceof PQCPublicKey)) {
+            // Try and convert this key to a usage PQCPublicKey
+            try {
+                KeyFactory kf = KeyFactory.getInstance(this.alg, this.provider.getName());
+                EncodedKeySpec publicKeySpec = new X509EncodedKeySpec(publicKey.getEncoded());
+                pubKey = kf.generatePublic(publicKeySpec);
+       
+            } catch (Exception e) {
+                throw new InvalidKeyException("unsupported key", e.getCause());
```
jasonkatonica: We want e here not the cause of e.

jasonkatonica (changes requested):

jasonkatonica (changes requested):

jasonkatonica (approved):
LGTM

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
```diff
@@ -175,6 +169,47 @@ public void testKEMSmallerSecret(String Algorithm) throws Exception {
...
+        } catch (InvalidKeyException ike) {
+            assertTrue(ike.getMessage().equals("unsupported key"));
+        }
+  
+        try {
+            kem.newDecapsulator(priv);
+            fail("testKEMKeys failed - NULL Private key did not cause an Invalid Key Excepton.");
+        } catch (InvalidKeyException ike) {
+            assertTrue(ike.getMessage().equals("unsupported key"));
+        }
```
KostasTsiounis: Would it make sense to separate this into another test method?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
```diff
@@ -129,14 +147,27 @@ public KEMSpi.DecapsulatorSpi engineNewDecapsulator(PrivateKey privateKey,
             AlgorithmParameterSpec spec)
             throws InvalidAlgorithmParameterException, InvalidKeyException {
  
-        if (privateKey == null) {
+        PrivateKey privKey = privateKey;
+
+        if (privKey == null) {
             throw new InvalidKeyException("unsupported key");
```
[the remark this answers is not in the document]
johnpeck-us-ibm: Updated

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
```diff
@@ -60,14 +64,28 @@ private int getEncapsulationLength() {
     public KEMSpi.EncapsulatorSpi engineNewEncapsulator(PublicKey publicKey,
             AlgorithmParameterSpec spec, SecureRandom secureRandom)
             throws InvalidAlgorithmParameterException, InvalidKeyException {
-        if (publicKey == null || !(publicKey instanceof PQCPublicKey) ) {
+        
+        PublicKey pubKey = publicKey;
+        if (pubKey == null) {
             throw new InvalidKeyException("unsupported key");
```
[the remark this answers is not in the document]
johnpeck-us-ibm: Updated

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
johnpeck-us-ibm: It is doing testing of the keys and we really expect the same Exception. So, I don't thing it needs to be broken up.

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
```diff
@@ -129,14 +147,27 @@ public KEMSpi.DecapsulatorSpi engineNewDecapsulator(PrivateKey privateKey,
...
+
+        if (privKey == null) {
+            throw new InvalidKeyException("Key is null.");
+        }
+
+        if (!(privKey instanceof PQCPrivateKey)) {
+            // Try and convert this key to a usage PQCPrivateKey
+            try {
+                KeyFactory kf = KeyFactory.getInstance(this.alg, this.provider.getName());      
+                PKCS8EncodedKeySpec privateKeySpec = new PKCS8EncodedKeySpec(privateKey.getEncoded());
```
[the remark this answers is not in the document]
johnpeck-us-ibm: My understanding of clone is that the result of clone is another separate object with the same value as the original. This is done so that changes can be done to the cloned instance with out affecting the original. So, there is no way to clear the encoding from PKCS8EncodedKeySpec.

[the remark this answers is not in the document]
KostasTsiounis: The call to `privateKey.getEncoded()` creates an encoding of the private key that is saved in memory. The constructor of `PKCS8EncodedKeySpec` though does a clone on that (doesn't use the original) and the original byte array is dereferenced and not used any more but the encoding is not zeroed out. The only way to do that is with something similar to what I suggested above.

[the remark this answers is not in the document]
johnpeck-us-ibm: I get what you are saying. I will fix it like you suggested.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
```diff
@@ -74,6 +75,38 @@ public void testPQCKeyGenKEM_PlusToInterop() throws Exception {
         assertTrue(same);
     } 
 
+    @Test
+    public void testPQCKeyGenKEMAutoKeyConvertion() throws Exception {
+        String pqcAlgorithm = "ML-KEM-512";
+
+        if (getProviderName().equals("OpenJCEPlusFIPS") ||
```
KostasTsiounis: We could substitute this with:
```
assumeFalse("OpenJCEPlusFIPS".equals(getProviderName()));
assumeFalse(Utils.PROVIDER_BC.equals(getProviderName()));
```

KostasTsiounis (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
```diff
@@ -175,6 +169,47 @@ public void testKEMSmallerSecret(String Algorithm) throws Exception {
...
+        try {
+            kem.newEncapsulator(pub);
+            fail("testKEMKeys failed - NULL Public key did not cause an Invalid Key Excepton.");
+        } catch (InvalidKeyException ike) {
+            assertTrue(ike.getMessage().equals("Key is null."));
+        }
+  
+        try {
+            kem.newDecapsulator(priv);
+            fail("testKEMKeys failed - NULL Private key did not cause an Invalid Key Excepton.");
```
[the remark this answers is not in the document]
johnpeck-us-ibm: Fixed

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
johnpeck-us-ibm: Updated

KostasTsiounis (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
```diff
@@ -175,6 +169,47 @@ public void testKEMSmallerSecret(String Algorithm) throws Exception {
...
+        try {
+            kem.newEncapsulator(pqcKeyPair.getPublic());
+            fail("testKEMKeys failed - RSA Public key did not cause an InvalidKeyExcepton.");
+        } catch (InvalidKeyException ike) {
+            assertTrue(ike.getMessage().equals("unsupported key"));
+        }
+  
+        try {
+            kem.newDecapsulator(pqcKeyPair.getPrivate());
+            fail("testKEMKeys failed - RSA Private key did not cause an InvalidKeyExcepton.");
```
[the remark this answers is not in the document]
johnpeck-us-ibm: Hope I got it right this time :(

src/test/java/ibm/jceplus/junit/base/BaseTestKEM.java
```diff
@@ -175,6 +169,47 @@ public void testKEMSmallerSecret(String Algorithm) throws Exception {
...
+            assertTrue(ike.getMessage().equals("unsupported key"));
+        }
+
+        // Test null keys
+        PublicKey pub = null;
+        PrivateKey priv = null;
+
+        try {
+            kem.newEncapsulator(pub);
+            fail("testKEMKeys failed - NULL Public key did not cause an InvalidKeyExcepton.");
```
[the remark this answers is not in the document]
johnpeck-us-ibm: Hope I got it right this time :(

KostasTsiounis (approved):
LGTM

PR #1044, opened 2025-12-11
Remove fallback option for deprecated OAEP parameter functionality

KostasTsiounis:
After changes to the OAEPParameters class to follow newer expectations, code and a flag to utilize it was left behind for compatibility reasons.

Nobody run into issues or requested to exercise this path, so we can safely remove it.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1012


jasonkatonica (approved):
LGTM

PR #1091, opened 2026-01-13
Update parsing/checking private key encoding and equals()

### LINKED ISSUE #1053: Fix EC Private Key equality mismatch after reconstruction from ECPrivateKeySpec
### Description

There is a discrepancy in the byte array returned by `getEncoded()` when an Elliptic Curve (EC) private key is reconstructed using `KeyFactory`. While the mathematical private scalar () remains identical, the DER-encoded PKCS#8 structure changes in length and content.

That is causing the **JTREG** `java/security/KeyAgreement/KeySpecTest.java` failure.

### Steps to Reproduce

1. Generate an EC Private Key using `KeyPairGenerator`.
2. Extract the `ECPrivateKeySpec` from the generated key.
3. Reconstruct a new `PrivateKey` object using `KeyFactory.generatePrivate(spec)`.
4. Compare `original.getEncoded()` and `reconstructed.getEncoded()`.

### Observed Behavior

`Arrays.equals()` returns `false`.

* **Original Key:** Often a "minimal" PKCS#8 structure (approx. 67 bytes).
* **Reconstructed Key:** Often a "full" PKCS#8 structure (approx. 150 bytes) containing optional attributes like the Public Key bits.

### Expected Behavior

While the binary encodings differ due to ASN.1 optional fields, the keys should be recognized as logically/cryptographically equivalent.

### Technical Analysis

The `OpenJCEPlus` (or SunEC) provider expands the encoded format during reconstruction.

* **Key 1 (Bare):** `VERSION + ALGO_ID + OCTET_STRING(S)`
* **Key 2 (Full):** `VERSION + ALGO_ID + OCTET_STRING(S) + [0]PARAMS + [1]PUBLIC_KEY`

### The test to reproduce:
```java
import java.security.*;
import java.security.interfaces.ECPrivateKey;
import java.security.spec.ECParameterSpec;
import java.security.spec.ECPrivateKeySpec;
import java.security.spec.KeySpec;
import java.util.HexFormat;

public class ECDiffTest {

        // Use your specific provider
        String provider = "OpenJCEPlus";

        // 1. Generate an initial EC Private Key (e.g., NIST P-256)
        KeyPairGenerator kpg = KeyPairGenerator.getInstance("EC", provider);
        kpg.initialize(256);
        PrivateKey originalKey = kpg.generateKeyPair().getPrivate();

        byte[] originalEncoded = originalKey.getEncoded();

        // 2. Build a new key using the KeySpec of the first one
        // This process often triggers the Provider to add optional metadata (like the Public Key)
        KeyFactory kf = KeyFactory.getInstance("EC", provider);
        KeySpec priSpec = kf.getKeySpec(originalKey, ECPrivateKeySpec.class);
        PrivateKey reconstructedKey = kf.generatePrivate(priSpec);

        byte[] reconstructedEncoded = reconstructedKey.getEncoded();

        // 3. Comparison Output
        System.out.println("=== Comparison Results ===");
        System.out.println("Original Length:      " + originalEncoded.length + " bytes");
        System.out.println("Reconstructed Length: " + reconstructedEncoded.length + " bytes");

        // Binary check
        boolean isBinaryEqual = java.util.Arrays.equals(originalEncoded, reconstructedEncoded);
        System.out.println("Are encoded bytes identical? " + (isBinaryEqual ? "YES" : "NO"));

        // Mathematical check
        boolean isMathematicallyEqual = areKeysEqual(originalKey, reconstructedKey);
        System.out.println("Are the keys mathematically identical? " + (isMathematicallyEqual ? "YES" : "NO"));

        if (!isBinaryEqual) {
            System.out.println("\n--- Hex Comparison ---");
            System.out.println("Original Hex:\n" + HexFormat.of().formatHex(originalEncoded));
            System.out.println("\nReconstructed Hex:\n" + HexFormat.of().formatHex(reconstructedEncoded));
            System.out.println("\nNote: The second key is longer because it likely includes the Public Key bits and Curve OID metadata.");
        }
    }

    /**
     * Compares two PrivateKeys by looking at the Private Scalar (S)
     * and the Curve Parameters rather than the encoded bytes.
     */
    public static boolean areKeysEqual(PrivateKey k1, PrivateKey k2) {
        if (k1 == k2) return true;
        if (!(k1 instanceof ECPrivateKey) || !(k2 instanceof ECPrivateKey)) return false;

        ECPrivateKey ec1 = (ECPrivateKey) k1;
        ECPrivateKey ec2 = (ECPrivateKey) k2;

        // 1. Compare the secret scalar (S)
        if (!ec1.getS().equals(ec2.getS())) {
            return false;
        }

        // 2. Compare the Curve Parameters
        ECParameterSpec s1 = ec1.getParams();
        ECParameterSpec s2 = ec2.getParams();

        return s1.getCurve().equals(s2.getCurve()) &&
                s1.getGenerator().equals(s2.getGenerator()) &&
                s1.getOrder().equals(s2.getOrder()) &&
                s1.getCofactor() == s2.getCofactor();
    }
}
```

### The failing JTREG
`java/security/KeyAgreement/KeySpecTest.java`

### The root cause PR
https://github.com/IBM/OpenJCEPlus/pull/682

### Suggested fix

Revert the EC private key `equals` method to check `getS()` and `params` (the same as the above test).

FYI @jasonkatonica @KostasTsiounis

KostasTsiounis:
The method that parses the private key encoding is updated to check the encoding's validity.

An `equals()` method is added that calls the `super.equals()` and, if that fails, directly compares the keys and params.

Several tests are, also, added to test key import, direct and interop, through `ECPrivateKeySpec`.

Fixes #1053 

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
@@ -103,17 +103,10 @@ final class ECPrivateKey extends PKCS8Key implements java.security.interfaces.EC
         this.provider = provider;
 
         try {
-            // Set parameters.
-            AlgorithmParameters algParams = this.algid.getParameters();
-            if (algParams == null) {
-                throw new IOException(
-                        "EC domain parameters must be encoded in the algorithm identifier");
-            }
-            this.params = algParams.getParameterSpec(ECParameterSpec.class);
-
             // Get from the encoding:
             //    * the private key as a BigInteger (this.s)
             //    * the public key, if available (this.pubKeyEncoded)
+            // and set parameters.
             parsePrivateKeyEncoding();
 
             // Create appropriate encoding and create ecKey.
@@ -140,20 +133,13 @@ final class ECPrivateKey extends PKCS8Key implements java.security.interfaces.EC
             algidOut.putDerValue(new DerValue(ecKey.getParameters()));
             this.algid = AlgorithmId
                     .parse(new DerValue(DerValue.tag_Sequence, algidOut.toByteArray()));
-
-            AlgorithmParameters algParams = this.algid.getParameters();
-            if (algParams == null) {
-                throw new IOException(
-                        "EC domain parameters must be encoded in the algorithm identifier");
-            }
-            this.params = algParams.getParameterSpec(ECParameterSpec.class);
-
             // Get private key encoding from ECKey.
             this.privKeyMaterial = ecKey.getPrivateKeyBytes();
 
             // Get from the encoding:
             //    * the private key as a BigInteger (this.s)
             //    * the public key, if available (this.pubKeyEncoded)
+            // and set parameters.
             parsePrivateKeyEncoding();
         } catch (Exception exception) {
             throw new InvalidKeyException("Failed to create EC private key", exception);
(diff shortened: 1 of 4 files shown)


src/test/java/ibm/jceplus/junit/base/BaseTestECKeyImportInterop.java
```diff
@@ -114,6 +116,61 @@ public void doCreateKeyPairECGenParamImport(String generateProviderName,
...
+        byte[] privKeyBytes = privateKey.getEncoded();
+
+        // System.out.println (methodName + " pubKeyBytes length=" +
+        // publicKeyBytes.length);
+        // System.out.println (methodName + " publicKeyBytes = " +
+        // BaseUtils.bytesToHex(publicKeyBytes));
+        // System.out.println (methodName + " privKeyBytes length=" +
+        // privKeyBytes.length);
+        // System.out.println (methodName + " privKeyBytes = " +
+        // BaseUtils.bytesToHex(privKeyBytes));
```
jasonkatonica: Any reason to leave these comments?

src/main/java/com/ibm/crypto/plus/provider/ECPrivateKey.java
```diff
@@ -219,25 +185,69 @@ private byte[] createEncodedPrivateKeyWithParams() throws IOException {
...
+                    return;
+                }
+                value = data.getDerValue();
+            }
 
-        for (int i = 2; i < inputDerValue.length; i++) {
-            DerValue v = inputDerValue[i];
-            if (v.isContextSpecific(TAG_PUBLIC_KEY_ATTRS)) {
-                DerValue bits = v.withTag(DerValue.tag_BitString);
+            if (value.isContextSpecific(TAG_PUBLIC_KEY_ATTRS)) {
```
jasonkatonica: My understanding is that public key and parameters are both optional. Is there a reason we dont get an exception here on line 242 when no public key is specified. The test coverage is also showing that the else statement is never executed and we always have a public key in our testing? This is unexpected to me since i expected the hardcoded values in the `BaseTestECKeyImport.java` to have variations with and without the optional public key.

src/test/java/ibm/jceplus/junit/base/BaseTestECKeyImportInterop.java
```diff
@@ -114,6 +116,61 @@ public void doCreateKeyPairECGenParamImport(String generateProviderName,
...
+        same = privateKey.equals(privateKey3);
+        assertTrue(same);
+        same = publicKey.equals(publicKey2);
+        assertTrue(same);
+
+        byte[] publicKey2Bytes = publicKey2.getEncoded();
+        byte[] privateKey2Bytes = privateKey2.getEncoded();
+
+        assertTrue(Arrays.equals(publicKey2Bytes, publicKeyBytes));
+        assertTrue(Arrays.equals(privateKey2Bytes, privKeyBytes));
```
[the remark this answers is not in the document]
KostasTsiounis: Changed.

PR #1092, opened 2026-01-14
Refactor JUnit 5 tests consolidate BaseTestAESGCM

Mohit-Rajbhar100698:
This change consolidates various tests associated with the former BaseTestAESGCM into a single parameterized test.


src/test/java/ibm/jceplus/junit/tests/TestArguments.java
```diff
@@ -19,14 +19,35 @@
...
+        int[] aesKeySizes = {128, 192, 256};
+        return keySizesAndJCEPlusProviders(aesKeySizes);
+    }
+
+    /**     
+     * Generates test combinations for each AESGCM key size with the active OpenJCEPlus* providers.
+     * 
+     * @return Stream of Arguments containing AESGCM key sizes and OpenJCEPlus* providers
+     */
+    public static Stream<Arguments> aesgcmKeySizesAndJCEPlusProviders() {
```
jasonkatonica: For the same inputs for symmetric keys i dont think there is an advantage of having both the methods `aesKeySizesAndJCEPlusProviders` and also `aesgcmKeySizesAndJCEPlusProviders`. I think we can eliminate `aesgcmKeySizesAndJCEPlusProviders` and just use `aesKeySizesAndJCEPlusProviders` for both the aes and aesgcm tests. I do like the idea of creating a helper method that takes key sizes and combines it with the providers liek you have done here. This will be handy in the future for RSA and other keys that are of different sizes.

src/test/java/ibm/jceplus/junit/TestMultithread.java
```diff
@@ -31,9 +31,6 @@ public class TestMultithread {
     private final int numThreads = 10;
     private final int timeoutSec = 4500;
     private final String[] testList = {
-            "ibm.jceplus.junit.openjceplus.multithread.TestAESGCM_128",
-            "ibm.jceplus.junit.openjceplus.multithread.TestAESGCM_192",
-            "ibm.jceplus.junit.openjceplus.multithread.TestAESGCM_256",
```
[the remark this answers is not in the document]
Mohit-Rajbhar100698: just to confirm my understanding:

we should create two more tags "OpenJCEPlusMultithread" and "OpenJCEPlusFIPSMultithread" then use these tags in all those test that we want to execute in  multithreaded as well and we will use these new tags in above suites.

Finally the test which runs multithreaded will have four tags in total 
@Tag(TestProvider.OPENJCEPLUS_NAME)
@Tag(TestProvider.OPENJCEPLUS_FIPS_NAME)
@Tag(TestProvider.OPENJCEPLUS_NAME_MULTITHREADED)
@Tag(TestProvider.OPENJCEPLUS_FIPS_MULTITHREADED)

And the test which do not run multithreaded will have only two tags.
@Tag(TestProvider.OPENJCEPLUS_NAME)
@Tag(TestProvider.OPENJCEPLUS_FIPS_NAME)

Is it ?

src/test/java/ibm/jceplus/junit/tests/TestProvider.java
```diff
@@ -20,14 +20,12 @@ public enum TestProvider {
...
     OpenJCEPlusFIPS("OpenJCEPlusFIPS", "com.ibm.crypto.plus.provider.OpenJCEPlusFIPS");
 
-    // Constants for provider names (can be used in annotations where compiler couldt use a runtime method)
+    // Constants for provider names (can be used in annotations where compiler couldn't use a runtime method)
     public static final String BC_NAME = "BC";
     public static final String SUN_NAME = "SUN";
     public static final String SUNJCE_NAME = "SunJCE";
     public static final String SUNRSASIGN_NAME = "SunRsaSign";
     public static final String SUNEC_NAME = "SunEC";
-    public static final String OPENJCEPLUS_NAME = "OpenJCEPlus";
```
jasonkatonica: I still think we want to keep these two provider names here so they can be used for provider names through the code.

src/test/java/ibm/jceplus/junit/tests/Tags.java
```diff
@@ -0,0 +1,41 @@
...
+ * Enumeration of test tags used to categorize tests.
+ */
+public enum Tags {
+
+    OPENJCEPLUS("OpenJCEPlus"),
+    OPENJCEPLUS_FIPS("OpenJCEPlusFIPS"),
+    OPENJCEPLUS_MULTITHREAD("OpenJCEPlusMultithread"),
+    OPENJCEPLUS_FIPS_MULTITHREAD("OpenJCEPlusFIPSMultithread");
+
+    // Constants for tag names (can be used in annotations where compiler couldn't use a runtime method)
```
jasonkatonica: We can use the value from the `TestProvider` enum in this spot such that our string literals will always match the provider name.

src/test/java/ibm/jceplus/junit/tests/TestProvider.java
Mohit-Rajbhar100698: Reverted changes here.

PR #1108, opened 2026-01-15
Update parsing/checking private key encoding and equals()

KostasTsiounis:
The method that parses the private key encoding is updated to check the encoding's validity.

An `equals()` method is added that calls the `super.equals()` and, if that fails, directly compares the keys and params.

Several tests are, also, added to test key import, direct and interop, through `ECPrivateKeySpec`.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1091


jasonkatonica (approved):
LGTM

PR #1112, opened 2026-01-16
Fix PQC Memory leak issues.

### LINKED ISSUE #1088: Fix memory leaks in PQC native files
**1.**
This line in KEM.c is causing a memory leak
https://github.com/IBM/OpenJCEPlus/blob/87e0b5c79e83867c26965868fc5006770de345fc/src/main/native/KEM.c#L70

That should be
```java
if (genkeylocal != NULL) {
    free(genkeylocal );
}
```

**2.**
This line in MLKey.c
https://github.com/IBM/OpenJCEPlus/blob/87e0b5c79e83867c26965868fc5006770de345fc/src/main/native/MLKey.c#L656

If an error occurs during `ICC_i2d_PublicKey`, the code attempts to clean up by calling `DeleteLocalRef(env, keyBytes)`. However, it then returns that deleted reference to Java.
The Java side receives an invalid/deleted pointer, leading to an immediate JVM crash or undefined behavior. If the function succeeds, it works, but the cleanup logic is effectively bypassed in the failure path.

It should be  `return retKeyBytes;`.

**3.**

https://github.com/IBM/OpenJCEPlus/blob/87e0b5c79e83867c26965868fc5006770de345fc/src/main/native/MLKey.c#L394

**Improper Use of Critical Regions:** When using functions like `GetPrimitiveArrayCritical` and `ReleasePrimitiveArrayCritical`, the native code is in a "critical region". Within this region, native code must not call other JNI functions or any system calls that might cause the current thread to block and wait for another Java thread. Violating this can interfere with the JVM's garbage collection mechanisms.

In many JVMs, `GetPrimitiveArrayCritical` disables garbage collection. Calling another JNI function that triggers a GC while it is disabled can lead to a deadlock or system-wide memory pressure that looks like a leak.

It should call `GetStringUTFChars` before entering the Critical block.

**4.**

This line in SignaturePQC.c
https://github.com/IBM/OpenJCEPlus/blob/87e0b5c79e83867c26965868fc5006770de345fc/src/main/native/SignaturePQC.c#L180

It should be `sigBytesNative`:
```java
if (dataNative == NULL) {
    (*env)->ReleasePrimitiveArrayCritical(env, sigBytes, sigBytesNative, JNI_ABORT);
    throwICCException(env, 0, "GetPrimitiveArrayCritical failed");
    return verified;
}
```

@jasonkatonica @johnpeck-us-ibm

johnpeck-us-ibm:
Make fixes found by zOS. That could or will cause
a memory leak.

Fixes: [https://github.com/IBM/OpenJCEPlus/issues/1088](https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_IBM_OpenJCEPlus_issues_1088&d=DwQCaQ&c=BSDicqBQBDjDI9RkVyTcHQ&r=Me_Bp8tCcDmNNqv1ATjFs3RXEaTd4kAsjiSPsbsUPt0&m=xcYh9oIVF0UVd3wpo1rmX1CFIxkAy5g6LxnZqgvq8k2p_SS1ANTrKX-2cVerXARI&s=b5UM6KVo7fgc4JdVUdkWhJQNeV_YAqUAizGhpfyLCck&e=)

### DIFF
--- a/src/main/native/KEM.c
+++ b/src/main/native/KEM.c
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2025
+ * Copyright IBM Corp. 2025, 2026
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms provided by IBM in the LICENSE file that accompanied
@@ -67,7 +67,7 @@ Java_com_ibm_crypto_plus_provider_ock_NativeInterface_KEM_1encapsulate(
             free(wrappedKeyLocal);
         }
         if (genkeylocal != NULL) {
-            free(wrappedKeyLocal);
+            free(genkeylocal);
         }
         ICC_EVP_PKEY_CTX_free(ockCtx, evp_pk);
         throwOCKException(env, 0, "malloc failed");
--- a/src/main/native/MLKey.c
+++ b/src/main/native/MLKey.c
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corp. 2025
+ * Copyright IBM Corp. 2025, 2026
  *
  * This code is free software; you can redistribute it and/or modify it
  * under the terms provided by IBM in the LICENSE file that accompanied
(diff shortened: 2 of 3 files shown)


KostasTsiounis (changes requested):

jasonkatonica (commented):
Code looks good to me only one nit.

KostasTsiounis (approved):
LGTM

jasonkatonica (approved):
LGTM

PR #1121, opened 2026-01-21
Restrict RSA paddings allowed through OpenJCEPlusFIPS

KostasTsiounis:
This change adds attributes in the RSA service registration in the `OpenJCEPlusFIPS` provider to only support OAEP paddings.

It, also, update the `engineSetPadding()` method to only allow OAEP paddings to be set when the FIPS provider is used.

Tests are updated accordingly to skip invalid padding tests when running with `OpenJCEPlusFIPS`.

A temporary flag that allows the use of other paddings with `OpenJCEPlusFIPS` is introduced to facilitate migration of users utilizing the previous behaviour. The flag to be set to revert this behaviour is `-Dcom.ibm.openjceplusfips.allowNonOAEP=true`.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1036


taoliult (approved):
LGTM

PR #1123, opened 2026-01-21
Change package name and move native files

KostasTsiounis:
The package for the supporting classes that invoke the native library is changed from `com.ibm.crypto.plus.provider.ock` to `com.ibm.crypto.plus.provider.base`, since the same classes can be used to invoke functionality from other libraries in the future.

The native files pertaining to `OCKC` functionality are moved to a newly created subdirectory (i.e., they are now under `native/ock/` instead of just `native/`).

The makefiles and `pom.xml` are updated to account for these changes.


KostasTsiounis:
> I didn’t check all the copyright notices, loading all of them in the browser makes it respond quite slowly... Please update them if needed.

I updated the copyrights in all files involved.

KostasTsiounis:
> Second, I like to see everything collected together that is related. This is more a preference than anything. It just makes finding related things easier. Especially for someone who is not familiar with the code. But looking forward unless we have a bunch of back ends it really does not matter.

I'm not sure if you mean changing the whole directory structure to put the java and native code for OCK for example under the same directory. But Maven prefers that Java classes are under `src/main/java` and native code is under `src/main/native`. Changing that would break or at least make the mvn process error-prone.

jasonkatonica:
I think it would be best if we stayed as close as possible to the standard Maven directory layout. 

The convention is documented here ( even thought they dont explictly state where native code goes ) although the most common one googling is `src/main/cpp` or `src/main/native`:


https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html


Following that structure, we could eventually align things like this:

```
src/main/native/shared
src/main/native/ockc
src/main/native/openssl
src/main/java/.../shared
src/main/java/.../ockc
src/main/java/.../openssl
```

The main objective is to keep native code isolated in the native hierarchy and Java code isolated in the java hierarchy, while still keeping backend-specific code grouped logically.
At one point, Kostas and I discussed creating a src/main/java/.../shared package now and splitting out the Java pieces accordingly. However, doing that now would start to blur the line between this PR (which is intended to avoid functional changes) and the larger refactoring work we want to approach later. My preference is to keep this PR focused on structural cleanup and defer the package reorganization to a follow‑up PR.

PR #1141, opened 2026-01-30
Fail OpenJCEPlus compilation on warnings

sabrinajlee:
Compilation will now fail if any warnings appear. Warnings that were previously appearing have been resolved so this option will make it easier to catch warnings as they appear, especially when backporting changes. This will not affect building OpenJDK.


KostasTsiounis (approved):
LGTM

PR #1076, opened 2026-01-07
Cleanup Java versions and update JaCoCo

KostasTsiounis:
Remove unnecessary Java versions from Jenkins scripts and mvn profiles.

Update JaCoCo to the latest version.

Signed-off-by: Kostas Tsiounis <kostas.tsiounis@ibm.com>

PR #1165, opened 2026-02-05
Disallow weak RSA keys on macOS and Linux (aarch64) in FIPS developer mode

Mohit-Rajbhar100698:
This change prevents weak RSA key sizes (512 and 1024 bits) in OpenJCEPlusFIPS on macOS and Linux (aarch64) where FIPS behavior is simulated using developer mode.

Fixes: [#1076](https://github.ibm.com/runtimes/jit-crypto/issues/1076)

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
@@ -689,6 +689,32 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
         return new OpenJCEPlusFIPSContext();
     }
 
+    /**
+     * Check if the current platform is Mac or Linux on aarch64 architecture,
+     */
+    @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -689,6 +689,27 @@ ProviderContext getProviderContext() {
...
+            boolean isMac = osName.toLowerCase().contains("mac");
+            boolean isLinuxAarch64 = osName.toLowerCase().contains("linux")
+                            && osArch.toLowerCase().contains("aarch64");
+
+            if (isMac || isLinuxAarch64) {
+                return true;
+            }
+        }
+
+        return super.isFIPS();
```
KostasTsiounis: I would suggest calling `super.isFIPS()` first and only do the rest if it's false. That way we avoid going through that logic in most platforms.

PR #1167, opened 2026-02-10
Add Ability to use Config File

johnpeck-us-ibm:
Added the ability to use a config file to create a a new OpenJCEPlus provider based on that file

Signed-off-by: John Peck [140550562+johnpeck-us-ibm@users.noreply.github.com](mailto:140550562+johnpeck-us-ibm@users.noreply.github.com)


src/main/java/com/ibm/crypto/plus/provider/DefaultProviderAttrs.java
```diff
@@ -0,0 +1,476 @@
...
+                + "KeyGenerator.kda-hkdf-with-sha1.alias.add = kda-hkdf-with-sha-1\n"
+                + "Service.KeyGenerator.kda-hkdf-with-sha1 = com.ibm.crypto.plus.provider.HKDFGenerator$HKDFwithSHA1\n"
+
+                + "KeyGenerator.kda-hkdf-with-sha224.alias.add = kda-hkdf-with-sha-224\n"
+                + "Service.KeyGenerator.kda-hkdf-with-sha224 = com.ibm.crypto.plus.provider.HKDFGenerator$HKDFwithSHA224\n"
+
+                + "KeyGenerator.kda-hkdf-with-sha256.alias.add = kda-hkdf-with-sha-256\n"
+                + "Service.KeyGenerator.kda-hkdf-with-sha256 = com.ibm.crypto.plus.provider.HKDFGenerator$HKDFwithSHA256\n"
+            
+                + "KeyGenerator.kda-hkdf-with-sha384.alias.add = kkda-hkdf-with-sha-384\n"
```
jasonkatonica: This should be `kkda-hkdf-with-sha-384` instead of `kda-hkdf-with-sha-384` like the others.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
```diff
@@ -93,8 +95,18 @@ public OpenJCEPlus() {
...
+            ProviderServiceReader config = new ProviderServiceReader(new BufferedReader(new StringReader(DefaultProviderAttrs.getConfigString())));  
+            List<ProviderServiceReader.ServiceDefinition> services = config.readServices();
+            for (ProviderServiceReader.ServiceDefinition service : services) {
+                putService(new OpenJCEPlusService(jce, service.getType(), service.getAlgorithm(),
+                    service.getClassName(), service.getAliases().toArray(new String[service.getAliases().size()]), service.getAttributes()));
+            }
+        } catch (IOException e) {
+            throw new InvalidParameterException("Error configuring OpenJCEPlus provider - " + e.getMessage()); 
+        }    
+
```
jasonkatonica: This code ( lines 99 - 108 ) is very similar if not exactly the same as the OpenJCEPlusFIPS class. Seems we could put this into a private helper method in the OpenJCEPlusProvider class that can be used in both providers.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
```diff
@@ -93,8 +95,18 @@ public OpenJCEPlus() {
...
+ 
+        try {
+            ProviderServiceReader config = new ProviderServiceReader(new BufferedReader(new StringReader(DefaultProviderAttrs.getConfigString())));  
+            List<ProviderServiceReader.ServiceDefinition> services = config.readServices();
+            for (ProviderServiceReader.ServiceDefinition service : services) {
+                putService(new OpenJCEPlusService(jce, service.getType(), service.getAlgorithm(),
+                    service.getClassName(), service.getAliases().toArray(new String[service.getAliases().size()]), service.getAttributes()));
+            }
+        } catch (IOException e) {
+            throw new InvalidParameterException("Error configuring OpenJCEPlus provider - " + e.getMessage());
```
jasonkatonica: When throwing the exception on this line can we throw the entire `e` in the caused by for the exception instead of just the message.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -140,583 +153,6 @@ public OpenJCEPlusFIPS() {
         }
     }
```
jasonkatonica, line 130: OpenJCEPlus has both a configure(String) and configure(BufferedReader) method. I dont see these added here to the FIPS provider. For consistancy and for allowing support / extension / changes of the OpenJCEPlus-xxx providers in the future should these be added here also?

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -0,0 +1,430 @@
...
+ * This class reads files containing putService() calls in the format:
+ * putService(new OpenJCEPlusService(provider, "Type", "Algorithm", "ClassName", aliases));
+ * 
+ * Example usage:
+ * <pre>
+ * ProviderServiceReader reader = new ProviderServiceReader("services.txt");
+ * List<ServiceDefinition> services = reader.readServices();
+ * for (ServiceDefinition service : services) {
+ *     System.out.println(service.getType() + ": " + service.getAlgorithm());
+ * }
```
jasonkatonica: We are using properties now from the file so this javadoc should reflect that change.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -0,0 +1,430 @@
...
+        this.filePath = filePath;
+    }
+   
+    /**
+     * Creates a new ProviderServiceReader for the specified file.
+     * 
+     * @param filePath the path to the file containing service definitions
+     */
+    public ProviderServiceReader(BufferedReader br) {
+        this.reader = br;
```
jasonkatonica, line 120: Can we check for null here?

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -0,0 +1,430 @@
...
+        }
+    }
+    
+    /**
+     * Creates a new ProviderServiceReader for the specified file.
+     * 
+     * @param filePath the path to the file containing service definitions
+     */
+    public ProviderServiceReader(String filePath) {
+        this.filePath = filePath;
```
jasonkatonica, line 111: Can we check for null here?

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -0,0 +1,430 @@
...
+        try {
+            if (filePath == null && this.reader == null) {
+                throw new IOException("No file specified");
+            } else if (null == filePath && this.reader != null) {
+                rd = this.reader;
+            } else if (filePath != null && !Files.exists(Paths.get(filePath))) {
+                throw new IOException("File not found: " + filePath);
+            } else {
+                // this filePath != null && Files.exists(Paths.get(filePath))
+                rd = new BufferedReader(new FileReader(filePath));
```
jasonkatonica, line 146: We should ensure that `rd` is always closed.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -0,0 +1,430 @@
...
+                Map<String, String> attributes = processAttributes(parts, setAttributes, pr, null);
+                ServiceDefinition service = new ServiceDefinition(parts[1], parts[2], pr.getProperty(key), aliases, attributes);
+                if (service != null) {
+                    services.add(service);
+                    aliases = null;
+                    attributes.clear();
+                }
+            }
+        } catch (Exception e) {
+            throw new IOException("File issue: " + e.getMessage());
```
jasonkatonica, line 220: Please throw `e` in the caused by of the exception so its not hidden if an error occurs.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -0,0 +1,430 @@
...
+        //.delete will remove the aliasses from the current list of aliases
+        //.replace will remove the current list of aliases and then add the new ones.
+
+        //add the default aliases if applicable.
+        if (defaultPr != null) {
+            String value = defaultPr.getProperty(keyBase + ".add");
+            if (value != null) {
+                String[] aliases = value.split("\\s*,\\s*");
+                for (String alias : aliases) {
+                    Aliases.add(alias);
```
jasonkatonica: Should we ensure that the alias is not null or empty string due to the regex above?

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -0,0 +1,430 @@
...
+     * per Type and Algorithm.
+     *
+     * @param parts the service key parts (Service, Type, Algorithm)
+     * @param defaultPr the default properties (can be null)
+     * @param configPr the config file properties (can be null)
+     * @param configAliases the set of alias keys from config file (can be null)
+     * @return a list of alias strings
+     */
+    private List<String> processAliases(String[] parts, Properties defaultPr, Properties configPr, Set<String> configAliases) {
+        List<String> Aliases = new ArrayList<>();
```
jasonkatonica: Perhaps this should be lower case since its a local variable.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
```diff
@@ -111,1123 +123,45 @@ public OpenJCEPlus() {
...
-                    "com.ibm.crypto.plus.provider.HKDFGenerator$HKDFwithSHA1", aliases));
+    public Provider configure(BufferedReader br) throws InvalidParameterException {
+        return new OpenJCEPlus(new ProviderServiceReader(br));
+    }
 
-            aliases = new String[] {"kda-hkdf-with-sha-224"};
-            putService(new OpenJCEPlusService(jce, "KeyGenerator", "kda-hkdf-with-sha224",
-                    "com.ibm.crypto.plus.provider.HKDFGenerator$HKDFwithSHA224", aliases));
+    public OpenJCEPlus(ProviderServiceReader config) {
+        super("OpenJCEPlus-" + config.getName(), config.getDesc());
```
jasonkatonica, line 149: Should we check that the name and description is not null or empty?

Also we should add the name of the provider we use in the debug on line 138.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -0,0 +1,430 @@
...
+                    name = pr.getProperty(key);
+                } else if (parts.length == 1 && parts[0].equalsIgnoreCase("description")) {
+                    description = pr.getProperty(key);
+                } else if (parts.length == 1 && parts[0].equalsIgnoreCase("default")) {
+                    defaults = pr.getProperty(key);
+                } else {
+                    throw new IOException("Invalid key: " + key);
+                }
+            }
+
```
jasonkatonica: At this point in the code should we validate a few of the parameters to ensure we read them from the config file. This should include description and default and name at least from above that have been expected to be parsed.

jasonkatonica (changes requested):

PROVIDER_CONFIG_FORMAT.md
```diff
@@ -0,0 +1,451 @@
...
+
+#### Parser Behavior:
+- Keys with exactly 4 parts where part[2] is `alias` are recognized as alias definitions
+- The value is split by commas with whitespace trimmed: `value.split("\\s*,\\s*")`
+- Processing order:
+  1. Default aliases are added first (if `default = true`)
+  2. Config file `.add` operations are applied
+  3. Config file `.delete` operations are applied
+  4. Config file `.replace` operations are applied (clears list first)
+- Only one operation (add/delete/replace) per service type and algorithm is expected
```
jasonkatonica, line 178: What happens if we dont follow this rule?

PROVIDER_CONFIG_FORMAT.md
```diff
@@ -0,0 +1,451 @@
...
+
+```java
+public class ServiceDefinition {
+    private final String type;              // Service type (e.g., "Cipher")
+    private final String algorithm;         // Algorithm name (e.g., "AES")
+    private final String className;         // Implementation class
+    private final List<String> aliases;     // List of alias names
+    private final Map<String, String> attributes;  // Attribute key-value pairs
+}
+```
```
jasonkatonica, line 418: I would be ok leaving code out of the user facing documenation for this feature unless there is a reason?

PROVIDER_CONFIG_FORMAT.md
```diff
@@ -0,0 +1,451 @@
+# Provider Configuration File Format Documentation
+
+This document describes the format and syntax for provider configuration files used in OpenJCEPlus, based on the `ProviderServiceReader.java` implementation and example configuration files `ProviderDefAttrs.config` and `ProviderFIPSDefAttrs.config`.
```
jasonkatonica: Giving the user a direct link in the code to these two files would be helpful for them to have as "examples". Over time we could also provide and link other test configuration files such as OpenSSL related files.

PROVIDER_CONFIG_FORMAT.md
```diff
@@ -0,0 +1,451 @@
...
+Register cryptographic service implementations.
+
+#### Format:
+```properties
+Service.<ServiceType>.<AlgorithmName> = <ImplementationClassName>
+```
+
+#### Key Structure:
+- **Part 0**: Must be `Service` (case-insensitive)
+- **Part 1**: Service type (e.g., Cipher, Signature, MessageDigest)
```
jasonkatonica: Should we give the reader all possible values here in case they are not familiar with the JCE service type definitions?

There is a section below "Common Service Types:" that seems to cover them. Maybe we link to that spot here?

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
```diff
@@ -111,1123 +123,45 @@ public OpenJCEPlus() {
         }
```
jasonkatonica, line 101: I think it would be great for us to print in debugs the final services that are being registered after all the parsing, replacing, and joining operations have occurred so its clear what we registered along with various other service attributes, aliases, provider name, and classes we ended up with after we are done processing. Perhaps we could add a helper function to the `OpenJCEPlusProvider` to print everything registered in both the OpenJCEPlus and OpenJCEPlusFIPS class.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
johnpeck-us-ibm: There is no need to have similar methods in OpenJCEPlusFIPS. We only need one place to create these types of providers from. Having two places probably only complicates things and raises expectations. Once, we are done with the adding code to support multiple back ends. This provider really will only needs to be one created from OpenJCEPlus with the correct Configuration file. Today this is needed for the context for OCKC.

src/main/java/com/ibm/crypto/plus/provider/DefaultProviderAttrs.java
johnpeck-us-ibm: fixed

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
johnpeck-us-ibm: Moved for both providers.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
johnpeck-us-ibm: I updated this. But this might be a security issue. The messages are what is indicating the issue. Stack traces may give attackers too much information.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: Removed and pointed to the document

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: This checked when readServices is called.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: This checked when readServices is called.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: done.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: The message is the important information and the trace might expose information that an application may not want to expose.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: Changes to ali

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: The use of Properties class causes the config file to be completely read.

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: Nothing really. Other than the config file will be very difficult to read as a human, if the file is of any size. The code searches the properties file for the entries which allows the entries to be any where in the config file.

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: This is a way to tie in the Java Doc for what this files represents. This might be helpful for someone who understands what we are doing or wants to understand the code better.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
johnpeck-us-ibm: This is not needed. The an empty name will result in the provider being are called "OpenJCEPlus-" and the description is optional anyway.

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: Added.

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: This is for the file. Not for the actual ProviderService for JCE

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
johnpeck-us-ibm: As we register them or after they are already registered? Would be easier to add a debug as it is done. But not sure that is what you want.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: done.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
jasonkatonica: I dont see this as a security issue. In any case we should at least make this servicable such that its possible to guess what happened. We would at least need a debug statement added here for serviceability.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
jasonkatonica: One of our main use cases is for users to be able to support OpenSSL fips mode. Id expect that we should be able to use our FIPS provider yet just configure the openssl library in this case that is in fips mode. In this scenario i think we would need to have the same approach for our FIPS provider. This also brings the question up that we likely will need to make large scale changes through our code to support this setup given that we will no longer have the "OpenJCEPlusFIPS" provider name but instead "OpenJCEPlus-XXX" name.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
jasonkatonica: I dont see this as a security issue. In any case we should at least make this serviceable such that its possible to guess what happened. We would at least need a debug statement added here for serviceability.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
jasonkatonica: I would think that we want to require a name. That way they know that they will have an entirely new provider name. This will also be a big problem for the restricted security mode setup within Semeru runtimes. No longer does our changes work there for FIPS 140-3 enforcement.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
jasonkatonica: Yes i think we shoudl at least require a name. A description seems like it would be ok to be optional. Id also say we should version our file format so its easier to change in the future.

PROVIDER_CONFIG_FORMAT.md
jasonkatonica: OK these still appear to be service names of a certain format and expectation ( even though they look like JCE service names to me ). Should we still document what is expected for the service type?

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
jasonkatonica: I think we want to know what services are registered after our process of reading and parsing the file is all done. I think we want to have in the debug all values for services, aliases, attributes, classes, etc for everything that we parse. This will be the only way to figure out user configurations when they are not working as expected by users.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
johnpeck-us-ibm: The whole expection is part of the rethrown exceptiion. Bu it is now in OpenJCEPlusProvider.java

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
johnpeck-us-ibm: The FIPS provider is our provider and our provider only. We others do is up to them.  That said the design for the attributes that will decide the back end is: 
Create 2 new Attributes:
        •NativeProvider – can have 3 values NULL, OCK or OpenSSL
                                     NULL will default to OCK

        •NativeProviderFIPS – same values and default as above.

These can be set on any algorithm in a config file. If someone wanted only FIPS they would need to use the NativeProviderFIPS attribute for all algorithms. However, mixing back ends (OCK and OpenSSL) in a FIPS provider would not be good and may end up not being complaint.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: I will add a debug here. The issue with a stack traces in this case is that the it will expose the caller's stack as well and that could contain information that could be used to attack that application.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
johnpeck-us-ibm: We can talk about the issues here for the profile code. I do not think you can use custom providers in the out of the box profiles. Our profiles should only deal with our providers. If some one wants to restrict things based on their profile that should be up to them. The issue with that is that a provider's config file could be updated such that the profile really will not do what was originally planned. I think there are ways around this, but it more work for the person creating the profile.

I will add a check to make sure a name is specified.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: Will make name required and will add test for this.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
johnpeck-us-ibm: For Attributes this is not possible. You have to know the name of the Attribute in order to query it. Currently the JVM does not give us a way to get the all the attributes. The display will be similar to the compare function in the tests cases. I will add what I can but it will not be everything from the config file.

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: Sure, I will add them here. However, I expect people will just look at the config files provided.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
jasonkatonica: OK after talking some I understand the approach here. FIPS algorithms will be marked with the FIPS attributes accordingly instead of using the provider that they came from.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
jasonkatonica: Could we modify our `OpenJCEPlusService` service to track this information instead of relying upon the parent `Service` class? For example when using `OpenJCEPlusService(Provider provider, String type, String algorithm, String className, String[] aliases, Map<String, String> attributes)` Could we save the attributes into a `Map<String, String>` of our own for our own tracking purposes ( and debugs )?

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
```diff
@@ -0,0 +1,362 @@
...
+
+    @Test
+    public void testProviderServicesErrorTest() throws Exception {
+        String config = null;
+        BufferedReader rd = null;
+        
+        try {
+            //File not found
+            ProviderServiceReader reader = new ProviderServiceReader("./src/test/ProviderDefAttrs.confg");
+            reader.readServices();
```
jasonkatonica, line 303: I think we would want to add a check if it worked such as `fail("File was expected to not be found");` here just in case it works by accident.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
```diff
@@ -0,0 +1,362 @@
...
+        } catch (IOException ioe) {
+            System.out.println(ioe.getMessage());
+        }
+        try {
+            //File null
+            ProviderServiceReader reader = new ProviderServiceReader(config);
+            reader.readServices();
```
jasonkatonica, line 322: Similar comment here as above.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
```diff
@@ -0,0 +1,362 @@
...
+        } catch (IOException ioe) {
+            System.out.println(ioe.getMessage());
+        }
+        try {
+            //bufferedreader null
+            ProviderServiceReader reader = new ProviderServiceReader(rd);
+            reader.readServices();
```
jasonkatonica, line 341: Similar comment here as above.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
```diff
@@ -0,0 +1,362 @@
...
+            //No Name in config
+            config = "description =  OpenJCEPlus-test Provider\n"
+                + "default = true\n"
+                + "AlgorithmParameters.CCM.attr.add.TestAttr1 = TestValue1\n"
+                + "AlgorithmParameters.CCM.attr.add.TestAttr2 = TestValue2\n"
+                + "AlgorithmParameters.CCM.attr.delete.TestAttr1 = ";
+            rd = new BufferedReader(new StringReader(config));
+            ProviderServiceReader reader = new ProviderServiceReader(rd);
+            reader.readServices();
+        } catch (IOException ioe) {
```
jasonkatonica: Similar comment here as above.

src/test/ProviderDefAttrs.config
```diff
@@ -0,0 +1,1048 @@
+/*
+ * Copyright IBM Corp. 2026
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms provided by IBM in the LICENSE file that accompanied
+ * this code, including the "Classpath" Exception described therein.
+ */
+
```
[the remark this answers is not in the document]
jasonkatonica: We decided to go unversioned for now and can add a version once a file change is made that is not compatible to our initial release.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java
johnpeck-us-ibm: I added more debug. this way if someone creates a new provider they can check the config and the services created.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: Added this test to testProviderServicesErrorTest

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
johnpeck-us-ibm: Done

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
johnpeck-us-ibm: Done

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
johnpeck-us-ibm: Done

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
johnpeck-us-ibm: Done

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

Issue #1138, opened 2026-01-26
Fix RSASSA-PSS native memory leak


farshadasl:
RSASSA-PSS (RSA Signature Scheme with Appendix - Probabilistic Signature Scheme)

The native file RsaPss.c has a potential memory leak:
This line https://github.com/IBM/OpenJCEPlus/blob/7d1995555320c86b347c28c4089bf5710f0feb3e/src/main/native/RsaPss.c#L550 should free the memory in failure:

```java
FREE_N_NULL(ockDigest);
```

@jasonkatonica

PR #1169, opened 2026-02-11
Fix potential memory leak in RSAPSS

jasonkatonica:
When creating a new context with ICC_EVP_MD_CTX_new potential failures can occur. In the case that this API fails we should release the entire ockDigest structure upon failure.

Fixes: https://github.com/IBM/OpenJCEPlus/issues/1138


johnpeck-us-ibm (approved):

taoliult (approved):
LGTM

PR #1175, opened 2026-02-12
Change package name and move native files

KostasTsiounis:
The package for the supporting classes that invoke the native library is changed from `com.ibm.crypto.plus.provider.ock` to `com.ibm.crypto.plus.provider.base`, since the same classes can be used to invoke functionality from other libraries in the future.

The native files pertaining to `OCKC` functionality are moved to a newly created subdirectory (i.e., they are now under `native/ock/` instead of just `native/`).

The makefiles and `pom.xml` are updated to account for these changes.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1123


KostasTsiounis:
The `HKDFKeyDerivation.java` file is missing in 21 and below. Otherwise a clean back-port.

PR #1170, opened 2026-02-11
Add benchmark for RSA asymmetric cipher

KostasTsiounis:
A new benchmark to measure `RSA` ciphers with different paddings is added.

The `CipherBase` is split into two separate classes, namely `AsymmetricCipherBase` and `SymmetricCipherBase` to account for the differences.

Signed-off-by: Kostas Tsiounis <kostas.tsiounis@ibm.com>

PR #1203, opened 2026-02-20
Add benchmark for RSA asymmetric cipher

KostasTsiounis:
A new benchmark to measure `RSA` ciphers with different paddings is added.

The `CipherBase` is split into two separate classes, namely `AsymmetricCipherBase` and `SymmetricCipherBase` to account for the differences.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1170


jasonkatonica (approved):
LGTM

PR #1255, opened 2026-03-09
Use ThreadLocal caching for ExtendedRandom PRNG contexts

taoliult:
Add ThreadLocal caching for native PRNG contexts used by ExtendedRandom. Each thread creates and reuses a PRNG context for supported DRBG algorithms.

This avoids repeated EXTRAND_create calls when instances are created frequently.

Benefits:
- Reduce native allocation overhead
- Reuse PRNG contexts per thread
- Improve performance in RNG-heavy workloads

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
@@ -16,7 +16,13 @@ public final class ExtendedRandom {
 
     private OpenJCEPlusProvider provider;
     private NativeInterface nativeInterface;
-    final long ockPRNGContextId;
+    private final String algName;
+
+    private long ockPRNGContextId;
+    private boolean usingThreadLocalContext = true;
+
+    private static final ThreadLocal<PRNGContextPointer> prngContextBufferSha256 = new ThreadLocal<PRNGContextPointer>();
+    private static final ThreadLocal<PRNGContextPointer> prngContextBufferSha512 = new ThreadLocal<PRNGContextPointer>();
 
     public static ExtendedRandom getInstance(String algName, OpenJCEPlusProvider provider)
             throws NativeException {
@@ -32,11 +38,35 @@ public final class ExtendedRandom {
     }
 
     private ExtendedRandom(String algName, OpenJCEPlusProvider provider) throws NativeException {
+        this.algName = algName;
         this.provider = provider;
         this.nativeInterface = provider.isFIPS() ? NativeOCKAdapterFIPS.getInstance() : NativeOCKAdapterNonFIPS.getInstance();
-        this.ockPRNGContextId = this.nativeInterface.EXTRAND_create(algName);
+        this.ockPRNGContextId = getPRNGContext(algName);
+    }
 
-        this.provider.registerCleanable(this, cleanOCKResources(ockPRNGContextId, nativeInterface));
+    private long getPRNGContext(String algName) throws NativeException {
+        PRNGContextPointer prngCtx = null;
+        ThreadLocal<PRNGContextPointer> prngCtxBuffer = null;
+
+        switch (algName) {
+            case "SHA256":
+                prngCtxBuffer = prngContextBufferSha256;
+                break;
+            case "SHA512":
+                prngCtxBuffer = prngContextBufferSha512;
+                break;
+            default:
+                throw new IllegalArgumentException(
+                        "Unsupported HASHDRBG algorithm: " + algName);
+        }
+
+        prngCtx = prngCtxBuffer.get();
+        if (prngCtx == null) {
+            prngCtx = new PRNGContextPointer(algName, this.nativeInterface, this.provider);
+            prngCtxBuffer.set(prngCtx);
+        }
+
+        return prngCtx.getCtx();
     }
 
     public synchronized void nextBytes(byte[] bytes) throws NativeException {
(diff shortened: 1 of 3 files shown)


src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
```diff
@@ -30,18 +34,44 @@ public static ExtendedRandom getInstance(OCKContext ockContext, String algName,
...
+                prngCtxBufer.set(prngCtx);
+            }
+            return prngCtx.getCtx();
+        } else {
+            return 0;
+        }
     }
 
-    public synchronized void nextBytes(byte[] bytes) throws OCKException {
+    public void nextBytes(byte[] bytes) throws OCKException {
```
KostasTsiounis: Did we find a way to check that ICC doesn't have an issue with concurrent calls, even with separate contexts?

KostasTsiounis (changes requested):

johnpeck-us-ibm (approved):

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
```diff
@@ -16,6 +16,9 @@ public final class ExtendedRandom {
     OCKContext ockContext;
     final long ockPRNGContextId;
 
+    private static final ThreadLocal<PRNGContextPointer> prngContextBufferSha256 = new ThreadLocal<PRNGContextPointer>();
```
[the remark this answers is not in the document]
taoliult: Sorry for the late update. I ran all the benchmark tests in the `SecurityPerformancePipeline` on both `ppc64_aix` and `x86-64_windows`, and the results look good. I did not see any issues.

src/test/java/ibm/jceplus/jmh/RandomNewInstanceBenchmark.java
```diff
@@ -0,0 +1,67 @@
...
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+@State(Scope.Benchmark)
+@Warmup(iterations = 3, time = 10, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 4, time = 30, timeUnit = TimeUnit.SECONDS)
+public class RandomNewInstanceBenchmark extends JMHBase {
```
jasonkatonica, line 32: Since this is a new test it would need to be added to the JenkinsfilePerformance file such that is available for users to select.

jasonkatonica (changes requested):

taoliult:
@johnpeck-us-ibm Sorry, I accidentally clicked the refresh button. I noticed that you had already approved this PR, but I’m not sure if you need to click Approve button again because of that.

src/test/java/ibm/jceplus/jmh/RandomNewInstanceBenchmark.java
taoliult: Yes, added into JenkinsfilePerformance.

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
```diff
@@ -16,7 +16,13 @@ public final class ExtendedRandom {
 
     private OpenJCEPlusProvider provider;
     private NativeInterface nativeInterface;
-    final long ockPRNGContextId;
+    private final String algName;
+
+    long ockPRNGContextId;
```
KostasTsiounis: Shouldn't this be `private` too?

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
```diff
@@ -55,22 +85,63 @@ public synchronized void setSeed(byte[] seed) throws NativeException {
...
 
         if (seed.length > 0) {
+            createNonSharedContextForReSeed();
             this.nativeInterface.EXTRAND_setSeed(ockPRNGContextId, seed);
         }
     }
 
-    private Runnable cleanOCKResources(long ockPRNGContextId, NativeInterface nativeInterface) {
+    private void createNonSharedContextForReSeed() throws NativeException {
+        if (!usingSharedContext) {
```
KostasTsiounis: I think the name of the variable here is a bit misleading. It's not whether you're using a shared context or not (it's not a choice at this point since you will create another one), but rather if the instance context has been initialized. I think a name like `instanceContextInitialized` or something similar would make this more understandable for future readers.

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
```diff
@@ -55,22 +85,63 @@ public synchronized void setSeed(byte[] seed) throws NativeException {
...
+                    }
+                } catch (Exception e) {
+                    if (OpenJCEPlusProvider.getDebug() != null) {
+                        OpenJCEPlusProvider.getDebug().println("An error occurred while cleaning : " + e.getMessage());
+                        e.printStackTrace();
+                    }
+                }
+            };
+        }
+    }
```
KostasTsiounis: Given that this is a static method, can you reuse the one in the enclosing class?

KostasTsiounis (changes requested):

jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
taoliult: Yes, updated.

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
taoliult: Agree. I renamed `usingSharedContext` to `usingThreadLocalContext`, since this flag indicates whether the it is still using the thread-local context.

I also renamed `createNonSharedContextForReSeed()` to `createInstanceContextForReSeed()`, since the new context in this scenario is instance-specific and separate from the thread-local shared context.

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
taoliult: Agree. I removed the duplicate static cleanup method from `PRNGContextPointer` and reused the cleanup method from the ExtendedRandom class.

KostasTsiounis (approved):
LGTM

JinhangZhang (approved):
LGTM

PR #1304, opened 2026-03-26
Migrate CI CD infrastructure to GA version Java 26

jasonkatonica:
Java 26 is now generally available. Our github action and Jenkins pipeleines should be updated accordingly.

Signed-off-by: Jason Katonica <katonica@us.ibm.com>

PR #1351, opened 2026-04-08
Migrate CI CD infrastructure to GA version Java 26

jasonkatonica:
Java 26 is now generally available. Our github action and Jenkins
pipe lines should be updated accordingly.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1304

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -103,6 +103,26 @@ def getBinaries(hardware, software) {
 }
 
 /*
+ * Constructs the appropriate Semeru JDK download URL based on the Java version,
+ * hardware platform, and software OS. Supports both latest GA releases and
+ * specific version releases using the AdoptOpenJDK API.
+ */
+def getJavaDownloadUrl(javaVersion, hardware, software, javaRelease) {
+    def java_link = ""
+    
+    if (javaRelease == "") {
+        // Use latest GA version
+        java_link = "https://api.adoptopenjdk.net/v3/binary/latest/${javaVersion}/ga/${software}/${hardware}/jdk/openj9/normal/ibm?project=jdk"
+    } else {
+        // Use specific version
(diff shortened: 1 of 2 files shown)


jasonkatonica:
I plan to backport this code base further to java 11 once its understood that this approach is what we want to do. At that time i will also update each Java build to the latest for a given release.

taoliult (approved):
LGTM

utils.groovy
```diff
@@ -117,34 +137,11 @@ def getJava(hardware, software) {
...
-    } else {
-        def java_release_link = JAVA_RELEASE.replace("+", "%2B")
-        java_link = "https://api.adoptopenjdk.net/v3/binary/version/${java_release_link}/${software}/${hardware}/jdk/openj9/normal/ibm?project=jdk"
-    }
+    // Get the download URL using the helper method that uses an API
+    def java_link = getJavaDownloadUrl(JAVA_VERSION, hardware, software, JAVA_RELEASE)
 
     dir("java") {
-        sh "curl -u $ARTIFACTORY_USERNAME:$ARTIFACTORY_PASSWORD ${java_link} > java.tar.gz"
+        sh "curl -LJkO ${java_link}"
```
JinhangZhang, line 144: Maybe we can add `-fS` for curl `LJkO` command as well otherwise, if the java_link doesn't work, we will not know

jasonkatonica: Hi Jinhang this doesnt work if downloading from artifactory and our code has been flip flopping over time on main branch due to this different. This particular download of the file so seems to be ok as it is now.

JinhangZhang (approved):
LGTM

johnpeck-us-ibm (approved):

PR #1374, opened 2026-04-13
Support for Artifactory URL downloads in pipelines

jasonkatonica:
Enhanced the JAVA_RELEASE parameter to accept either a release version
or a direct Artifactory URL for Java builds. This provides more
flexibility in specifying custom Java builds during CI/CD execution.

Changes:
- Updated Jenkinsfile and JenkinsfilePerformance parameter descriptions
  to document two options: release version or Artifactory URL
- Modified utils.groovy getJava() function to detect and handle
  Artifactory URLs (na.artifactory.swg-devops.com)
- Added validation to reject non-Artifactory URLs for security
- Implemented authenticated downloads using ARTIFACTORY_USERNAME and
  ARTIFACTORY_PASSWORD credentials for Artifactory URLs
- Maintained backward compatibility with existing release version format
- Detect if JDK extracted already results in a jdk directory

The implementation ensures that only trusted Artifactory URLs are
accepted and automatically applies authentication when downloading
from Artifactory, while preserving the existing behavior for standard
release versions.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -184,15 +184,40 @@ def getJava(hardware, software) {
         hardware = "x64"
     }
 
-    // Get the download URL using the helper method that uses an API
-    def java_link = getJavaDownloadUrl(JAVA_VERSION, hardware, software, JAVA_RELEASE)
+    def java_link = ""
+    
+    // Check if JAVA_RELEASE is a URL that contains artifactory.
+    if (JAVA_RELEASE.contains("https://na.artifactory.swg-devops.com")) {
+        // JAVA_RELEASE is an Artifactory URL
+        java_link = JAVA_RELEASE
+        echo "Using Artifactory URL from JAVA_RELEASE: ${java_link}"
+    } else if (JAVA_RELEASE.startsWith("http://") || JAVA_RELEASE.startsWith("https://")) {
+        // JAVA_RELEASE is a URL but not from Artifactory - fail the build
+        error("JAVA_RELEASE contains a URL that is not from na.artifactory.swg-devops.com. Only Artifactory URLs are supported.")
+    } else {
+        // JAVA_RELEASE is a version string or empty - use the API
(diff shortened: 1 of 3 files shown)


JinhangZhang (approved):
LGTM

utils.groovy
```diff
@@ -172,15 +172,34 @@ def getJava(hardware, software) {
...
     dir("java") {
-        //sh "curl -u $ARTIFACTORY_USERNAME:$ARTIFACTORY_PASSWORD ${java_link} > java.tar.gz"
-        sh "curl -LJkO ${java_link}"
+        // Download Java - use credentials if it's an Artifactory URL
+        if (JAVA_RELEASE.contains("na.artifactory.swg-devops.com")) {
+            if (software == "windows") {
+                sh "curl -u \$ARTIFACTORY_USERNAME:\$ARTIFACTORY_PASSWORD ${java_link} > java.zip"
+            } else {
+                sh "curl -u \$ARTIFACTORY_USERNAME:\$ARTIFACTORY_PASSWORD ${java_link} > java.tar.gz"
+            }
```
KostasTsiounis: Should we extract and set the different ending per platform, but keep the command uniform? It could be helpful in case z/OS wants to use that too. I'm thinking something like:
```groovy
def file_extension = "tar.gz"
if (software == "windows") {
    file_extension = "zip"
} 
sh "curl -u \$ARTIFACTORY_USERNAME:\$ARTIFACTORY_PASSWORD ${java_link} > java.${file_extension}"
```

KostasTsiounis (changes requested):

utils.groovy
jasonkatonica: Updated code as suggested to isolate the file extension part. Hopefully this reads ok.

KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

PR #1409, opened 2026-04-22
Add TLS handshake benchmark and enhance JMH base

jasonkatonica:
- Add TLSHandshakeBenchmark for measuring TLS 1.3 handshake performance
  * Support for multiple named groups: X25519, X25519MLKEM768, SecP256r1, SecP256r1MLKEM768, SecP384r1, SecP384r1MLKEM1024
  * Configurable session caching (cached/non-cached) to test full vs resumed handshakes
  * Provider comparison support (OpenJCEPlus vs SunJCE)
  * Uses hardcoded EC P-256 certificate for consistent testing
  * Single-threaded design to prevent client/server deadlocks
  * 5-minute socket timeout protection

- Enhance JMHBase provider setup and diagnostics
  * Add Java version logging (version, runtime version, VM version)
  * Fix provider registration by removing existing provider before re-insertion
  * Add OpenJCEPlusFIPS provider support with proper cleanup
  * Ensure BC provider is properly removed before re-registration

### DIFF
--- a/src/test/java/ibm/jceplus/jmh/TLSHandshakeBenchmark.java
+++ b/src/test/java/ibm/jceplus/jmh/TLSHandshakeBenchmark.java
@@ -1,0 +1,277 @@
+/*
+ * Copyright IBM Corp. 2026, 2026
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms provided by IBM in the LICENSE file that accompanied
+ * this code, including the "Classpath" Exception described therein.
+ */
+
+package ibm.jceplus.jmh;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.SocketTimeoutException;
+import java.security.KeyFactory;
+import java.security.KeyStore;
+import java.security.PrivateKey;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateFactory;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.Base64;
+import java.util.concurrent.TimeUnit;
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLServerSocket;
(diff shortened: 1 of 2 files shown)


src/test/java/ibm/jceplus/jmh/TLSHandshakeBenchmark.java
```diff
@@ -0,0 +1,277 @@
...
+
+    private static final int payload = 1024;
+    private static final String cipherSuite = "TLS_AES_256_GCM_SHA384";
+
+    // Hardcoded EC certificate (P-256) from BaseTestTLS.java (ecdsa_sha256 enum)
+    // Source: src/test/java/ibm/jceplus/junit/base/integration/BaseTestTLS.java
+    // Signature Algorithm: ecdsa-with-SHA256
+    // Subject: CN = Unknown
+    // Public Key Algorithm: id-ecPublicKey (P-256/secp256r1)
+    private static final String EC_CERT =
```
johnpeck-us-ibm, line 65: Is there any way you could generate the certificates on the fly? This issue is at some point this certificate will expire and this test will fail.

jasonkatonica: This certificate was copied from the existing test at `src/test/java/ibm/jceplus/junit/base/integration/BaseTestTLS.java`. I decoded this certificate and it has expired already. I dont believe that the self signed certificate that is added to the truststore/keystore will actually fail validation given that this has been working in the past so i think this is OK.

I actually did start off by using keytool to generate the certificate but this would involve adding JAVA_HOME correctly in our pipelines and also in JDK pipelines such that the correct version of Java is used with the processbuilder / keytool command. This seemed a bit less reliable then in-lining the certificate and key material as we always know the runtime matches what is under test.

johnpeck-us-ibm: That is fine. As long as the test works,

johnpeck-us-ibm (approved):

JinhangZhang (approved):
LGTM

taoliult (approved):
LGTM

KostasTsiounis (approved):
LGTM

PR #1412, opened 2026-04-28
Create native exception hierarchy and simplify handling

KostasTsiounis:
A new native exception hierarchy is created to include a `NativeException` superclass. This will allow for different native exception subclassses thrown by different native libraries.

The `try-catch` blocks were also simplified to improve readability.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright IBM Corp. 2026
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms provided by IBM in the LICENSE file that accompanied
+ * this code, including the "Classpath" Exception described therein.
+ */
+
+package com.ibm.crypto.plus.provider.base;
+
+/**
+ * This is a class used for exceptions created by the native code that
+ * utilizes the equivalent native libraries.
+ *
+ * The class can be subclassed by library-specific exception classes, or
+ * be used directly by other classes who need to indicate that the problem
+ * occured in native code, but might not have knowledge of the exact library
+ * used.
+ */
+public class NativeException extends java.lang.Exception {
+
+    private static final long serialVersionUID = 9223372036854775807L;
+
+    protected int code = -1; // Non-specific value designed to be overriden.
+
+    // These codes must be overriden by library-specific native exceptions.
+    public static final int GKR_FIPS_MODE_INVALID = -1;
+    public static final int GKR_OCK_ATTACH_FAILED = -1;
+    public static final int GKR_DECRYPT_FINAL_BAD_PADDING_ERROR = -1;
+    public static final int GKR_UNSPECIFIED = -1;
+
+    public NativeException(String s) {
+        super(s);
+    }
+
+    public NativeException(String s, Throwable cause) {
+        super(s, cause);
+    }
+
+    public int getCode() {
+        return code;
+    }
+}
--- a/src/main/java/com/ibm/crypto/plus/provider/ock/OCKException.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ock/OCKException.java
@@ -6,28 +6,28 @@
  * this code, including the "Classpath" Exception described therein.
  */
 
-package com.ibm.crypto.plus.provider.base;
+package com.ibm.crypto.plus.provider.ock;
 
+import com.ibm.crypto.plus.provider.base.NativeException;
 import java.util.Hashtable;
 import java.util.Map;
 
-public final class OCKException extends java.lang.Exception {
+public class OCKException extends NativeException {
 
     /**
      * 
      */
     private static final long serialVersionUID = -3104732494450550839L;
 
-    // These codes must match those defined in ExceptionCodes.h.
-    // 
+    /* These codes are overriding the ones specified in the superclass NativeException
+     * and must match those defined in native/ock/ExceptionCodes.h.
+     */
     public static final int GKR_FIPS_MODE_INVALID = 0x00000001;
     public static final int GKR_OCK_ATTACH_FAILED = 0x00000002;
     public static final int GKR_DECRYPT_FINAL_BAD_PADDING_ERROR = 0x00000003;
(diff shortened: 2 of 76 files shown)


src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright IBM Corp. 2026
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms provided by IBM in the LICENSE file that accompanied
+ * this code, including the "Classpath" Exception described therein.
+ */
+
+package com.ibm.crypto.plus.provider.base;
+
+public class NativeException extends java.lang.Exception {
```
jasonkatonica, line 20: Can we declare a UUID for this class since it is serializable? All `Throwables` [are serializable](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html) and should have a UUID.

src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright IBM Corp. 2026
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms provided by IBM in the LICENSE file that accompanied
+ * this code, including the "Classpath" Exception described therein.
+ */
+
+package com.ibm.crypto.plus.provider.base;
+
```
jasonkatonica, line 10: Some javadoc to describe what this new class is ?

src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
```diff
@@ -0,0 +1,33 @@
...
+public class NativeException extends java.lang.Exception {
+
+    protected int code = -1; // Non-specific value designed to be overriden.
+
+    // These codes be overriden by specific native exceptions.
```
jasonkatonica: This reads a bit off. Perhaps could do `'These codes can be overridden by specific native exceptions.`

src/main/java/com/ibm/crypto/plus/provider/ock/OCKException.java
```diff
@@ -6,29 +6,29 @@
  * this code, including the "Classpath" Exception described therein.
  */
 
-package com.ibm.crypto.plus.provider.base;
```
jasonkatonica, line 9: Is it possible somebody is referring to this exception externally ? I assume not given none of our tests needed to be updated.....I wasnt sure why this could not be externally referenced but there must be a reason.

KostasTsiounis: I don't believe so. Our higher level classes always wrap this in one of the more well-known and JDK compatible exceptions. I, also, don't see a reason to reference it as a user, given that it is only used to indicate problems with the native code, which we don't directly expose.

src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
KostasTsiounis: Added one.

src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
KostasTsiounis: Added a description.

src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
KostasTsiounis: Changed it to make sense.

src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private static final long serialVersionUID = 9223372036854775807L;
+
+    protected int code = -1; // Non-specific value designed to be overriden.
+
+    // These codes must be overriden by library-specific native exceptions.
+    public static final int GKR_FIPS_MODE_INVALID = -1;
+    public static final int GKR_OCK_ATTACH_FAILED = -1;
+    public static final int GKR_DECRYPT_FINAL_BAD_PADDING_ERROR = -1;
+    public static final int GKR_UNSPECIFIED = -1;
```
johnpeck-us-ibm, line 30: Are these really needed here? They seem to be only related to OCKC and the comment even indicates these need to be overriden.

johnpeck-us-ibm (changes requested):

src/main/java/com/ibm/crypto/plus/provider/ock/OCKException.java
jasonkatonica: OK given that none of our tests use these i dont expect anybody to use them externally. Additionally the [module-info](https://github.com/IBM/OpenJCEPlus/blob/main/src/main/java/module-info.java) has not exposed this package to users so seems like somebody would have to go out of their way to reference this class.

src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
KostasTsiounis: I originally didn't have them here, but they are used by `Poly1305Cipher.java` [here](https://github.com/IBM/OpenJCEPlus/pull/1412/changes#diff-3cb40242e389c7c44a8bdb16f936915a93c11523e757e1da4bda6cafdedc6c00R363) (this is the only spot by the way). The java classes are supposed to be native library agnostic, so I didn't want to refer to `OCKException` directly.

We can open an issue to improve this part of the code and remove these error codes from the `NativeException` class at that point.

jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/base/NativeException.java
johnpeck-us-ibm: I am okay with that.

johnpeck-us-ibm (approved):

JinhangZhang (approved):
LGTM

taoliult (approved):
LGTM

PR #1419, opened 2026-04-30
Extract library-specific elements to separate files

KostasTsiounis:
For each of the platform-specific makefiles, the rules and variables that are applicable to all potential native libraries are extracted to common makefiles.

The original makefiles only maintain the library-specific elements and include the equivalent common makefiles for the rest of them.

Signed-off-by: Kostas Tsiounis <kostas.tsiounis@ibm.com>

PR #1440, opened 2026-05-05
Extract library-specific elements to separate files

KostasTsiounis:
For each of the platform-specific makefiles, the rules and variables that are applicable to all potential native libraries are extracted to common makefiles.

The original makefiles only maintain the library-specific elements and include the equivalent common makefiles for the rest of them.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1419

### DIFF
--- a/src/main/native/ock/jgskit.mac.mak
+++ b/src/main/native/ock/jgskit.mac.mak
@@ -7,33 +7,12 @@
 # this code, including the "Classpath" Exception described therein.
 ###############################################################################
 
-TOPDIR=../../../..
-
-CFLAGS= -fPIC -DMAC -Werror -std=gnu99 -pedantic -Wall -fstack-protector -m64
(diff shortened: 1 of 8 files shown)


KostasTsiounis:
The difference in lines between this and the 25+ versions is due to the following reasons:
- AIX has been updated to use the `ibm-clang_r` compiler in later versions, which requires different flags => I've maintained the appropriate compiler per version
- z/OS has also been updated to use the `ibm-clang64` compiler in later versions, which requires different flags => I've synced all versions given that they probably won't be directly used
- There was a discrepancy in the `displaycompiler` rule in the linux makefile => it has been synced up to `main` in all versions

taoliult (approved):
LGTM

johnpeck-us-ibm (approved):

PR #1443, opened 2026-05-06
improve ML-KEM algorithm handling and test coverage

jasonkatonica:
- Fix ML-KEM encapsulation length calculation to use actual key algorithm instead of generic 'ML-KEM' string
- Add validation for encapsulation message length in decapsulation
- Improve key conversion to use actual algorithm from key instead of generic algorithm parameter
- Add comprehensive test for invalid encapsulation length handling
- Refactor test skip conditions to use assumeFalse for better JUnit 5 compatibility
- Add new interoperability tests using NamedParameterSpec:
  * testMLKEMInteropWithNamedParameterSpec
  * testMLKEMInteropEmptyParamsWithNamedParameterSpec
  * testMLKEMInteropSmallerSecretWithNamedParameterSpec
  * testMLKEMBidirectionalInteropWithNamedParameterSpec
- Remove unused import and improve code consistency

This ensures ML-KEM operations correctly handle different parameter sets (ML-KEM-512, ML-KEM-768, ML-KEM-1024) and provides better error messages when encapsulation length mismatches occur.


KostasTsiounis (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
```diff
@@ -38,18 +38,22 @@ public MLKEMImpl(OpenJCEPlusProvider provider, String alg) {
         this.alg = alg;
     }
     
-    private int getEncapsulationLength() {
+    private int getEncapsulationLength(String algorithm) {
```
johnpeck-us-ibm, line 64: Since, this is no longer using the Algorithm for this Implementation instance. Couldn't this cause issues with callers mix matching things easier?

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
```diff
@@ -72,8 +76,15 @@ public KEMSpi.EncapsulatorSpi engineNewEncapsulator(PublicKey publicKey,
...
+            // First verify it's an ML-KEM key
+            String keyAlgorithm = publicKey.getAlgorithm();
+            if (keyAlgorithm == null || !keyAlgorithm.startsWith("ML-KEM")) {
+                throw new InvalidKeyException("unsupported key");
+            }
+            
+            // Use the key's actual algorithm, not the generic "ML-KEM"
             try {
-                KeyFactory kf = KeyFactory.getInstance(this.alg, this.provider.getName());
+                KeyFactory kf = KeyFactory.getInstance(keyAlgorithm, this.provider.getName());
```
johnpeck-us-ibm, line 113: This can cause a algorithm miss match if they create this instance as ML-KEM-786, but pass in a ML-KEM-512 key that should not be allowed.

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
```diff
@@ -105,7 +116,9 @@ class MLKEMEncapsulator implements KEMSpi.EncapsulatorSpi {
 
         @Override
         public KEM.Encapsulated engineEncapsulate(int from, int to, String algorithm) {
-            int encapLen = getEncapsulationLength();
+            // Get the actual algorithm from the public key
+            String keyAlgorithm = publicKey.getAlgorithm();
```
johnpeck-us-ibm, line 149: Same issue here.

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
```diff
@@ -210,8 +242,8 @@ public SecretKey engineDecapsulate(byte[] cipherText, int from, int to, String a
 
         @Override
         public int engineEncapsulationSize() {
-
-            return getEncapsulationLength();
+            String keyAlgorithm = privateKey.getAlgorithm();
+            return getEncapsulationLength(keyAlgorithm);
```
johnpeck-us-ibm, line 281: Same type of issue here too. The algorithm the should be based on the one requested when this was created.

JinhangZhang (approved):
LGTM

taoliult (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/MLKEMImpl.java
jasonkatonica: Good catch here. Agreed that we could end up having the opposite problem here ( just using the key algorithm associated with the key instead of using the algorithm used by the getInstance call ). I added commit [1c01248](https://github.com/IBM/OpenJCEPlus/pull/1443/commits/1c0124860ed1ec94dd7ee5c99f39005fefc611f6) to validate that the algorithm on the getInstance matches the key if the user specified a specific parameter set. For the case where generics are used such as `ML-KEM` it will still default to the 768 parameter set. I also added some additional tests that recreated this problem and validates this behavior in the new commit.

johnpeck-us-ibm (approved):

PR #1468, opened 2026-05-14
Refactor JUnit 5 Tests for BaseTestDSASignatureInterop

Mohit-Rajbhar100698:
Replace multiple test classes extending BaseTestDSASignatureInterop with a single JUnit 5 parameterized test class


johnpeck-us-ibm (approved):

src/test/java/ibm/jceplus/junit/tests/TestDSASignatureInteropSUN.java
```diff
@@ -21,16 +21,37 @@
...
 
+    @Parameter(0)
+    TestProvider provider;
 
     static final byte[] origMsg = "this is the original message to be signed".getBytes();
 
+    @BeforeEach
+    public void setUp() throws Exception {
+        setAndInsertProvider(provider);
+        setInteropProviderName(TestProvider.SUN.getProviderName());
```
KostasTsiounis: Should we hardcode this here or have the interop provider as a parameter too? @jasonkatonica what do you think?

Mohit-Rajbhar100698: I’m only using enum `TestProvider.SUN` here to get the provider name for the interop provider . 
not passing through the JUnit 5 parameterization.

jasonkatonica: For consistancy i think i woudl prefer using parameters like other interop tests. Sure in this case we are only doing it against SUN however in the future it would be easier to add other providers using same code pattern as other interop tests.

jasonkatonica: Additionally we could remove SUN from the testcase name to generalize it as a interop test for DSA signatures.

jasonkatonica: Also upon further inspection there is a test `TestDSASignatureInteropBC` I think this test should also be included in this update ( and refactored to use parameters and the BC provider )

Mohit-Rajbhar100698: > For consistancy i think i woudl prefer using parameters like other interop tests. Sure in this case we are only doing it against SUN however in the future it would be easier to add other providers using same code pattern as other interop tests.

I have parameterized the interop provider as well.

Mohit-Rajbhar100698: > Additionally we could remove SUN from the testcase name to generalize it as a interop test for DSA signatures.

Removed SUN from test name

Mohit-Rajbhar100698: > Also upon further inspection there is a test `TestDSASignatureInteropBC` I think this test should also be included in this update ( and refactored to use parameters and the BC provider )

Since `TestDSASignatureInteropBC` extends `BaseTestDSASignatureInterop2`,  As we discussed we will handle that refactoring separately in another PR.

src/test/java/ibm/jceplus/junit/tests/TestDSASignatureInterop.java
```diff
@@ -21,16 +21,40 @@
...
+
+    @Parameter(1)
+    TestProvider interopProvider;
 
     static final byte[] origMsg = "this is the original message to be signed".getBytes();
 
+    @BeforeEach
+    public void setUp() throws Exception {
+        setAndInsertProvider(provider);
+        setInteropProviderName(interopProvider.getProviderName());
```
KostasTsiounis: Would it make sense to do a `setAndInsertProvider()` for the interop provider too, in case it's not in the provider list? I know we are just using the Sun provider here, but I think it's good practice.

KostasTsiounis (changes requested):

jasonkatonica (approved):
LGTM

src/test/java/ibm/jceplus/junit/tests/TestDSASignatureInterop.java
jasonkatonica: After discussion we think we should create a new method named `setAndInsertInteropProvider` for this line.

johnpeck-us-ibm (approved):

src/test/java/ibm/jceplus/junit/tests/TestDSASignatureInterop.java
Mohit-Rajbhar100698: Added `setAndInsertInteropProvider()` method for interop provider.

jasonkatonica (approved):
LGTM

src/test/java/ibm/jceplus/junit/tests/BaseTest.java
```diff
@@ -30,6 +30,19 @@ public void setProviderName(String providerName) {
...
         switch (provider) {
+            case OpenJCEPlus:
+                loadProvider(TestProvider.OpenJCEPlus);
+                break;
+            case OpenJCEPlusFIPS:
+                loadProvider(TestProvider.OpenJCEPlusFIPS);
+                break;
+            default:
+                throw new Exception("Provider not supported: " + provider.getProviderName());
+        }
```
KostasTsiounis: Should we extract this into a helper method?

src/test/java/ibm/jceplus/junit/tests/BaseTest.java
```diff
@@ -30,6 +30,19 @@ public void setProviderName(String providerName) {
...
+    }
+
+    public void setAndInsertInteropProvider(TestProvider interopProvider) throws Exception {
+        switch (interopProvider) {
```
KostasTsiounis: Shouldn't we set the interop provider here with something like `this.interopProvider = interopProvider`? Since this variable is not available in `BaseTest` maybe this method belongs to `BaseTestInterop`.

KostasTsiounis (changes requested):

src/test/java/ibm/jceplus/junit/tests/BaseTest.java
Mohit-Rajbhar100698: ok , we can move this switch statement into a seperate helper method.
Done

src/test/java/ibm/jceplus/junit/tests/BaseTest.java
Mohit-Rajbhar100698: This change have been addressed

src/test/java/ibm/jceplus/junit/tests/BaseTestInterop.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright IBM Corp. 2023, 2026
```
KostasTsiounis: This is entirely new, so the copyright should only say 2026.

src/test/java/ibm/jceplus/junit/tests/BaseTestSignatureInterop.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright IBM Corp. 2023, 2026
```
KostasTsiounis: This is entirely new, so the copyright should only say 2026.

KostasTsiounis (changes requested):

johnpeck-us-ibm (approved):

src/test/java/ibm/jceplus/junit/tests/BaseTestSignatureInterop.java
KostasTsiounis: This wasn't fixed.

KostasTsiounis (approved):
LGTM

PR #1485, opened 2026-05-20
Support for PBMAC1 algorithms

dev-koan:
This update adds support for PBMAC1 algorithms. The algorithms are supported for OpenJCEPlus and OpenJCEPlusFIPS provider. The PBMAC1Core class extends the regular HmacCore class overriding engineInit method as PBMAC1 uses PBKDF2 key derivation.
PBMAC1 tests have also been added.


JinhangZhang:
Did you by any chance make a build and put openjceplus on the top of the order and create a PKCS12 keystore file with keytool command? e.g. by specifying `-J-Dkeystore.pkcs12.macAlgorithm=PBEWithHmacSHA256`

johnpeck-us-ibm:
Since, the need here is for use in keystore. I would like to see some keystore tests that test this functionality. First to make sure this work, but also to make sure the keystores created using our impliementation work when Sun's implementation is used.

src/test/java/ibm/jceplus/junit/base/params/file1.txt
```diff
@@ -0,0 +1,57 @@
+MIIKigIBAzCCCgUGCSqGSIb3DQEHAaCCCfYEggnyMIIJ7jCCBGIGCSqGSIb3DQEH
```
jasonkatonica: For known answers and for test inputs we have typically inlined these into the tests as variables/fields. Could we also do this here given that this is base 64 encoded text and should be ok to inline in the test classes. Similar comment for the other `.txt` files.

src/test/java/ibm/jceplus/junit/base/BaseTestPBMAC1.java
```diff
@@ -0,0 +1,213 @@
+/*
```
jasonkatonica: The various tests should be moved to the new test package `ibm.jceplus.junit.tests`. We are trying to put all new tests there in parameterized form. Similar comment for `BaseTestPBMAC1Interop.java`. Files `TestPBMAC1.java` ( both fips and non fips ) and `TestPBMAC1Interop.java` could also be eliminated once migrated to parameterized form.

README.md
```diff
@@ -441,6 +441,13 @@ Mac                         | HmacSHA3-384                    |X
 Mac                         | HmacSHA3-512                    |X                |X             |              |
 Mac                         | HmacSHA384                      |X                |X             |              |
 Mac                         | HmacSHA512                      |X                |X             |              |
+Mac                         | HmacSHA1                        |                 |X             |              |
+Mac                         | HmacSHA224                      |                 |X             |              |
+Mac                         | HmacSHA256                      |                 |X             |              |
+Mac                         | HmacSHA384                      |X                |X             |              |
+Mac                         | HmacSHA512                      |X                |X             |              |
+Mac                         | HmacSHA512/224                  |X                |X             |              |
+Mac                         | HmacSHA512/256                  |X                |X             |              |
```
jasonkatonica: Some of these lines seem like duplicates in the table. See lines above 435 - 443. The `HmacSHA512/224` and `HmacSHA512/256` algorithms seem new. The others seem like duplicates?

src/main/java/com/ibm/crypto/plus/provider/PBMAC1Core.java
```diff
@@ -0,0 +1,198 @@
...
+        SecretKeySpec cipherKey = null;
+        try {
+            PBKDF2Core kdf = getKDFImpl(kdfAlgo);
+            s = (PBKDF2KeyImpl) kdf.engineGenerateSecret(pbeSpec);
+            derivedKey = s.getEncoded();
+            cipherKey = new SecretKeySpec(derivedKey, kdfAlgo);
+            super.engineInit(cipherKey, null);
+        } catch (InvalidKeySpecException ikse) {
+            throw new InvalidKeyException("Cannot construct PBE key", ikse);
+        } finally {
```
jasonkatonica, line 144: This finally block should explicitly clear sensitive value `s`:
```
            if (s != null) {
                s.clear();
            }
```

jasonkatonica (changes requested):

src/test/java/ibm/jceplus/junit/base/params/file1.txt
dev-koan: Updated.

src/test/java/ibm/jceplus/junit/base/BaseTestPBMAC1.java
dev-koan: Updated.

README.md
dev-koan: Updated.

src/main/java/com/ibm/crypto/plus/provider/PBMAC1Core.java
dev-koan: The PBKDF2Impl class doesn't have a method `clear`.
The opedjdk clear method (https://github.com/ibmruntimes/openj9-openjdk-jdk26/blob/openj9/src/java.base/share/classes/com/sun/crypto/provider/PBKDF2KeyImpl.java#L286)
calls cleaner.clean() not sure if it is required or if can add it in the first place.

dev-koan:
> Since, the need here is for use in keystore. I would like to see some keystore tests that test this functionality. First to make sure this work, but also to make sure the keystores created using our impliementation work when Sun's implementation is used.

I am reading P12 files using PKCS12 KeyStore after adding OpenJCEPlus to the top the security provider list.
If we have the required crypto the openjdk keystore class will use ours.

src/main/java/com/ibm/crypto/plus/provider/PBMAC1Core.java
KostasTsiounis: This `cleaner.clean()` method was added to clean the native key created by OpenSSL. I think a similar approach can be followed here to clear whatever native keys we might have.

johnpeck-us-ibm:
> > Since, the need here is for use in keystore. I would like to see some keystore tests that test this functionality. First to make sure this work, but also to make sure the keystores created using our impliementation work when Sun's implementation is used.
> 
> I am reading P12 files using PKCS12 KeyStore after adding OpenJCEPlus to the top the security provider list. If we have the required crypto the openjdk keystore class will use ours.

It might be fine than. However, sometimes keytool does things a little differently than our code and that causes issues. So, at the bare minimum I think it is best to verify the code using keytool. Even if the test is a one off and not added to our tests.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
```diff
@@ -0,0 +1,573 @@
...
+    public void setUp() throws Exception {
+        setAndInsertProvider(provider);
+    }
+
+    @ParameterizedTest
+    @FieldSource("algorithms")
+    void testPBMACFunctionality(String alg) throws Exception {
+        if (getProviderName().equalsIgnoreCase("OpenJCEPlusFIPS") && (alg.equalsIgnoreCase("PBEWithHmacSHA1") || 
+            alg.equalsIgnoreCase("PBEWithHmacSHA224") || alg.equalsIgnoreCase("PBEWithHmacSHA256"))) {
+            return;
```
jasonkatonica: Instead of returning and assuming that the test was run we instead have been using assume statements. Something like the following would assume we are not using OpenJCEPlusFIPS for example and record it as skipped in Junit. Similar comment to other places where you check for fips provider and excute `return`.

```
assumeFalse("OpenJCEPlusFIPS".equals(getProviderName()) && ... .... ...    );
```

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
```diff
@@ -0,0 +1,573 @@
...
+        SecretKey key = new SecretKeySpec(PASSWORD.toString().getBytes(), alg);
+        Mac mac = Mac.getInstance(alg, getProviderName());
+
+        try {
+            byte[] smallSalt = new byte[8];
+            secureRandom.nextBytes(smallSalt);
+            mac.init(key, new PBEParameterSpec(smallSalt, iterationCount));
+            fail("Expected InvalidKeyException not thrown, small salt length");
+        } catch (InvalidKeyException e) {
+            assertTrue(true);
```
jasonkatonica: No reason to assert true. You could however assert that the exception message is as expected in all these locations where we are doing an a`ssertTrue(true)`;

src/main/java/com/ibm/crypto/plus/provider/PBMAC1Core.java
jasonkatonica: After looking again I think we are ok in this spot without explicitly clearing data since the code currently registers the `PBKDF2KeyImpl` sensitive data with a cleaner already on [this line](https://github.com/IBM/OpenJCEPlus/blob/main/src/main/java/com/ibm/crypto/plus/provider/PBKDF2KeyImpl.java#L135) this should later clear all the sensitive fields and release native memory.

The value of `s` should go out of scope no matter what path we take in this code so perhaps a finally and cleanup logic is not really needed afterall.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
```diff
@@ -0,0 +1,573 @@
...
+            "PBEWithHmacSHA512", "PBEWithHmacSHA512/224", "PBEWithHmacSHA512/256");
+    private List<String> fipsalgorithms = Arrays.asList("PBEWithHmacSHA384", "PBEWithHmacSHA512", "PBEWithHmacSHA512/224", "PBEWithHmacSHA512/256");
+
+    private final String message = "This is a message for PBMAC1 testing";
+    private final char[] PASSWORD = "passwordtryagain".toCharArray();
+    private SecureRandom secureRandom = new SecureRandom();
+    private byte[] salt = new byte[20];
+    private int iterationCount = 300000;
+
+    private String p12file1 = "MIIKigIBAzCCCgUGCSqGSIb3DQEHAaCCCfYEggnyMIIJ7jCCBGIGCSqGSIb3DQEH" +
```
jasonkatonica, line 68: Can you please add some comments here as to where these pkcs12 file values were found. I believe that these were found in an RFC as a known answer test? Besides a comment saying where we obtained these values ideally we would add a comment showing the human readable asn1 breakdown of the value so we can easily know what is within these pcks12 files in terms of crypto. 

Also my understanding is that these are not really PKCS12 standard files. They are RFC 9579 (and its successor, RFC 9879) format files correct?

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
```diff
@@ -0,0 +1,573 @@
...
+            Security.insertProviderAt(new OpenJCEPlusFIPS(), 1);
+        }
+        KeyStore ks = KeyStore.getInstance("PKCS12");
+
+        String cleanedBase64 = p12file1.replaceAll("\\s+", "");
+        byte[] decodedBytes = Base64.getDecoder().decode(cleanedBase64);
+        try {
+            ks.load(new ByteArrayInputStream(decodedBytes), "1234".toCharArray());
+        } catch (Exception e) {
+            fail(e.getMessage());
```
jasonkatonica: No need to strip out all the stack trace information or even to catch the exception. The try catch can be removed and just allow the exception to be thrown in the case of a failure. Similar comment for the rest of the checks being done here. 

Maybe the test should also be parameterized such that there is not so much duplicate cod here taking in a pkcs12 file instead?

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
```diff
@@ -0,0 +1,573 @@
...
+        } catch (java.io.IOException e) {
+            fail(e.getMessage());
+        }
+
+        cleanedBase64 = p12file4.replaceAll("\\s+", "");
+        decodedBytes = Base64.getDecoder().decode(cleanedBase64);
+        try {
+            ks.load(new ByteArrayInputStream(decodedBytes), "1234".toCharArray());
+        } catch (java.io.IOException e) {
+            // Expected error (Incorrect iteration count)
```
jasonkatonica, line 567: Should we assert the error message? Seems checking we are are getting a reasonable error message would be better then accepting any IOException that occurs. Similar comment on line 561 and 569.

jasonkatonica (changes requested):

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan: Updated.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan: Updated.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan: The decoded value is extremely long and on directly copying pasting as text there is no indentation making it hard to read.
can still add it if preferred.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan: Updated.
Now using a helper method and letting the exception get thrown.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan:
```
WARNING: OpenJCEPlusFIPS is running in developer mode. Non production workload assumed. This environment is not certified for FIPS 140-3: Mac OS X:aarch64
WARNING: OpenJCEPlusFIPS is running in developer mode. Non production workload assumed. This environment is not certified for FIPS 140-3: Mac OS X:aarch64
Integrity check failed: java.security.UnrecoverableKeyException: Failed PKCS12 integrity checking
```
This is the exception message not sure if asserting the message provides much value.

dev-koan:
> > > Since, the need here is for use in keystore. I would like to see some keystore tests that test this functionality. First to make sure this work, but also to make sure the keystores created using our impliementation work when Sun's implementation is used.
> > 
> > 
> > I am reading P12 files using PKCS12 KeyStore after adding OpenJCEPlus to the top the security provider list. If we have the required crypto the openjdk keystore class will use ours.
> 
> It might be fine than. However, sometimes keytool does things a little differently than our code and that causes issues. So, at the bare minimum I think it is best to verify the code using keytool. Even if the test is a one off and not added to our tests.

I took the following steps to test PBMAC1 with keytool (with Bob's help)

1. A personal build with `OpenJCEPlus` first and bundled with PBMAC1 algos
2. Used the generated jdk to generate my p12 file using keytool with the command `jdk/bin/keytool -J-Dkeystore.pkcs12.macAlgorithm=PBEWithHmacSHA256 -genkeypair -alias mykey -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore mykeystore.p12 -validity 365`
3. Exported it to base64 to verify the crypto (Screenshot of the decoding is attached below)
4. Ran the following program to validate everything was working as expected.
```
public class run {
    public void main() throws Exception {
        // Security.insertProviderAt(new OpenJCEPlusFIPS(), 1);
        System.out.println(Security.getProviders()[0]);
        Mac t = Mac.getInstance("PBEWithHmacSHA384");
        System.out.println(t.getProvider().getName());

        KeyStore ks = KeyStore.getInstance("PKCS12");
        ks.load(new FileInputStream("mykeystore.p12"), "ibmcanada".toCharArray());

        String alias = "mykey";
        PrivateKey privateKey = (PrivateKey) ks.getKey(alias, "ibmcanada".toCharArray());
        Certificate cert = ks.getCertificate(alias);
        PublicKey publicKey = cert.getPublicKey();
        
        // 3. Sign some data
        Signature signature = Signature.getInstance("SHA256withRSA");
        signature.initSign(privateKey);
        String data = "Test data to sign";
        signature.update(data.getBytes());
        byte[] signedData = signature.sign();
        
        // 4. Verify the signature
        signature.initVerify(publicKey);
        signature.update(data.getBytes());
        boolean verified = signature.verify(signedData);
        
        System.out.println("P12 file is valid: " + verified);
    }
}
```


image

johnpeck-us-ibm:
> > > > Since, the need here is for use in keystore. I would like to see some keystore tests that test this functionality. First to make sure this work, but also to make sure the keystores created using our impliementation work when Sun's implementation is used.
> > > 
> > > 
> > > I am reading P12 files using PKCS12 KeyStore after adding OpenJCEPlus to the top the security provider list. If we have the required crypto the openjdk keystore class will use ours.
> > 
> > 
> > It might be fine than. However, sometimes keytool does things a little differently than our code and that causes issues. So, at the bare minimum I think it is best to verify the code using keytool. Even if the test is a one off and not added to our tests.
> 
> I took the following steps to test PBMAC1 with keytool (with Bob's help)
> 
>     1. A personal build with `OpenJCEPlus` first and bundled with PBMAC1 algos
> 
>     2. Used the generated jdk to generate my p12 file using keytool with the command `jdk/bin/keytool -J-Dkeystore.pkcs12.macAlgorithm=PBEWithHmacSHA256 -genkeypair -alias mykey -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore mykeystore.p12 -validity 365`
> 
>     3. Exported it to base64 to verify the crypto (Screenshot of the decoding is attached below)
> 
>     4. Ran the following program to validate everything was working as expected.
> 
> 
> ```
> public class run {
>     public void main() throws Exception {
>         // Security.insertProviderAt(new OpenJCEPlusFIPS(), 1);
>         System.out.println(Security.getProviders()[0]);
>         Mac t = Mac.getInstance("PBEWithHmacSHA384");
>         System.out.println(t.getProvider().getName());
> 
>         KeyStore ks = KeyStore.getInstance("PKCS12");
>         ks.load(new FileInputStream("mykeystore.p12"), "ibmcanada".toCharArray());
> 
>         String alias = "mykey";
>         PrivateKey privateKey = (PrivateKey) ks.getKey(alias, "ibmcanada".toCharArray());
>         Certificate cert = ks.getCertificate(alias);
>         PublicKey publicKey = cert.getPublicKey();
>         
>         // 3. Sign some data
>         Signature signature = Signature.getInstance("SHA256withRSA");
>         signature.initSign(privateKey);
>         String data = "Test data to sign";
>         signature.update(data.getBytes());
>         byte[] signedData = signature.sign();
>         
>         // 4. Verify the signature
>         signature.initVerify(publicKey);
>         signature.update(data.getBytes());
>         boolean verified = signature.verify(signedData);
>         
>         System.out.println("P12 file is valid: " + verified);
>     }
> }
> ```
> 
> image

Did you verify that keytool used OpenJCEPlus when it created the file?

dev-koan:
> Did you verify that keytool used OpenJCEPlus when it created the file?

The KeyStore used is provided by SunJCE because we dont support PKCS12 KeyStore but a good chunk (including PBMAC1) of underlying crypto used will be coming from `OpenJCEPlus`.
In my program i did a `getInstance` of a Mac algorithm to verify that crypto does come from `OpenJCEPlus`.
I tried using `openssl/keytool` info commands to get PMAC1 provider but the output doesn't include it.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
jasonkatonica: OK we can leave the comment out. We do say where we got these values from and if needed somebody can asn.1 decode them later instead of inlining everything here.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
jasonkatonica: Sure i am ok for now asserting we get `Failed PKCS12 integrity checking` this also allows us to document in a round about way what happens during this scenerio ( incorrect iteration count )

README.md
```diff
@@ -442,6 +442,13 @@ Mac                         | HmacSHA3-384                    |X
 Mac                         | HmacSHA3-512                    |X                |X             |              |
 Mac                         | HmacSHA384                      |X                |X             |              |
 Mac                         | HmacSHA512                      |X                |X             |              |
+Mac                         | PBEWithHmacSHA1                 |                 |X             |              |
```
jasonkatonica, line 445: Should we add a few of these to a JMH performance test ? Seems we should add at least PBEWithHmacSHA256 and PBEWithHmacSHA512

jasonkatonica (changes requested):

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan: Updated.

dev-koan:
# x86-linux

### Generating the tag (`doFinal()` operation only )
image

### Initialization with tag generation (`doFinal()` operation with key derivation)
image

# ppc64le

### Generating the tag (`doFinal()` operation only )
image


### Initialization with tag generation (`doFinal()` operation with key derivation)
image

# aix

### Generating the tag (`doFinal()` operation only )
image



### Initialization with tag generation (`doFinal()` operation with key derivation)
image

README.md
dev-koan: Updated.

johnpeck-us-ibm (approved):

jasonkatonica (approved):
LGTM

src/test/java/ibm/jceplus/jmh/PBMAC1Benchmark.java
```diff
@@ -0,0 +1,83 @@
...
+@OutputTimeUnit(TimeUnit.SECONDS)
+@State(Scope.Benchmark)
+@Warmup(iterations = 3, time = 10, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 4, time = 30, timeUnit = TimeUnit.SECONDS)
+public class PBMAC1Benchmark extends JMHBase {
+
+    private SecureRandom random = new SecureRandom();
+    private byte[] salt = new byte[20];
+    private int iterationCount = 300000;
+    private byte[] text = "Bob the builder by IBM".getBytes();
```
KostasTsiounis, line 42: Was any of the code produced by Bob? Only asking cause we should include an acknowledgement.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
```diff
@@ -0,0 +1,556 @@
...
+import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assumptions.assumeFalse;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+@Tag(Tags.OPENJCEPLUS_NAME)
+@Tag(Tags.OPENJCEPLUS_FIPS_NAME)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@ParameterizedClass
+@MethodSource("ibm.jceplus.junit.tests.TestArguments#getEnabledProviders")
+public class TestPBMAC1 extends BaseTest {
```
KostasTsiounis, line 49: Should we add a test case where key and parameters don't match to check these paths as well?

src/test/java/ibm/jceplus/jmh/PBMAC1Benchmark.java
dev-koan: None of the code was produced by Bob
I added the message as a placeholder 😂

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan: I am not sure how to produce a PBEKey for these Mac algorithms (I dont think we can use the PBEKeyFactory for them)

dev-koan: Added a test for it.

jasonkatonica (approved):
LGTM

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
```diff
@@ -475,9 +478,38 @@ private byte[] update(Mac m, byte[] text, int updateLen) throws Exception {
         return finalUpdate;
     }
 
+    @ParameterizedTest
+    @FieldSource("algorithms")
+    void testPBAMCException(String alg) throws Exception {
```
KostasTsiounis: Small typo here. Should be `testPBMACException`.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
```diff
@@ -475,9 +478,38 @@ private byte[] update(Mac m, byte[] text, int updateLen) throws Exception {
...
+        assumeFalse(getProviderName().equalsIgnoreCase("OpenJCEPlusFIPS"));
+
+        secureRandom.nextBytes(salt);
+        Mac mac = Mac.getInstance(alg, getProviderName());
+
+        try {
+            SecretKeyFactory kdf = SecretKeyFactory.getInstance("PBKDF2With" + alg.substring(7), getProviderName());
```
KostasTsiounis: This can be created before the try block and be used for both cases.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
```diff
@@ -475,9 +478,38 @@ private byte[] update(Mac m, byte[] text, int updateLen) throws Exception {
...
+            SecretKey keyDiffIterationCount = (SecretKey) kdf.generateSecret(new PBEKeySpec(PASSWORD, salt, iterationCount + 5, 512)); 
+            mac.init(keyDiffIterationCount, new PBEParameterSpec(salt, iterationCount));
```
KostasTsiounis: Similarly, the "correct" spec can be created and used in both cases.

KostasTsiounis (changes requested):

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan: Updated.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan: Updated.

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
dev-koan: Updated.

JinhangZhang (approved):
LGTM

taoliult (approved):
LGTM

src/test/java/ibm/jceplus/junit/tests/TestPBMAC1.java
KostasTsiounis: I meant the `new PBEParameterSpec(salt, iterationCount)` that is passed as parameter in lines 495 and 505

dev-koan: Updated.

KostasTsiounis (approved):
LGTM

jasonkatonica:
Hi @dev-koan Please squash into a single commit based upon latest `main` branch. I believe that this feature is ready to merge.

PR #1491, opened 2026-05-22
Add OpenJCEPlusFIPS provider to JMH benchmarks

jasonkatonica:
Add OpenJCEPlusFIPS as a benchmark parameter across some JMH
performance tests to enable FIPS-compliant provider testing.
Update JenkinsfilePerformance to set OpenJCEPlusFIPS default
to false to avoid running FIPS benchmarks by default unless
the user requests it.

Updated each JMH test run from 1G initial and maximum heap to 2G.


johnpeck-us-ibm (approved):

KostasTsiounis (approved):
LGTM

PR #1492, opened 2026-05-22
Support OpenSSL backend and implement message digest

KostasTsiounis:
With this change, all appropriate code is added to support the use of `OpenSSL` as a backend. That includes:
- Main Java class to load the libraries required and offer native methods.
- Adapter Java classes to smoothly integrate without changes with existing code.
- Makefiles to compile and link required C code to use `OpenSSL`.
- Additional test tags and general test setup to allow the execution of appropriate tests with the new backend.

Additionally, the first set of algorithms, namely the message digests, are implemented. The aforementioned tags are, also, added to the corresponding tests.


johnpeck-us-ibm:
I know we talked about contexts and one for FIPS and one for non-FIPS. Could we use the Property query instead?
sha256 = EVP_MD_fetch(NULL, "SHA2-256", "fips=yes");  - FIPS
sha256 = EVP_MD_fetch(NULL, "SHA2-256", "provider=default"); - Non-FIPS.
Note the NULL for the context.

Does using this have any performance ramifications?

src/main/java/com/ibm/crypto/plus/provider/base/NativeImplementation.java
```diff
@@ -0,0 +1,48 @@
...
+import sun.security.util.Debug;
+
+public abstract class NativeImplementation {
+    // User enabled debugging
+    private static Debug debug = Debug.getInstance("jceplus");
+
+    @SuppressWarnings("restricted")
+    protected static boolean loadIfExists(File libraryFile) {
+        String libraryName = libraryFile.getAbsolutePath();
+        System.out.println("Library name: " + libraryName);
```
johnpeck-us-ibm: Need to remove this System.out.println and the others or convert them to debugs.

src/main/java/com/ibm/crypto/plus/provider/ock/NativeOCKAdapterNonFIPS.java
```diff
@@ -16,6 +16,7 @@ private NativeOCKAdapterNonFIPS() {
     }
 
     public static NativeOCKAdapterNonFIPS getInstance() {
+        System.out.println("Using OCK non-FIPS adapter.");
```
johnpeck-us-ibm: Same as above.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
```diff
@@ -0,0 +1,1224 @@
...
+        }
+    }
+
+    static public ProviderException providerException(String message, Throwable throwable) {
+        return new ProviderException(message, throwable);
+    }
+
+    @Override
+    public void validateLibraryLocation() throws ProviderException, OpenSSLException {
+        /*if (NativeOpenSSLImplementation.requirePreloadOSSL == false) {
```
johnpeck-us-ibm, line 175: Do we not need this to maybe help us determine where the OpenSSL DLL is coming from?

src/main/native/openssl/Digest.c
```diff
@@ -0,0 +1,446 @@
...
+    int         rc               = 1;
+
+    if (NULL == digestAlgo) {
+        throwOSSLException(env, 0, "DIGEST_create: The specified digest algorithm is null");
+        return 0;
+    }
+
+    if (!(digestAlgoChars = (const char *)(*env)->GetStringUTFChars(env, digestAlgo, NULL))) {
+        throwOSSLException(env, 0, "DIGEST_create: GetStringUTFChars() failed");
+        goto cleanup;
```
johnpeck-us-ibm: This should be a return here instead of goto.

src/main/native/openssl/Digest.c
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    rc = EVP_DigestInit_ex2(mdCtx, md, NULL);
+    if (1 != rc) {
+        throwOSSLException(env, 0, "DIGEST_create: EVP_DigestInit_ex2 failed");
+        goto cleanup;
+    }
+
+    // Everything succeeded. Set digestId to created EVP_MD_CTX.
+    digestId = (jlong)((intptr_t)mdCtx);
```
johnpeck-us-ibm, line 65: Should this be (intptr_t) (intptr_t*) instead?

johnpeck-us-ibm (changes requested):

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
```diff
@@ -0,0 +1,1224 @@
...
+        // made the method synchronizaed to ensure only one thread can execute
+        // the method at a time.
+        //
+        if (osslInitialized) {
+            return;
+        }
+
+        try {
+            //long osslContextId =  NativeOpenSSLImplementation.initializeOSSL(this.useFIPSMode);
+            long osslContextId = 0;
```
thu-ibm, line 64: It appears in this case OpenSSL context is never initialized since the real call is commented out

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapterNonFIPS.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class NativeOpenSSLAdapterNonFIPS extends NativeOpenSSLAdapter {
+    private static NativeOpenSSLAdapterNonFIPS instance = null;
+
+    private NativeOpenSSLAdapterNonFIPS() {
+        super(false);
+    }
+
+    public static NativeOpenSSLAdapterNonFIPS getInstance() {
+        System.out.println("Using OpenSSL non-FIPS adapter.");
```
thu-ibm: Using System.out.println pollutes stdout in production

src/main/java/com/ibm/crypto/plus/provider/openssl/OpenSSLException.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * 
+     */
+    private static final long serialVersionUID = -3104732494450550831L;
+
+    /* These codes are overriding the ones specified in the superclass NativeException
+     * and must match those defined in native/openssl/ExceptionCodes.h.
+     */
+    public static final int GKR_FIPS_MODE_INVALID = 0x00000001;
+    public static final int GKR_OCK_ATTACH_FAILED = 0x00000002;
```
thu-ibm: Error constants like GKR_OCK_ATTACH_FAILED are OCK-specific

src/main/java/com/ibm/crypto/plus/provider/openssl/OpenSSLException.java
```diff
@@ -0,0 +1,63 @@
...
+ * this code, including the "Classpath" Exception described therein.
+ */
+
+package com.ibm.crypto.plus.provider.openssl;
+
+import com.ibm.crypto.plus.provider.base.NativeException;
+import java.util.Hashtable;
+import java.util.Map;
+
+public class OpenSSLException extends NativeException {
```
thu-ibm, line 15: NativeException appears to be an OCK-specific base class

src/main/native/openssl/Utils.c
```diff
@@ -0,0 +1,232 @@
...
+#include <jni.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <openssl/err.h>
+
+#include "Utils.h"
+#include <string.h>
+
+int debug = 0;  // FIXME
```
thu-ibm, line 18: All gslogMessage calls always print to stderr regardless of the flag value

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
KostasTsiounis: We could add a JNI method to do this, but the question is whether we need this given that it's not like OCK that is provided through a binary. In this case, the user provides their own OpenSSL library.

src/main/native/openssl/Digest.c
KostasTsiounis: The type `intptr_t` is supposed to hold pointers itself, so we don't need a `intptr_t *`.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
KostasTsiounis: That is because the OpenSSL context is not yet needed in the use case I'm introducing, given that we decided not to do FIPS at the moment. So, I only left this here for future use and is not needed now.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapterNonFIPS.java
KostasTsiounis: Sorry, those were leftovers from debugging. Removing them.

src/main/java/com/ibm/crypto/plus/provider/openssl/OpenSSLException.java
KostasTsiounis: Removed the 2 that are irrelevant  and kept the two that are used in common parts. Code [here](https://github.com/IBM/OpenJCEPlus/blob/b0629b113384e92a0e7b7ee7debe2fae1168e691/src/main/java/com/ibm/crypto/plus/provider/base/Poly1305Cipher.java#L361) can be reworked in a future PR to not use these constants at all.

src/main/java/com/ibm/crypto/plus/provider/openssl/OpenSSLException.java
KostasTsiounis: I'm not sure what you mean here. `NativeException` is supposed to be a superclass for all exceptions coming from native code. So, native code that uses OCK throws an `OCKException`, whereas the new digest code that uses OpenSSL throws an `OpenSSLException`.

src/main/native/openssl/Utils.c
KostasTsiounis: This is following the same pattern that all OCK native code has. Do you think we should change this to have different levels of logging? We could open an issue to do this everywhere if we want.

src/main/java/com/ibm/crypto/plus/provider/base/NativeImplementation.java
KostasTsiounis: Left from debug. Removed.

src/main/java/com/ibm/crypto/plus/provider/ock/NativeOCKAdapterNonFIPS.java
KostasTsiounis: Left from debug. Removed.

src/main/native/openssl/Digest.c
KostasTsiounis: Fixed.

src/test/java/ibm/jceplus/junit/tests/TestArguments.java
```diff
@@ -63,81 +44,53 @@ protected static Stream<Arguments> getOpenJCEPlusWithSUNInteropProvider() {
...
     }
 
     /**
-     * Resolves enabled OpenJCEPlus* providers from -Dgroups, defaulting to all when none are specified.
+     * Resolves enabled OpenJCEPlus* providers from -Dgroups, defaulting to all specified through tags, if none are specified.
      *
      * @return A stream of enabled TestProvider.
      */
-    protected static Stream<TestProvider> getEnabledProviders() {
+    protected static Stream<TestProvider> getEnabledProviders(Set<String> providers) {
```
taoliult, line 74: This method `getEnabledProviders` changed from `no parameters` to requires a `Set<String> providers` argument. But the test classes that use `@MethodSource` to call this method are not updated to provide that argument (For example, the annotation still like this:  `@MethodSource("ibm.jceplus.junit.tests.TestArguments#getEnabledProviders")`) . Is this correct?

src/main/java/com/ibm/crypto/plus/provider/openssl/OpenSSLException.java
thu-ibm: In NativeException.java, the naming in the following gives me the impression they are used for OCK.

    public static final int GKR_FIPS_MODE_INVALID = -1;
    public static final int GKR_OCK_ATTACH_FAILED = -1;
    public static final int GKR_DECRYPT_FINAL_BAD_PADDING_ERROR = -1;
    public static final int GKR_UNSPECIFIED = -1;

KostasTsiounis: Yeah, I can see why. But I only kept them for the reasons mentioned in https://github.com/IBM/OpenJCEPlus/pull/1492#discussion_r3730854662. We'll remove them in the future.

src/test/java/ibm/jceplus/junit/tests/TestArguments.java
KostasTsiounis: This is part of the https://github.com/IBM/OpenJCEPlus/pull/1625 that this is built on top. But to answer your question, JUnit looks for `ParameterResolver` classes, which are specified through `@ExtendWith` for each test, to figure out the parameters that will be supplied to the method in `@MethodSource`. One cannot specify them directly with the method name.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
johnpeck-us-ibm: Okay I am fine not using this then.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
johnpeck-us-ibm: ahh okay.

johnpeck-us-ibm (approved):

src/main/native/openssl/Digest.c
```diff
@@ -0,0 +1,446 @@
...
+
+    return rc;
+}
+/*
+ * Class:     com_ibm_crypto_plus_provider_openssl_NativeOpenSSLImplementation
+ * Method:    DIGEST_digest_and_reset
+ * Signature: (JJI)V
+ */
+JNIEXPORT void JNICALL
+Java_com_ibm_crypto_plus_provider_openssl_NativeOpenSSLImplementation_DIGEST_1digest_1and_1reset__JJI(
```
JinhangZhang, line 326: According to the parameters, this should be _JJJI

src/main/java/com/ibm/crypto/plus/provider/base/Digest.java
```diff
@@ -43,11 +45,9 @@ public final class Digest implements Cloneable {
     final static int numShaAlgos = 5;
     private static final String DIGEST_CONTEXT_CACHE_SIZE = "com.ibm.crypto.provider.DigestContextCacheSize";
 
-    private static boolean needsInit = true;
+    private static final Map<NativeInterface, ConcurrentLinkedQueueLong[]> contextsPerBackend = new HashMap<>();
 
-    static private ConcurrentLinkedQueueLong contexts[];
-
-    static private int runtimeContextNum[];
+    private static final Map<NativeInterface, Integer[]> runtimeContextNumPerBackend = new HashMap<>();
```
JinhangZhang, line 50: I have a concern here by using two hashmaps. If I understand correctly, the backend cache state is published through two separate maps. Another thread can observe contextsPerBackend initialized before the corresponding runtimeContextNumPerBackend entry is published, skip initialization, and later dereference a null runtimeContextNum.

KostasTsiounis: I'm not sure I follow the train of thought here. Are you talking about the static class variables? These will be assigned to a new hashmap during the class initialization. When a thread actually attempts to get a context, it will go through the double check and synchronization lock in lines 75-78. The context queue per backend is also a concurrent one.

Did you mean something else that I missed here?

src/main/native/openssl/Digest.c
```diff
@@ -0,0 +1,446 @@
...
+        return 0;
+    }
+
+    digestBytesNative = (unsigned char *)((*env)->GetPrimitiveArrayCritical(env, digestBytes, &isCopy));
+    if (NULL == digestBytesNative) {
+        throwOSSLException(env, 0, "DIGEST_digest: GetPrimitiveArrayCritical failed");
+        goto cleanup;
+    }
+
+    rc = EVP_DigestFinal_ex(mdCtx, digestBytesNative, (unsigned int *)&digestLen);
```
JinhangZhang, line 262: Is there any reason why we call EVP_DigestFinal_ex twice? If you want to get the output buffer size, i believe you can do something like

```
int digestLen = EVP_MD_CTX_get_size(mdCtx);

digestBytes = (*env)->NewByteArray(env, digestLen);

unsigned int actualLen = 0;

EVP_DigestFinal_ex(
    mdCtx,
    digestBytesNative,
    &actualLen);
```

src/main/java/com/ibm/crypto/plus/provider/base/Digest.java
JinhangZhang: Yes, I’m referring to the static per-backend maps, but the issue is not the initialization of the maps themselves or the ConcurrentLinkedQueue.

Consider two threads entering getContext() for a backend that has not been initialized yet. Both threads can initially `read contexts == null` and `runtimeContextNum == null`. Thread A acquires the lock first, initializes both entries, and releases the lock. Thread B then acquires the lock and re-reads only contextsPerBackend. Since contexts is now non-null, it skips the initialization block. However, Thread B never re-reads runtimeContextNumPerBackend, so its local runtimeContextNum variable is still null from the read before acquiring the lock.

src/main/native/openssl/Digest.c
```diff
@@ -0,0 +1,446 @@
...
+/*
+ * Class:     com_ibm_crypto_plus_provider_openssl_NativeOpenSSLImplementation
+ * Method:    DIGEST_update
+ * Signature: (J[BII)V
+ */
+JNIEXPORT jint JNICALL
+Java_com_ibm_crypto_plus_provider_openssl_NativeOpenSSLImplementation_DIGEST_1update(
+    JNIEnv *env, jclass thisObj, jlong osslContextId, jlong digestId,
+    jbyteArray data, jint offset, jint dataLen) {
+    //static const char *functionName = "NativeOSSLImplementation.DIGEST_update";
```
jasonkatonica, line 167: Comment can be removed.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
jasonkatonica: Could we add a comment describing this? Maybe a `TODO` ? Also appears to be commented lines here i assume you will need those if / when we do FIPS?

src/main/native/openssl/Digest.c
```diff
@@ -0,0 +1,446 @@
...
+
+    if (NULL == digestBytesNative) {
+        throwOSSLException(env, 0, "DIGEST_digest_and_reset_internal: The pointer to the specified data array is null");
+        return 0;
+    }
+
+    if (digestLen < 0) {
+        throwOSSLException(env, 0, "DIGEST_digest_and_reset_internal: The specified data length is negative");
+        return 0;
+    }
```
jasonkatonica, line 302: Perhaps we can skip this check since a `unsigned int` could not be `< 0`

src/main/native/openssl/Utils.c
```diff
@@ -0,0 +1,232 @@
...
+    return gslogMessage("Exiting %s", functionName);
+}
+
+//============================================================================
+//
+//
+void osslCheckStatus(void) {
+    unsigned long errCode;
+
+    while ((errCode = ERR_get_error()) == 1) {
```
jasonkatonica, line 120: Are we sure all errors return as a value of `1` ? The OpenSSL doc for this API [ERR_get_error](https://docs.openssl.org/3.6/man3/ERR_get_error/#description) indicates a value of 0 if there are no errors left to read from the queue.
```
while ((errCode = ERR_get_error()) != 0)
```

src/main/native/openssl/Utils.c
```diff
@@ -0,0 +1,232 @@
...
+//============================================================================
+//
+//
+int gslogError(const char *formatString, ...) {
+    int         charsPrinted;
+    va_list     formatArgs;
+    static char printBuffer[4096];
+
+    va_start(formatArgs, formatString);
+    charsPrinted = vsprintf(printBuffer, formatString, formatArgs);
```
jasonkatonica, line 36: Could we use `vsnprintf` instead of `vsprintf` in all our new openssl code? If `formatString` were large here or was missing a null terminator I think it would overlay `printBuffer`. In general all our new code we should try to make as bounded as possible. Something like the following in various places:

```
vsnprintf(printBuffer, sizeof(printBuffer), formatString, formatArgs);
```

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
```diff
@@ -0,0 +1,1224 @@
...
+import java.nio.ByteBuffer;
+import java.security.ProviderException;
+import sun.security.util.Debug;
+
+public abstract class NativeOpenSSLAdapter implements NativeInterface {
+    // These code values must match those defined in Context.h.
+    //
+    private static final int VALUE_ID_FIPS_APPROVED_MODE = 0;
+    private static final int VALUE_OCK_INSTALL_PATH = 1;
+    private static final int VALUE_OCK_VERSION = 2;
```
jasonkatonica, line 21: Seems out of place to have OCK named variable here? These variables do seem like they are in use so just renaming them would work.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
```diff
@@ -0,0 +1,1224 @@
...
+
+                if (debug != null) {
+                    exceptionMessage = "Dependent library was loaded from " + ockInstallPath;
+                }
+
+                throw new ProviderException(exceptionMessage);
+            }
+        } catch (java.io.IOException e) {
+            throw new ProviderException("Failed to validate dependent library", e);
+        }*/
```
jasonkatonica, line 204: Do we need this commented out code? It has references to OCK and is not functional so perhaps could be replaced by a simple comment to indicate its not in use at the moment. Same comment below for `getExpectedLibraryVersion`

src/test/OpenSSLOnly.config
```diff
@@ -0,0 +1,65 @@
...
+# Copyright IBM Corp. 2026
+#
+# This code is free software; you can redistribute it and/or modify it
+# under the terms provided by IBM in the LICENSE file that accompanied
+# this code, including the "Classpath" Exception described therein.
+
+
+name = OpenSSL
+
+description = OpenJCEPlus-OpenSSL Provider
```
jasonkatonica, line 11: In our `Tags.java` file we call the test provider `OpenJCEPlus_OpenSSL` however in this location we call it `OpenJCEPlus-OpenSSL`. I think we should choose one or the other. I prefer `-` character but either way can work.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
```diff
@@ -0,0 +1,1224 @@
...
+        return NativeOpenSSLImplementation.CIPHER_encryptFinal(osslContext.getId(), cipherId,
+            input, inOffset, inLen, ciphertext, ciphertextOffset, needsReinit);
+    }
+
+    @Override
+    public int CIPHER_decryptFinal(long cipherId, byte[] ciphertext, int cipherOffset, int cipherLen,
+            byte[] plaintext, int plaintextOffset, boolean needsReinit) throws OpenSSLException {
+        return NativeOpenSSLImplementation.CIPHER_decryptFinal(osslContext.getId(), cipherId,
+            ciphertext, cipherOffset, cipherLen, plaintext, plaintextOffset, needsReinit);
+    }
```
jasonkatonica, line 387: Most of these methods would fail since they are not implemented in our OpenSSL native code yet. Should we be throwing exceptions for all of these such that it is obvious what is and is not implemented? Also if one were to make a call to most of these methods I would assume a `UnsatisifedLink` / crash error would occur. It might be cleaner to throw exceptions here so its obvious what is and is not complete in the backend code. Maybe `UnsupportedOperationException` is appropriate given its a `RuntimeException` in all the unsupported methods:
```
throw new UnsupportedOperationException
```

src/main/java/com/ibm/crypto/plus/provider/base/Digest.java
jasonkatonica: Maybe `ConcurrentHashMap` instead of `Map` ?

src/main/native/openssl/Digest.c
```diff
@@ -0,0 +1,446 @@
...
+/*
+ * Class:     com_ibm_crypto_plus_provider_openssl_NativeOpenSSLImplementation
+ * Method:    DIGEST_digest_and_reset
+ * Signature: (JJI)V
+ */
+JNIEXPORT void JNICALL
+Java_com_ibm_crypto_plus_provider_openssl_NativeOpenSSLImplementation_DIGEST_1digest_1and_1reset__JJI(
+    JNIEnv *env, jclass thisObj, jlong osslContextId, jlong digestId,
+    jlong digestBytes, jint length) {
+    //static const char *functionName = "NativeOSSLImplementation.DIGEST_digest_and_reset";
```
jasonkatonica, line 329: Any reason to keep these method names through the code? If we are going to use them then we should add debugging pathing, if not i think we could eliminate these. We could always later copy the value from the comment for the method for the `Method:` name such that the value would not be lost.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
```diff
@@ -0,0 +1,1224 @@
...
+            }
+        }*/
+
+        return null;
+    }
+
+    @Override
+    public String getLibraryBuildDate() {
+        if (libraryBuildDate == unobtainedValue) {
+            libraryBuildDate = NativeOpenSSLImplementation.getLibraryBuildDate();;
```
jasonkatonica, line 265: Two `;;` characters.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
```diff
@@ -0,0 +1,1224 @@
...
+    private boolean useFIPSMode;
+
+    private String osslVersion = unobtainedValue;
+    private String osslInstallPath = unobtainedValue;
+
+    // The following is a special String instance to indicate that a
+    // value has not yet been obtained.  We do this because some values
+    // may be null and we only want to query the value one time.
+    //
+    private static String libraryBuildDate = unobtainedValue;
```
jasonkatonica, line 42: Does this have to be `static` ? I could see how there are two threads trying to set `libraryBuildDate` below might cause some issues?

src/main/native/openssl/Digest.c
```diff
@@ -0,0 +1,446 @@
...
+     * ctx->fetched_digest inside mdCtx.  EVP_MD_CTX_free will release that
+     * reference.  We must always release our own fetch reference here,
+     * regardless of success or failure, to avoid a refcount leak.
+     * On the error path mdCtx is freed below (which drops its reference),
+     * then we drop ours.  If mdCtx was never initialised we just drop ours.
+     */
+    EVP_MD_free(md);
+    md = NULL;
+
+    (*env)->ReleaseStringUTFChars(env, digestAlgo, digestAlgoChars);
```
jasonkatonica, line 79: We might need a null check on `digestAlgoChars` here? If an exception is thrown prior to line 41 then `digestAlgoChars` would be null.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapterNonFIPS.java
```diff
@@ -0,0 +1,25 @@
...
+    private static NativeOpenSSLAdapterNonFIPS instance = null;
+
+    private NativeOpenSSLAdapterNonFIPS() {
+        super(false);
+    }
+
+    public static NativeOpenSSLAdapterNonFIPS getInstance() {
+        if (instance == null) {
+            instance = new NativeOpenSSLAdapterNonFIPS();
+        }
```
JinhangZhang, line 21: Is it guaranteed that NativeOpenSSLAdapterNonFIPS.getInstance() returns a single instance under concurrent first access? getInstance() itself is not synchronized and its instance field is not volatile, so two threads may construct different adapter instances.

This seems particularly relevant now that Digest uses the NativeInterface instance itself as the key for contextsPerBackend. Two adapter instances representing the same OpenSSL backend would therefore create separate digest caches.

src/main/native/openssl/Digest.c
KostasTsiounis: If an exception is thrown that early, we don't go to `cleanup`. We just `return 0`.

src/main/native/openssl/Digest.c
KostasTsiounis: Updated comment and method signature.

src/main/java/com/ibm/crypto/plus/provider/base/Digest.java
KostasTsiounis: I make sure to set `runtimeContextNum` in the synchronized block if `contexts` has already been set.

src/main/native/openssl/Digest.c
KostasTsiounis: Changed to suggested API. It looks like `EVP_DigestFinal_ex` doesn't guarantee proper size.

src/main/native/openssl/Digest.c
KostasTsiounis: Removed.

src/main/native/openssl/Digest.c
KostasTsiounis: Removed.

src/main/native/openssl/Utils.c
KostasTsiounis: Changed to suggestion.

src/main/native/openssl/Utils.c
KostasTsiounis: Switched to `vsnprintf`.

src/test/OpenSSLOnly.config
KostasTsiounis: `OpenJCEPlus_OpenSSL` is only used as the string in the `@Tag` annotation due to restrictions. Everything else, including `TestProvider` definition and name of provider registered, uses `OpenJCEPlus-OpenSSL`.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
KostasTsiounis: Changed all unimplemented ones to throw exception.

src/main/native/openssl/Digest.c
KostasTsiounis: Followed the same pattern as OCK initially, but I removed them now.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
KostasTsiounis: They would just set it to the same value. I don't think there is a need to keep one per instance.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapterNonFIPS.java
KostasTsiounis: You are right. Changed to do a double check with synchronization in between.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
KostasTsiounis: Renamed.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
KostasTsiounis: Changed them entirely to use OpenSSL methods to get and somewhat check version and location.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
KostasTsiounis: Fixed.

thu-ibm (approved):

KostasTsiounis:
I added some additional guards for `z/OS`, since OpenSSL is not available there with e6a77cb21b9278eba38e18a0241adf5f0b13caed

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
```diff
@@ -0,0 +1,1347 @@
...
+            }
+        } catch (java.io.IOException e) {
+            throw new ProviderException("Incorrect file specification for dependent library", e);
+        }
+    }
+
+    @Override
+    public void validateLibraryVersion() throws ProviderException, OpenSSLException {
+        String[] expectedVersion = getExpectedLibraryVersion().split("\\.");
+        String[] actualVersion = getLibraryVersion().split("\\.");
```
jasonkatonica, line 201: Can we debug the library version eitehr in this location or in the helper method `getLibraryVersion`. This would be handy to have at times.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/base/Digest.java
JinhangZhang: I still have one concern here. Consider a backend that has not been initialized yet, then there are two threads.

Thread A enters the synchronized block and creates both `contexts` and `runtimeContextNum`. Then, thread A executes:

```
contextsPerBackend.put(this.nativeInterface, contexts);
```

but has not yet executed:

```
runtimeContextNumPerBackend.put(this.nativeInterface, runtimeContextNum);
```

At this point, thread B enters `getContext()` and performs the initial reads outside the synchronized block:

```
contexts = contextsPerBackend.get(this.nativeInterface);
runtimeContextNum = runtimeContextNumPerBackend.get(this.nativeInterface);
```

Thread B can observe:

```
contexts != null
runtimeContextNum == null
```

Since `contexts != null`, Thread B skips the synchronized block entirely, so the new re-read of `runtimeContextNumPerBackend` inside the lock is never reached.

I think either the two backend states need to be published atomically as one object, or both the get and put need to use the same synchronization.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
KostasTsiounis: Added.

src/main/java/com/ibm/crypto/plus/provider/base/Digest.java
KostasTsiounis: I updated to check both variables and wait for the `synchronized` block if either of them is not set. That way I think we can guarantee that no thread will move on without having values for both. See d3e59addfbc83eb9f31dd67431a358cff27001ac.

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
```diff
@@ -0,0 +1,1352 @@
...
+                    + ", got " + actualVersion);
+            } else if ((minorExpected == minorActual) && (patchExpected > patchActual)) {
+                throw new ProviderException("Expected OpenSSL library version greater than " + expectedVersion
+                        + ", got " + actualVersion);
+            }
+        }
+    }
+
+    private String getExpectedLibraryVersion() {
+        return "3.0.0";
```
jasonkatonica: Can we bump this to the latest LTS release that is currently in service for openssl? Also please define this as a static final constant.

Image

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/openssl/NativeOpenSSLAdapter.java
KostasTsiounis: Changed to a constant and hardcoded "3.5.0" as the minimum allowed version.

taoliult (approved):
LGTM

jasonkatonica (approved):
LGTM

JinhangZhang (approved):
LGTM

PR #1493, opened 2026-05-24
Set XEC secret buffer size explicitly on z/OS

taoliult:
Set the expected XEC secret buffer size explicitly on z/OS before calling the native ECDH secret computation.

On other platforms, the native layer can query the required secret size by calling ICC_EVP_PKEY_derive. However, this size query does not work on z/OS.

To handle this platform difference, determine the curve name from the public key parameters on z/OS and pass the expected secret buffer size to the native layer.

Other platforms continue to use the ICC_EVP_PKEY_derive size query.

### DIFF
--- a/src/main/native/ock/ECKey.c
+++ b/src/main/native/ock/ECKey.c
@@ -2227,7 +2227,7 @@ Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_ECKEY_1computeECDH
 JNIEXPORT jbyteArray JNICALL
 Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_XECKEY_1computeECDHSecret(
     JNIEnv *env, jclass thisObj, jlong ockContextId, jlong genCtx,
-    jlong pubXecKeyId, jlong privXecKeyId) {
+    jlong pubXecKeyId, jlong privXecKeyId, jint secretBufferSize) {
     static const char *functionName =
         "NativeInterface_XECKEY_1computeECDHSecret";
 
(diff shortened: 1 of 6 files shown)


jasonkatonica (approved):
LGTM

src/main/native/ock/ECKey.c
```diff
@@ -2266,22 +2266,32 @@ Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_XECKEY_1computeECD
...
-        throwOCKException(env, 0, "Derived secret size is zero");
-        goto cleanup;
-    }
+    if (secretBufferSize > 0) {
+        /*
+         * On z/OS, querying the required secret size may not work properly.
+         * In that case, use the caller-provided secretBufferSize as the
+         * expected derived secret size.
+         */
+        required_len = (size_t)secretBufferSize;
```
KostasTsiounis: Do we need a space after the cast?

src/main/native/ock/ECKey.c
```diff
@@ -2266,22 +2266,32 @@ Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_XECKEY_1computeECD
...
+        secret_key_len = required_len;
```
KostasTsiounis: I think this and line 2293 can be consolidated into one after the if-else block.

KostasTsiounis (changes requested):

src/main/native/ock/ECKey.c
taoliult: Yes, updated.

src/main/native/ock/ECKey.c
taoliult: Yes, updated.

KostasTsiounis (approved):
LGTM

johnpeck-us-ibm (approved):

src/main/native/ock/ECKey.c
```diff
@@ -2227,7 +2227,7 @@ Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_ECKEY_1computeECDH
 JNIEXPORT jbyteArray JNICALL
 Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_XECKEY_1computeECDHSecret(
     JNIEnv *env, jclass thisObj, jlong ockContextId, jlong genCtx,
-    jlong pubXecKeyId, jlong privXecKeyId) {
+    jlong pubXecKeyId, jlong privXecKeyId, jint secretBufferSize) {
```
keithc-ca, line 2230: Line 2225 should be updated to reflect the new signature.

taoliult: I will open another PR to correct the signature.

PR #1499, opened 2026-06-01
Check that integer representation of message is smaller than modulus

KostasTsiounis:
In RSA cipher operations, the message is converted into an integer that needs to be smaller than the public key's modulus.

A check is added to fail early with the proper message, instead of failing during the actual operation with a not so helpful message.

An additional test to verify that behaviour is introduced as well.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1252


johnpeck-us-ibm (approved):

taoliult (approved):
LGTM

jasonkatonica (approved):
LGTM

PR #1508, opened 2026-06-03
Use ThreadLocal caching for ExtendedRandom PRNG contexts

taoliult:
The first two commits have already been reviewed. However, due to ICC_RNG_Generate() failures in the GitHub Actions build, I reverted the first two commits and added a third commit to fix the ICC_RNG_Generate() failures.

This PR includes the following changes:

Add ThreadLocal caching for native PRNG contexts used by ExtendedRandom. Each thread creates and reuses a PRNG context for supported DRBG algorithms, avoiding repeated EXTRAND_create calls when ExtendedRandom instances are created frequently.

On z/OS, ThreadLocal caching is disabled because it can cause native TCB context issues. z/OS now uses instance-specific PRNG contexts, while other platforms continue using ThreadLocal caching for performance.

Also fix ThreadLocal PRNG context ownership by resolving the current thread’s PRNGContextPointer in nextBytes() instead of storing ThreadLocal-owned native context IDs in ockPRNGContextId, which is now used only for instance-owned contexts.

This is a back port PR from PR: https://github.com/IBM/OpenJCEPlus/pull/1549

### DIFF
--- a/JenkinsfilePerformance
+++ b/JenkinsfilePerformance
@@ -385,6 +385,7 @@ pipeline {
                                             'ibm.jceplus.jmh.PBEBenchmark', \
                                             'ibm.jceplus.jmh.PBKDF2Benchmark', \
                                             'ibm.jceplus.jmh.RandomBenchmark', \
+                                            'ibm.jceplus.jmh.RandomNewInstanceBenchmark', \
                                             'ibm.jceplus.jmh.RSACipherBenchmark', \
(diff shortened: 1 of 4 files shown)


KostasTsiounis (approved):
LGTM

jasonkatonica (approved):
LGTM

johnpeck-us-ibm (approved):

PR #1512, opened 2026-06-03
Add Support to Select native Crypto Provider

johnpeck-us-ibm:
Code was added to allow the use of  a Provider.Service attrobute called "NativeProvider" to specify the native back end to use. Only OCK and OPENSSL are support. Any thing specified in the "NativeProvider" attribute other than OPENSSL will result in OCK being used.


src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,150 @@
...
+    
+    /**
+     * Sets the OpenSSL backend implementation.
+     * This should be called during system initialization.
+     * 
+     * @param backend the OpenSSL backend implementation
+     */
+    public static void setOpenSSLBackend(NativeInterface backend) {
+        opensslBackend = backend;
+    }
```
KostasTsiounis: Do we want to allow external setting of the backends? I would think we set them to what we want.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Gets the backend implementation for the specified backend type.
+     * 
+     * @param backend the backend type
+     * @return the backend implementation, or null if not set
+     */
+    public static NativeInterface getBackend(Backend backend, boolean isFIPS) {
+        if (backend == Backend.OCK) {
+            if (isFIPS) {
+                ockBackendFIPS = Objects.requireNonNullElseGet(ockBackendFIPS, () -> NativeOCKAdapterFIPS.getInstance());
```
KostasTsiounis: The `NativeOCKAdapterFIPS` is a singleton, so `ockBackendFIPS = NativeOCKAdapterFIPS.getInstance();` should suffice.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,150 @@
...
+                return ockBackendFIPS;
+            } else {
+                ockBackend = Objects.requireNonNullElseGet(ockBackend, () -> NativeOCKAdapterNonFIPS.getInstance());
+                return ockBackend;
+            }
+        } else if (backend == Backend.OPENSSL) {
+            return opensslBackend;
```
KostasTsiounis, line 67: Should we check for FIPS here and fail if OpenSSL and FIPS are set together?

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,150 @@
...
+        if (provider != null && type != null && algorithm != null) {
+            
+            if (debug != null) {
+                debug.println("Service - " + type + "\nAlg - " + algorithm);
+            }
+
+            service = provider.getService(type, algorithm);
+
+            if (service != null) {
+                // Service not found, default to OCK
```
KostasTsiounis: Do we want to silently default to OCK, or throw an exception, given that this shouldn't happen?

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,150 @@
...
+            // Default to OCK when no value or blank
+            return Backend.OCK;
+        }
+        
+        String normalized = nativeProviderValue.trim().toUpperCase();
+        
+        if ("OPENSSL".equals(normalized)) {
+            return Backend.OPENSSL;
+        } else {
+            // Default to OCK for any other value including "OCK"
```
KostasTsiounis: Similar question here. Do we silently default to OCK regardless of what they put in the attribute? Potential typos would cause problems I think.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,150 @@
...
+
+import com.ibm.crypto.plus.provider.OpenJCEPlusProvider;
+import com.ibm.crypto.plus.provider.ock.NativeOCKAdapterFIPS;
+import com.ibm.crypto.plus.provider.ock.NativeOCKAdapterNonFIPS;
+import java.security.Provider;
+import java.util.Objects;
+import sun.security.util.Debug;
+
+/**
+ * BackendCryptoSelector manages the selection and initialization of cryptographic backends.
```
jasonkatonica: This doc should read `NativeCryptoSelector` instead of `BackendCryptoSelector` to match the classname.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,150 @@
...
+
+    // Backend implementation instances (will be set by concrete implementations)
+    private static volatile NativeInterface ockBackend = null;
+    private static volatile NativeInterface opensslBackend = null;
+    private static volatile NativeInterface ockBackendFIPS = null;
+    private static volatile NativeInterface opensslBackendFIPS = null;
+    
+    // Locks for thread-safe initialization
+    private static final Object ockLock = new Object();
+    private static final Object opensslLock = new Object();
```
jasonkatonica: What are these locks used for. I dont see them being used below so perhaps they could be removed?

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,150 @@
...
+                ockBackendFIPS = Objects.requireNonNullElseGet(ockBackendFIPS, () -> NativeOCKAdapterFIPS.getInstance());
+                return ockBackendFIPS;
+            } else {
+                ockBackend = Objects.requireNonNullElseGet(ockBackend, () -> NativeOCKAdapterNonFIPS.getInstance());
+                return ockBackend;
+            }
+        } else if (backend == Backend.OPENSSL) {
+            return opensslBackend;
+        }
+        return null;
```
jasonkatonica, line 69: Can we throw an exception here instead of returning `null`? We dont expect this to be run so might be more defensive that it would fail with an obvious error.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
jasonkatonica: I do like the idea that algorithms must be explictly set to a backend as it would be hard to notice defaulting behavior otherwise. For example in the case here where `provider.getService` returns null on the lookup for one reason or another.

src/main/java/com/ibm/crypto/plus/provider/base/CCMCipher.java
```diff
@@ -193,7 +192,7 @@ public static int doCCMFinal_Decrypt(byte[] key, byte[] iv, int tagLen,
 
         int aadLen = authenticationData.length;
 
-        NativeInterface nativeInterface = provider.isFIPS() ? NativeOCKAdapterFIPS.getInstance() : NativeOCKAdapterNonFIPS.getInstance();
+        NativeInterface nativeInterface = NativeCryptoSelector.selectBackend(provider, "Cipher", "AES/CCM/NoPadding");
```
jasonkatonica, line 195: Do we think its worth introducing a enumeration or string constant someplace in our code base to define the repeat strings we use for the type and algorithm? This would save some allocations of string objects while also makeing it less likely we will have a allocation problem with a misspelled string for algorithm name or type.

pom.xml
```diff
@@ -261,6 +261,7 @@
                       --add-opens=openjceplus/ibm.jceplus.junit.base=ALL-UNNAMED
                       --add-opens=openjceplus/ibm.jceplus.junit.tests=ALL-UNNAMED
                       --patch-module openjceplus="target${file.separator}classes${path.separator}target${file.separator}test-classes"
+                      -Djava.security.auth.debug=${java.security.debug}
```
jasonkatonica: I dont think we would want to leave this in permanently to enable debug in all our tests? Line 236 has this already in commented out form.

src/test/ProviderDefAttrs.config
```diff
@@ -303,11 +303,11 @@ Service.KeyGenerator.SunTls12RsaPremasterSecret = com.ibm.crypto.plus.provider.T
 KeyGenerator.SunTlsMasterSecret.alias.add = TlsMasterSecret, TlsExtendedMasterSecret, SunTlsExtendedMasterSecret
 Service.KeyGenerator.SunTlsMasterSecret = com.ibm.crypto.plus.provider.TlsMasterSecretGenerator
 
-KeyGenerator.SunTlsKeyMaterial.alias.add = Tls12MasterSecret
-Service.KeyGenerator.SunTlsKeyMaterial = com.ibm.crypto.plus.provider.TlsMasterSecretGenerator
+KeyGenerator.SunTls12MasterSecret.alias.add = Tls12MasterSecret
```
jasonkatonica, line 306: Was any issues noticed for this change? Seems the registration was incorrect in the past but wasnt sure if you noticed any issues from this or was found during general review?

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Not sure what you are asking? Are you asking for some why to set the backend or create a backend using environment variables?

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: The code either creates the backend or returns the current value. I could have used if's but this was cleaner.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: The code here needs to be updated once we have the OpenSSL native back end in the code. So, I suggest doing that change at that point. which would be in your code.

johnpeck-us-ibm: This was there for me to be able to make sure the selector code was working.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: This is probably the safest thing to do. However, we can make it so that Only OpenSSL or OCKC can be specified. But we still need to default to OCKC if no value is specified. Does that sound okay?

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Fixed

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: They are there if we needed them. I will remove and we can add them back if needed in the future.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: We need to update this code when we have OpenSSL code supported. I suggest doing that than. But I am not sure the null can actually be hit even than,

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
KostasTsiounis: Yes, that makes sense. If no value is specified, we default to OCK. But if the user specifies one, we check it's one or the other to avoid someone making a typo or specifying something that doesn't exist.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
KostasTsiounis: I expect that we will set the backends directly based on the adapters we create. Not sure why we need setters in this class. To my understanding, we would hardcode `ockBackend = NativeOCKAdaptersNonFIPS.getInstance()` etc.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
KostasTsiounis: Doing the `getInstance()` does that too. It doesn't create a new instance every time. The class has a static instance variable and only creates it once.

src/main/java/com/ibm/crypto/plus/provider/base/CCMCipher.java
johnpeck-us-ibm: We could duplicate what is here - https://github.com/openjdk/jdk26u/blob/master/src/java.base/share/classes/java/security/Provider.java#L1411 for the Type. The Algorithm we could create it. I think this would only be useful if we can figure out a way to  use these constants in the config files. That way it is easy to match the two. If you want this it probably makes sense to do that in a different issue.

pom.xml
johnpeck-us-ibm: It doesn't do anything unless you specify java.security.debug on the mvn command. Because is it null and null does not enable debug

src/test/ProviderDefAttrs.config
johnpeck-us-ibm: What is wierd is that everything worked fine and than I add some code I had test failures. So, yes this should have been like this since the being.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Okay. I will add  exceptions here for all the negative cases.  But this has a ripple effect to the other classes to catch this exception. Working on this.

KostasTsiounis: Should they catch it? Or should we let it fail and avoid loading until the user fixes that? Unless you mean that the other classes should expect it in the `throws` part.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: I do not see a need for them today. But they maybe helpful for testing or such in the future. I can remove them if you want and we can add them back in later if necessary.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: There are places in the rest of the code we need to expect this exception. The latest code should have that done now.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Removed.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Changed

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Done

src/main/java/com/ibm/crypto/plus/provider/base/CCMCipher.java
jasonkatonica: After talking some we decided that for now we will make use of enumerations and strings in a different way in the future. For now we decided to stick with the string approach as is already in the code base.

pom.xml
jasonkatonica: OK in this case can we make use of the same property name `java.security.auth.debug` since we are familiar with the value instead of a new value ( `java.security.debug` ) ? This line would become :
```
-Djava.security.auth.debug=${java.security.auth.debug}
```
Additionally we should remove the commented out value just above this change on line 236 above. No need to have that commented value there with this new approach.

Also we should add this same logic to the other test profile in the pom.xml file. We can add the same line near current line 403 ( comment associated with debug can also be deleted from this test profile )

johnpeck-us-ibm: Updated.

jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,133 @@
...
+        
+        String normalized = nativeProviderValue.trim().toUpperCase();
+        
+        switch (normalized) {
+            case "OPENSSL":
+                return Backend.OPENSSL;
+            case "OCK":
+                return Backend.OCK;
+            default:
+                // If not OCK or OpenSSL throw and exception
```
KostasTsiounis: Small typo. Throw `an` instead of `and`.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,133 @@
...
+    private static volatile NativeInterface ockBackendFIPS = null;
+    private static volatile NativeInterface opensslBackendFIPS = null;
+    
+    /**
+     * Gets the backend implementation for the specified backend type.
+     * 
+     * @param backend the backend type
+     * @return the backend implementation, or null if not set
+     */
+    public static NativeInterface getBackend(Backend backend, boolean isFIPS) {
```
KostasTsiounis: Should this be `private`?

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,133 @@
...
+     * Determines which backend to use by querying the Provider service attribute.
+     * Retrieves the NativeProvider attribute from provider.getService(type, algorithm).
+     * 
+     * @param provider the security provider
+     * @param type the service type (e.g., "Cipher", "MessageDigest")
+     * @param algorithm the algorithm name (e.g., "AES", "SHA-256")
+     * @return the Backend to use
+     */
+    public static NativeInterface selectBackend(OpenJCEPlusProvider provider, String type, String algorithm) throws 
+        NativeException {
```
KostasTsiounis: Since we use the `NativeException` class to indicate exceptions thrown from our native code, should we create another exception (like `ConfigurationException`) for this case?

src/main/java/com/ibm/crypto/plus/provider/base/CCMCipher.java
```diff
@@ -490,7 +489,7 @@ static int useHardwareCCM(boolean isEncrypt, int inputLen, int ivLen, int keyLen
         putLongtoByteArray(inputLen * 8, addedParams, TPCLOffset); // Add TPCL
         parameters.put(paramBlockOffset, addedParams, 0, addedParams.length);
 
-        NativeInterface nativeInterface = NativeOCKAdapterNonFIPS.getInstance();
+        NativeInterface nativeInterface = NativeOCKAdapterNonFIPS.getInstance(); // Need to look at this later <------------
```
KostasTsiounis, line 492: Shouldn't this follow the same pattern?

src/main/java/com/ibm/crypto/plus/provider/base/HMAC.java
```diff
@@ -48,7 +46,13 @@ static void throwNativeException(int errorCode) throws NativeException {
     private HMAC(String digestAlgo, OpenJCEPlusProvider provider) throws NativeException {
         //final String methodName = "HMAC (String)";
         this.provider = provider;
-        this.nativeInterface = provider.isFIPS() ? NativeOCKAdapterFIPS.getInstance() : NativeOCKAdapterNonFIPS.getInstance();
+        String algo = "Hmac" + digestAlgo;
+        try {
+            this.nativeInterface = NativeCryptoSelector.selectBackend(provider, "Mac", algo);
+        } catch (Exception e) {
```
KostasTsiounis: What other type of exception can this throw?

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Yes. There is no need for someone to call this. This is used in the public method. We can change this later if there is really a need. But I do not see it today.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Fixed

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: We could, but there two issue type that I see. One where the config is not correct. Once, where the initialization of the back end fails. Do we want to be more specific with the exception or is the exception text enough?  Today all the classes in base use NativeException so keeping the same exception probably makes more sense. If we had more places in base to use that exception maybe it would make sense. I could look at creating a new class -ConfigurationException in the provider directory and look at using that in the ProviderServiceReader class and than make use of that class for the configuration errors? Do you think I should look at that?

src/main/java/com/ibm/crypto/plus/provider/base/CCMCipher.java
johnpeck-us-ibm: The issue here is that this area is the hardware support. We won't have this for OpenSSL and FIPS does not have this either. So, this is really a OCK non-FIPS thing. We need to take this an issue and see what we can do with the hardware support. Maybe this code needs to me move down into the OCK backend and out of base. That would resolve this issue. I will open a specific issue for this work and can address it there. I think there are other places too where we need to look at the hardware stuff. Here is the issue - https://github.com/IBM/OpenJCEPlus/issues/1555

src/main/java/com/ibm/crypto/plus/provider/base/HMAC.java
johnpeck-us-ibm: None. I will update this. Not sure why it is Exception. So removed the try catch.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
KostasTsiounis: Are we switching this to `private`?

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
KostasTsiounis: Could we have `ConfigurationException` extend `RuntimeException` so that we don't have to add it in the throws part of all the method? If not, we can keep the `NativeException`.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Gets the backend implementation for the specified backend type.
+     * 
+     * @param backend the backend type
+     * @return the backend implementation, or null if not set
+     */
+    public static NativeInterface getBackend(Backend backend, boolean isFIPS) {
+        if (backend == Backend.OCK) {
+            if (isFIPS) {
+                ockBackendFIPS = NativeOCKAdapterFIPS.getInstance();
```
KostasTsiounis: Why set it every time if we're not gonna check if it's already set?

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
```diff
@@ -0,0 +1,133 @@
...
+                return ockBackendFIPS;
+            } else {
+                ockBackend = NativeOCKAdapterNonFIPS.getInstance();
```
KostasTsiounis: Same here.

src/main/java/com/ibm/crypto/plus/provider/base/NativeInterface.java
```diff
@@ -330,9 +329,9 @@ public byte[] RSAKEY_getPrivateKeyBytes(long rsaKeyId)
     public byte[] RSAKEY_getPublicKeyBytes(long rsaKeyId)
             throws NativeException;
 
-    public int RSAKEY_size(long rsaKeyId);
+    int RSAKEY_size(long rsaKeyId);
 
-    public void RSAKEY_delete(long rsaKeyId);
+    void RSAKEY_delete(long rsaKeyId);
```
KostasTsiounis: Why change the visibility in these methods?

src/main/java/com/ibm/crypto/plus/provider/base/NativeInterface.java
```diff
@@ -429,7 +428,7 @@ public int RSAPSS_signInit(long rsaPssId, long pkeyId,
     public int RSAPSS_verifyInit(long rsaPssId, long pkeyId,
             int saltlen) throws NativeException;
 
-    public int RSAPSS_getSigLen(long rsaPssId);
+    int RSAPSS_getSigLen(long rsaPssId);
```
KostasTsiounis: Same here.

src/main/java/com/ibm/crypto/plus/provider/base/Signature.java
```diff
@@ -23,16 +21,16 @@ public final class Signature {
...
     }
 
 
-    private Signature(String digestAlgo, OpenJCEPlusProvider provider) throws NativeException {
+    private Signature(String digestAlgo, OpenJCEPlusProvider provider, String algName) throws NativeException {
         //final String methodName = "Signature(String)";
-        this.nativeInterface = provider.isFIPS() ? NativeOCKAdapterFIPS.getInstance() : NativeOCKAdapterNonFIPS.getInstance();
-        this.digest = Digest.getInstance(digestAlgo, provider);
+        this.nativeInterface = NativeCryptoSelector.selectBackend(provider, "Signature", algName);
+        this.digest = Digest.getInstance(digestAlgo, provider, "Signature", algName);
```
KostasTsiounis, line 33: Do we want the digest to be coming from the same backend as signature, or leave it to what is set in the configuration?

src/main/java/com/ibm/crypto/plus/provider/MessageDigest.java
```diff
@@ -120,74 +129,78 @@ protected void engineReset() {
...
     }
 
     public static final class SHA1 extends MessageDigest {
         public SHA1(OpenJCEPlusProvider provider) {
-            super(provider, "SHA1"); // OCK digest name
+            super(provider, "SHA1", "SHA1"); // OCK digest name
+        }
+
+        public SHA1(OpenJCEPlusProvider provider, String ConfigType, String configAlgName) {
+            super(provider, "SHA1", ConfigType, configAlgName); // OCK digest name
```
KostasTsiounis, line 142: What is this method for?

src/main/java/com/ibm/crypto/plus/provider/PBES1Core.java
```diff
@@ -250,7 +252,7 @@ private byte[] passwordBigEndian(byte[] password) {
 
     private byte[] deriveKey(int n, int type, byte[] pass) {
         byte[] res = new byte[n];
-        MessageDigest sha = new MessageDigest.SHA1(provider);
+        MessageDigest sha = new MessageDigest.SHA1(provider, "SecretKeyFactory", configAlgName);
```
KostasTsiounis, line 256: Why do we need to get it as `SecretKeyFactory` instead of the regular way?

src/main/java/com/ibm/crypto/plus/provider/RC2Cipher.java
```diff
@@ -40,13 +40,19 @@ public final class RC2Cipher extends LegacyCipher {
...
 
     public RC2Cipher(OpenJCEPlusProvider provider) {
         this.provider = provider;
     }
 
+    public RC2Cipher(OpenJCEPlusProvider provider, String algName) {
+        this.provider = provider;
+        this.configAlgName = algName;
+    }
+
```
KostasTsiounis, line 55: Do we need both constructors?

src/main/java/com/ibm/crypto/plus/provider/RC4Cipher.java
```diff
@@ -34,13 +34,19 @@ public final class RC4Cipher extends LegacyCipher {
...
 
     public RC4Cipher(OpenJCEPlusProvider provider) {
         this.provider = provider;
     }
 
+    public RC4Cipher(OpenJCEPlusProvider provider, String configAlgName) {
+        this.provider = provider;
+        this.configAlgName = configAlgName;
+    }
+
```
KostasTsiounis, line 49: Similar comment.

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/base/Signature.java
KostasTsiounis: Sorry, I missed the part where the "Signature" is passed in the call to `Digest.getInstance()`.

src/main/java/com/ibm/crypto/plus/provider/MessageDigest.java
KostasTsiounis: I, also, figured out that it's used when SHA1 is utilized by another type.

src/main/java/com/ibm/crypto/plus/provider/PBES1Core.java
KostasTsiounis: Similar to other spots where it's used to use same backend as calling type. Understood now.

src/main/java/com/ibm/crypto/plus/provider/base/Signature.java
KostasTsiounis: A comment explaining that here would be nice.

src/main/java/com/ibm/crypto/plus/provider/PBES1Core.java
KostasTsiounis: A comment explaining that would be helpful here.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Changed to private

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Fixed.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Fixed.

src/main/java/com/ibm/crypto/plus/provider/base/NativeCryptoSelector.java
johnpeck-us-ibm: Added this new Exception.

src/main/java/com/ibm/crypto/plus/provider/base/NativeInterface.java
johnpeck-us-ibm: Changed this and int RSAKEY_size to be consistent.

src/main/java/com/ibm/crypto/plus/provider/base/NativeInterface.java
johnpeck-us-ibm: fixed

src/main/java/com/ibm/crypto/plus/provider/base/Signature.java
johnpeck-us-ibm: added

src/main/java/com/ibm/crypto/plus/provider/PBES1Core.java
johnpeck-us-ibm: Added comment

src/main/java/com/ibm/crypto/plus/provider/RC2Cipher.java
johnpeck-us-ibm: Removed other constructor.

src/main/java/com/ibm/crypto/plus/provider/RC4Cipher.java
johnpeck-us-ibm: Removed other constructor.

KostasTsiounis (approved):
LGTM

taoliult (approved):
LGTM

JinhangZhang (approved):
LGTM

PR #1519, opened 2026-06-05
Support SEED, EXPANDED, BOTH PrivateKey choice for FIPS 203 and 204

JinhangZhang:
Enhance PQCPrivateKey to recognize all supported PQC private key choice encodings instead of only handling DER OCTET STRING values with a fixed 4-byte header.

The previous OctectStringEncoded() logic only checked for tag 0x04 and assumed the length was encoded using two length bytes. This worked for some expanded private keys, but failed for other valid private key choices such as seed and both.

This update extends the detection logic to recognize:

- seed:     [0] IMPLICIT OCTET STRING, tag 0x80
- expanded: OCTET STRING, tag 0x04
- both:     SEQUENCE, tag 0x30

If the input is already encoded as a valid PQC private key choice, it is used directly. If no private key choice tag is present, the original behavior is preserved by wrapping the raw bytes as an OCTET STRING private key choice.

This avoids incorrectly re-wrapping seed or both encodings and prevents valid private key material from being rejected due to fixed-length header assumptions.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
@@ -46,24 +46,43 @@ final class PQCPrivateKey extends PKCS8Key {
         this.provider = provider;
         byte[] key = null;
         DerValue pkOct = null;
-        
-        //Check to determine if the key bytes already have the Octet tag.
-        if (OctectStringEncoded(keyBytes)) {
-            //Remove encoding OctetString encoding.
-            key = Arrays.copyOfRange(keyBytes, 4, keyBytes.length);
-        } else {
-            key = keyBytes;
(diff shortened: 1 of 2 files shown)


johnpeck-us-ibm:
I am not understanding how we can do this change. ICC today only works with the expanded private key. It will not work with just a seed value (BTW Bouncy Castle was just providing the seed at one point).  Now if the JVM will provider both in the encoding than we can strip off the seed part and make this work.  But I did not see that being done in the code.  Once. ICC will except all the encodings than this will be easy. The only issue with be performance with a seed. We may need an API to get the keys (public and private) from the seed so that we can add it to the encoding for our use.

JinhangZhang:
closed this one, opened another simple fix at https://github.com/IBM/OpenJCEPlus/pull/1520

PR #1520, opened 2026-06-10
Validate expanded PQC private key encodings

JinhangZhang:
Update PQCPrivateKey to validate ML-DSA and ML-KEM private key material using the RFC 9881 / RFC 9935 expandedKey CHOICE encoding.

The validation checks that the private key material contained in the PKCS#8 privateKey OCTET STRING is encoded as a DER OCTET STRING with the expected expanded key length for the corresponding parameter set. Unsupported private key encodings are rejected with an InvalidKeyException.

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,6 +532,8 @@ OpenJCEPlus provider enhances the security of Java applications by providing an 
 
 No keytool or certificate support was added other than what is already in a given Java runtime environment.
 
+OpenJCEPlus only accepts a `Expanded` Format of ML-KEM private key as defined in [RFC 9935](https://datatracker.ietf.org/doc/rfc9935/).
+
 ### ML-DSA
 
 Quantum-Resistant Module-Lattice-Based Digital Signature Algorithm (`ML-DSA`)
@@ -539,6 +541,8 @@ Quantum-Resistant Module-Lattice-Based Digital Signature Algorithm (`ML-DSA`)
 OpenJCEPlus provider enhances the security of Java applications by providing an implementation of quantum-resistant Module-Lattice-Based Digital Signature Algorithm (`ML-DSA`). Digital signatures are used to detect unauthorized modifications to data and to authenticate the identities of signatories. `ML-DSA` is designed to be secure against future quantum computing attacks and has been standardized by the United States National Institute of Standards and Technology (NIST) in [FIPS 204](https://csrc.nist.gov/pubs/fips/204/final).
 
 No keytool or certificate support was added other than what is already in a given Java runtime environment.
+
+OpenJCEPlus only accepts a `Expanded` Format of ML-DSA private key as defined in [RFC 9881](https://datatracker.ietf.org/doc/rfc9881/).
 
 ### ECKeyPairGenerator incorrect keysize
 
--- a/src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
@@ -33,6 +33,8 @@ final class PQCPrivateKey extends PKCS8Key {
 
     private transient boolean destroyed = false;
 
+    private boolean isSeed = false;
+
     /**
      * Create a PQC private key from the key data and the algorithm name.
      *
@@ -46,7 +48,9 @@ final class PQCPrivateKey extends PKCS8Key {
         this.provider = provider;
         byte[] key = null;
         DerValue pkOct = null;
-        
+
+        this.isSeed = isSeedOnly(this.name, keyBytes);
+
         //Check to determine if the key bytes already have the Octet tag.
         if (OctectStringEncoded(keyBytes)) {
             //Remove encoding OctetString encoding.
@@ -96,6 +100,7 @@ final class PQCPrivateKey extends PKCS8Key {
 
             this.name = PQCKnownOIDs.findMatch(pqcKey.getAlgorithm()).stdName();
             this.algid = new AlgorithmId(PQCAlgorithmId.getOID(name));
+            this.isSeed = isSeedOnly(this.name, pqcKey.getPrivateKeyBytes());
         } catch (Exception exception) {
             throw provider.providerException("Failure in PQCPrivateKey" + exception.getMessage(), exception);
         }
@@ -111,7 +116,7 @@ final class PQCPrivateKey extends PKCS8Key {
         this.provider = provider;
 
         this.name = PQCKnownOIDs.findMatch(this.algid.getName()).stdName();
-
+        this.isSeed = isSeedOnly(this.name, this.privKeyMaterial);
         //Check to determine if the key bytes have the Octet tag.
         if (!(OctectStringEncoded(this.privKeyMaterial))) {
             DerValue pkOct = null;
@@ -159,7 +164,11 @@ final class PQCPrivateKey extends PKCS8Key {
             DerOutputStream bytes = new DerOutputStream();
             bytes.putOID(algid.getOID());
             tmp.write(DerValue.tag_Sequence, bytes);
-            tmp.putOctetString(this.privKeyMaterial);
+            if (isSeed) {
+                tmp.write(this.privKeyMaterial);
+            } else {
+                tmp.putOctetString(this.privKeyMaterial);
+            }
             DerValue out = DerValue.wrap(DerValue.tag_Sequence, tmp);
             encodedKey = out.toByteArray();
             tmp.close();
@@ -235,4 +244,38 @@ final class PQCPrivateKey extends PKCS8Key {
         }
     }
 
+    /**
+     * Determines whether the supplied private key material represents a
+     * seed-only PQC private key.
+     *
+     * <p>RFC 9881 and RFC 9935 define PQC private key material as a CHOICE.
+     * A seed-only key is encoded with the context-specific [0] tag. For
+     * ML-DSA, the seed length is 32 bytes. For ML-KEM, the seed length is
+     * 64 bytes.</p>
+     *
+     * @param algName the standard PQC algorithm name
+     * @param key the private key material to check
+     *
+     * @return true if the key material is a seed-only private key encoding;
+     *         false otherwise
+     */
+    private boolean isSeedOnly(String algName, byte[] key) {
+        if (key == null || key.length < 2) {
+            return false;
+        }
+
+        int seedLen;
+
+        if (algName.startsWith("ML-DSA")) {
+            seedLen = 32;
+        } else if (algName.startsWith("ML-KEM")) {
+            seedLen = 64;
+        } else {
+            return false;
+        }
+
+        return (key.length == seedLen + 2)
+                && ((key[0] & 0xFF) == 0x80)
+                && ((key[1] & 0xFF) == seedLen);
+    }
 }
--- a/src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
@@ -11,6 +11,7 @@ package com.ibm.crypto.plus.provider;
 import com.ibm.crypto.plus.provider.base.PQCKey;
 import java.io.IOException;
 import java.security.InvalidKeyException;
+import java.security.ProviderException;
 import java.util.Arrays;
 import javax.security.auth.DestroyFailedException;
 import sun.security.pkcs.PKCS8Key;
@@ -79,7 +80,13 @@ final class PQCPrivateKey extends PKCS8Key {
         try {
             this.provider = provider;
             this.pqcKey = pqcKey;
+            this.name = PQCKnownOIDs.findMatch(pqcKey.getAlgorithm()).stdName();
+            this.algid = new AlgorithmId(PQCAlgorithmId.getOID(name));
 
+            validateKeyLength(pqcKey.getPrivateKeyBytes());
+            if (!isExpandedChoice(this.name, pqcKey.getPrivateKeyBytes())) {
+                throw new InvalidKeyException("Only expanded keys are supported by OpenJCEPlus");
+            }
             //Check to determine if the key bytes have the Octet tag.
             if (OctectStringEncoded(pqcKey.getPrivateKeyBytes())) {
                 this.privKeyMaterial = pqcKey.getPrivateKeyBytes();
@@ -93,9 +100,6 @@ final class PQCPrivateKey extends PKCS8Key {
                     pkOct.clear();
                 }
             }
-
-            this.name = PQCKnownOIDs.findMatch(pqcKey.getAlgorithm()).stdName();
-            this.algid = new AlgorithmId(PQCAlgorithmId.getOID(name));
         } catch (Exception exception) {
             throw provider.providerException("Failure in PQCPrivateKey" + exception.getMessage(), exception);
         }
@@ -111,7 +115,10 @@ final class PQCPrivateKey extends PKCS8Key {
         this.provider = provider;
 
         this.name = PQCKnownOIDs.findMatch(this.algid.getName()).stdName();
-
+        validateKeyLength(this.privKeyMaterial);
+        if (!isExpandedChoice(this.name, this.privKeyMaterial)) {
+            throw new InvalidKeyException("Only expanded keys are supported by OpenJCEPlus");
+        }
         //Check to determine if the key bytes have the Octet tag.
         if (!(OctectStringEncoded(this.privKeyMaterial))) {
             DerValue pkOct = null;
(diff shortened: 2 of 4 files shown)


src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
```diff
@@ -739,4 +741,80 @@ public void testMLKEMBidirectionalInteropWithNamedParameterSpec(String parameter
...
+
+    @ParameterizedTest
+    @CsvSource({"ML-DSA", "ML-DSA-44", "ML-DSA-65", "ML-DSA-87"})
+    public void testMLDSAGetKeySpecPrivateInteropToPlus(String algorithm)
+            throws Exception {
+        // This is not in the FIPS provider yet.
+        assumeFalse("OpenJCEPlusFIPS".equals(getProviderName()));
+
+        // This test is for SunJCE/OpenJCEPlus interop, not BC.
+        assumeFalse(Utils.PROVIDER_BC.equals(getInteropProviderName2()));
```
jasonkatonica: Does BC have a different encoding for its seeded form and hence does not work correctly and needs to be skiped?

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
```diff
@@ -739,4 +741,80 @@ public void testMLKEMBidirectionalInteropWithNamedParameterSpec(String parameter
...
 
+    @ParameterizedTest
+    @CsvSource({"ML-KEM", "ML-KEM-512", "ML-KEM-768", "ML-KEM-1024"})
+    public void testMLKEMGetKeySpecPrivateInteropToPlus(String algorithm)
+            throws Exception {
+        // This is not in the FIPS provider yet.
+        assumeFalse("OpenJCEPlusFIPS".equals(getProviderName()));
+
+        // This test is for SunJCE/OpenJCEPlus interop, not BC.
+        assumeFalse(Utils.PROVIDER_BC.equals(getInteropProviderName()));
```
jasonkatonica, line 754: Does BC have a different encoding for its seeded form and hence does not work correctly and needs to be skiped?

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -46,7 +48,11 @@ final class PQCPrivateKey extends PKCS8Key {
         this.provider = provider;
         byte[] key = null;
         DerValue pkOct = null;
-        
+
+        if (isSeedOnly(this.name, keyBytes)) {
+            this.isSeed = true;
+        }
```
KostasTsiounis: Why not just do `this.isSeed = isSeedOnly(this.name, keyBytes);` directly?

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -96,6 +102,9 @@ final class PQCPrivateKey extends PKCS8Key {
 
             this.name = PQCKnownOIDs.findMatch(pqcKey.getAlgorithm()).stdName();
             this.algid = new AlgorithmId(PQCAlgorithmId.getOID(name));
+            if (isSeedOnly(this.name, pqcKey.getPrivateKeyBytes())) {
+                this.isSeed = true;
+            }
```
KostasTsiounis: Similar comment as above.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -111,7 +120,9 @@ final class PQCPrivateKey extends PKCS8Key {
         this.provider = provider;
 
         this.name = PQCKnownOIDs.findMatch(this.algid.getName()).stdName();
-
+        if (isSeedOnly(this.name, this.privKeyMaterial)) {
+            this.isSeed = true;
+        }
```
KostasTsiounis: Similar comment as above.

KostasTsiounis (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
JinhangZhang: Doesnt seem so, added BC back, tests work fine.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
JinhangZhang: Doesnt seem so, added BC back, tests work fine.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: updated

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: updated

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: updated

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
jasonkatonica: OK in this case you can remove the commented `// assumeFalse(Utils.PROVIDER_BC.equals(getInteropProviderName()));` lines. Similar comment in the other locations.

jasonkatonica:
Please add a KAT that uses the seeded form of the key from the RFC. The jacoco code test coverage indicates that none of the paths added related to seeded form of the encoding are being run.

jasonkatonica:
The github actions are also currently failing. Likely due to the removal of the BC assumption talked about above.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
```diff
@@ -739,4 +1877,252 @@ public void testMLKEMBidirectionalInteropWithNamedParameterSpec(String parameter
...
+
+        byte[] rfcPrivateKeyEncoded = decodePEM(getRFCPrivateKeyPEM(privateKeyName));
+        byte[] rfcPublicKeyEncoded = decodePEM(getRFCPublicKeyPEM(publicKeyName));
+
+        PrivateKey sunPrivateKey = sunKeyFactory.generatePrivate(new PKCS8EncodedKeySpec(rfcPrivateKeyEncoded));
+        PublicKey sunPublicKey = sunKeyFactory.generatePublic(new X509EncodedKeySpec(rfcPublicKeyEncoded));
+
+        PKCS8EncodedKeySpec sunPrivateKeySpec = sunKeyFactory.getKeySpec(sunPrivateKey, PKCS8EncodedKeySpec.class);
+
+        PrivateKey openjceplusPrivateKey = openjceplusKeyFactory.generatePrivate(sunPrivateKeySpec);
```
johnpeck-us-ibm: This test is similar to the above ones where we are doing interopt testing. If this is a KAT test than ALL keys need to come from the specification. Also, tests should be done at some point such that given the SEED that the code produces the expected Private key and Public key.

johnpeck-us-ibm (changes requested):

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
JinhangZhang: Removed

JinhangZhang:
> Please add a KAT that uses the seeded form of the key from the RFC. The jacoco code test coverage indicates that none of the paths added related to seeded form of the encoding are being run.

Added

JinhangZhang:
> The github actions are also currently failing. Likely due to the removal of the BC assumption talked about above.

Disabled BC provider

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
JinhangZhang: Updated the test to only compare the encoding, and moved it to another based test

jasonkatonica (approved):
LGTM

johnpeck-us-ibm (approved):

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -100,7 +98,9 @@ final class PQCPrivateKey extends PKCS8Key {
 
             this.name = PQCKnownOIDs.findMatch(pqcKey.getAlgorithm()).stdName();
             this.algid = new AlgorithmId(PQCAlgorithmId.getOID(name));
-            this.isSeed = isSeedOnly(this.name, pqcKey.getPrivateKeyBytes());
+            if (isSeedOnly(this.name, pqcKey.getPrivateKeyBytes())) {
+                throw new InvalidKeyException("Seed-only keys are not supported by OpenJCEPlus at the moment");
```
jasonkatonica: I would leave out "at the moment" from the exception since this implies we will certainly fix this in the future which we are not sure of at this time.

taoliult (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -59,9 +58,12 @@ final class PQCPrivateKey extends PKCS8Key {
         try {
             try {
                 pkOct = new DerValue(DerValue.tag_OctetString, key);
+                this.privKeyMaterial = pkOct.toByteArray();
+                if (isSeedOnly(this.name, keyBytes)) {
```
jasonkatonica: Perhaps we should have the opposite logic here and check if we have the expanded form of the key encoding. If we do then it is ok to proceed. If not then we should fail with an exception.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -46,7 +46,9 @@ final class PQCPrivateKey extends PKCS8Key {
         this.provider = provider;
         byte[] key = null;
         DerValue pkOct = null;
-        
+        if (!isExpandedChoice(this.name, keyBytes)) {
+            throw new InvalidKeyException("Only expanded keys are supported by OpenJCEPlus at the moment");
```
jasonkatonica: Please remove the `at the moment` since this implies we will be able to certainly support other encodings in the future which is not correct. Similar comment in other exception messages below.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeys.java
```diff
@@ -134,6 +179,54 @@ public void genWithAlgParameterSpecMLDSAFaiure(String algParamSpecName) throws E
 
     }
 
+    @ParameterizedTest
+    @MethodSource("rfcSeedPrivateKeys")
+    public void testRFC9881MLDSARFC9935MLKEMKeyFactory(String algorithm, String privateKeyPem) throws Exception {
+
+        if (getProviderName().equals("OpenJCEPlusFIPS")) {
```
jasonkatonica: We should use the `assume` syntax here so we skip the test instead of marking the test as passing for `OpenJCEPlusFIPS` provider.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeys.java
```diff
@@ -134,6 +179,54 @@ public void genWithAlgParameterSpecMLDSAFaiure(String algParamSpecName) throws E
...
+
+        KeyFactory openjceplusKeyFactory = KeyFactory.getInstance(algorithm, getProviderName());
+        byte[] rfcPrivateKeyEncoded = decodePEM(privateKeyPem);
+
+        String expectedMessage = "Only expanded keys are supported by OpenJCEPlus at the moment";
+        try {
+            openjceplusKeyFactory.generatePrivate(new PKCS8EncodedKeySpec(rfcPrivateKeyEncoded));
+            fail("Expected InvalidKeySpecException for seed-only private key.");
+        } catch (InvalidKeySpecException e) {
+            assertTrue(e.getMessage().contains(expectedMessage)
```
jasonkatonica: This could be simplified by using Junit 5 `assertEquals(expected, actual);`. In this case it will assert the strings are the same and print them if they are not equal.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -235,4 +241,54 @@ private boolean OctectStringEncoded(byte[] key) {
...
+     * <p>This method checks the private key material contained in the PKCS#8
+     * privateKey OCTET STRING, not the complete PKCS#8 encoding.</p>
+     *
+     * @param algName the standard PQC algorithm name
+     * @param key the private key material to check
+     *
+     * @return true if the key material is an expanded private key encoding;
+     *         false otherwise
+     */
+
```
jasonkatonica: Space should be deleted here for javadoc.

jasonkatonica: Space should be deleted here for javadoc.

jasonkatonica: This is a duplicate comment for some reason resolving this.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: Removed

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeys.java
JinhangZhang: Use assume

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: deleted

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeys.java
JinhangZhang: updated

README.md
```diff
@@ -555,6 +555,9 @@ AES Key Wrap based on NIST SP800-38F.
 
 Code does not allow the specification of an IV. However, it will return the default ICV as defined in the NIST SP800-38F. 
 
+### MLDSA and MLKEM privatekey
```
jasonkatonica: In the table above we provide links to these sections. Please add a link to this header so users are able to see the notes associated with this in the PQC key generator and PQC key factory rows.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -235,4 +239,53 @@ private boolean OctectStringEncoded(byte[] key) {
...
+     * @param algName the standard PQC algorithm name
+     * @param key the private key material to check
+     *
+     * @return true if the key material is an expanded private key encoding;
+     *         false otherwise
+     */
+    private boolean isExpandedChoice(String algName, byte[] key) {
+        if (algName == null || key == null || key.length < 4) {
+            return false;
+        }
```
KostasTsiounis: Do these checks belong here? Or before you call this method? Only asking cause I would expect a different exception and exception message if any of these is true.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -235,4 +239,53 @@ private boolean OctectStringEncoded(byte[] key) {
...
+        } else if ("ML-DSA-87".equals(algName)) {
+            expandedLen = 4896;
+        } else if ("ML-KEM-512".equals(algName)) {
+            expandedLen = 1632;
+        } else if ("ML-KEM-768".equals(algName)) {
+            expandedLen = 2400;
+        } else if ("ML-KEM-1024".equals(algName)) {
+            expandedLen = 3168;
+        } else {
+            return false;
```
KostasTsiounis: Similar comment here. If one gets there, it's not just that it's not an expanded key, but rather the algorithm name is problematic.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
```diff
@@ -112,6 +114,8 @@ public void testPQCKeyGenKEM_Interop() throws Exception {
 
         //This is not in the FIPS provider yet.
         assumeFalse("OpenJCEPlusFIPS".equals(getProviderName()));
+        assumeFalse(Utils.PROVIDER_BC.equals(getInteropProviderName()));
+        System.out.println("Interop Provider: " + getInteropProviderName());
```
KostasTsiounis: I'm guessing this is leftover from debug?

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
```diff
@@ -214,7 +218,8 @@ public void testPQCKeyGenMLDSA_Interop() throws Exception {
 
         //This is not in the FIPS provider yet.
         assumeFalse("OpenJCEPlusFIPS".equals(getProviderName()));
-
+        assumeFalse(Utils.PROVIDER_BC.equals(getInteropProviderName()));
+        System.out.println("Interop Provider: " + getInteropProviderName2());
```
KostasTsiounis: Same as above.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
```diff
@@ -214,7 +218,8 @@ public void testPQCKeyGenMLDSA_Interop() throws Exception {
 
         //This is not in the FIPS provider yet.
         assumeFalse("OpenJCEPlusFIPS".equals(getProviderName()));
-
+        assumeFalse(Utils.PROVIDER_BC.equals(getInteropProviderName()));
```
KostasTsiounis, line 222: Could we add a short comment on why BC is not supported?

KostasTsiounis (changes requested):

README.md
JinhangZhang: Added the link

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: removed the `algName == null` but keep the rest since we didnt check the key before.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: Updated. you are right, the isExpandedChoice should only focus on the key length not the algorithm name.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
JinhangZhang: Right, removed

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
JinhangZhang: Right, removed

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
JinhangZhang: Added, BC generates the seed format.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
KostasTsiounis: I'm not saying we shouldn't check these, but rather that they should be checked before calling this method at all.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
KostasTsiounis: Similarly, should we check the algorithm name before getting here? Although I'd be ok leaving this as is, given that we'd need another if/switch statement to check separately.

README.md
jasonkatonica: Updated comments in the existing sections looks good to me.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
jasonkatonica: I agree that checking for null prior to calling this method makes sense. This will allow for a easier path in the future when we still want to check for null yet validate other key encodings such as `seed` and `both` forms when appropriate.

johnpeck-us-ibm (approved):

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: Refactored

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: Refactored

jasonkatonica: These checks dont seem to be called before the call to `isExpandedChoice`. Additionally this does not throw a different exception message to the caller such that they know they passed an incorrect lenght or null key.

jasonkatonica: Code look good now to me.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
jasonkatonica: Code look good now to me.

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

PR #1522, opened 2026-06-11
Disable ThreadLocal caching for ExtendedRandom on z/OS

taoliult:
ThreadLocal can cause native TCB context issues on z/OS. This change bypasses ThreadLocal on z/OS by creating instance-specific PRNG contexts, while other platforms continue using ThreadLocal for performance.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
@@ -14,6 +14,10 @@ import com.ibm.crypto.plus.provider.ock.NativeOCKAdapterNonFIPS;
 
 public final class ExtendedRandom {
 
+    private static final boolean IS_ZOS = System.getProperty("os.name")
+                                                .toLowerCase()
+                                                .contains("z/os");
+
     private OpenJCEPlusProvider provider;
     private NativeInterface nativeInterface;
     private final String algName;
@@ -21,8 +25,10 @@ public final class ExtendedRandom {
     private long ockPRNGContextId;
     private boolean usingThreadLocalContext = true;
 
-    private static final ThreadLocal<PRNGContextPointer> prngContextBufferSha256 = new ThreadLocal<PRNGContextPointer>();
-    private static final ThreadLocal<PRNGContextPointer> prngContextBufferSha512 = new ThreadLocal<PRNGContextPointer>();
+    private static final ThreadLocal<PRNGContextPointer> prngContextBufferSha256 =
+        IS_ZOS ? null : new ThreadLocal<PRNGContextPointer>();
+    private static final ThreadLocal<PRNGContextPointer> prngContextBufferSha512 =
+        IS_ZOS ? null : new ThreadLocal<PRNGContextPointer>();
 
     public static ExtendedRandom getInstance(String algName, OpenJCEPlusProvider provider)
             throws NativeException {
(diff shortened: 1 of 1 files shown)


src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
```diff
@@ -45,6 +51,15 @@ private ExtendedRandom(String algName, OpenJCEPlusProvider provider) throws Nati
     }
 
     private long getPRNGContext(String algName) throws NativeException {
+        // On z/OS, bypass ThreadLocal completely.
+        if (IS_ZOS) {
+            long instanceCtx = this.nativeInterface.EXTRAND_create(algName);
+            this.usingThreadLocalContext = false;
+            this.provider.registerCleanable(this, cleanOCKResources(instanceCtx, nativeInterface));
+            return instanceCtx;
```
KostasTsiounis: Would it make sense to extract this into a helper method given the exact same code is used in `createInstanceContextForReSeed()`?

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
```diff
@@ -45,6 +51,15 @@ private ExtendedRandom(String algName, OpenJCEPlusProvider provider) throws Nati
     }
 
     private long getPRNGContext(String algName) throws NativeException {
+        // On z/OS, bypass ThreadLocal completely.
```
KostasTsiounis: I would update the comment to indicate that a cached version is not used, instead of mentioning ThreadLocal.

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
```diff
@@ -45,6 +51,15 @@ private ExtendedRandom(String algName, OpenJCEPlusProvider provider) throws Nati
...
+        if (IS_ZOS) {
+            long instanceCtx = this.nativeInterface.EXTRAND_create(algName);
+            this.usingThreadLocalContext = false;
+            this.provider.registerCleanable(this, cleanOCKResources(instanceCtx, nativeInterface));
+            return instanceCtx;
+        }
+
+        // Non-z/OS platforms, use ThreadLocal.
```
KostasTsiounis: Same as above.

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
taoliult: Created `createInstanceContext()` helper method, and also removed `createInstanceContextForReSeed()` method - Inlined its logic into `setSeed()` and `createInstanceContext()`.

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
taoliult: Yes, updated.

src/main/java/com/ibm/crypto/plus/provider/base/ExtendedRandom.java
taoliult: Yes, updated.

johnpeck-us-ibm (approved):

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

JinhangZhang (approved):
LGTM

PR #1534, opened 2026-06-18
Fix AES/CCM ByteBuffer operations

### LINKED ISSUE #1504: Fix AESCCMCipher potential buffer overflow issue
https://github.com/IBM/OpenJCEPlus/blob/1df365e8ce77149140790afae650afcaa0002817/src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java#L163

https://github.com/IBM/OpenJCEPlus/blob/1df365e8ce77149140790afae650afcaa0002817/src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java#L188

**The Core Problem: The `array()` Misconception**

When the code calls `.array()` on a `ByteBuffer`, it is gaining direct access to the entire underlying heap array. However, a `ByteBuffer` is rarely meant to be read from index `0` to `array.length`. 
A `ByteBuffer` defines a specific view of data using three markers:
- `position`: Where the next read/write starts.
- `limit`: The end of the valid data you are allowed to read.
- `remaining()`: The actual slice of data meant to be processed (_limit - position_).

**Visualizing the Flaw**
Imagine a user passes the code a `ByteBuffer` that contains a slice of network data:
```
Backing Array [XX, XX, AA, BB, CC, DD, XX, XX, XX]
                       ^               ^
                    position         limit
```
- `inputByteBuffer.remaining()` is `4` (The valid data: `AA, BB, CC, DD`).
- `inputByteBuffer.array().length` is `9`.

By using `inputByteBuffer.array()`, the code is ignoring the user's `position` and `limit`. It is pulling all `9` bytes (including old/garbage data marked as `XX`) and passing them to the internal `engineDoFinal` method.

**3 Severe Consequences of the Current Code**
1. It Completely Ignores Sliced Buffers
If a developer uses `ByteBuffer.slice()` or modifies the `position()` before passing it to the cipher, the method completely ignores it. It will encrypt or decrypt data starting from index `0` of the backing array instead of the active position.

2. It Triggers `UnsupportedOperationException`
Not all `ByteBuffer` instances are backed by an accessible array.
  - If a user passes a **Direct ByteBuffer** (e.g., allocated via `ByteBuffer.allocateDirect()`, common in high-performance I/O or Netty), or a **Read-Only ByteBuffer**, calling `.array()` will immediately throw an `UnsupportedOperationException`. The code will instantly crash.

3. Incorrect Output Allocation Size
Because it sizes the `output` array using `inputByteBuffer.array().length`, the code is allocating memory based on the **capacity** of the incoming buffer, not the **amount** of data to be processed. This leads to wasted memory allocation and potential `BufferOverflowException` downstream.

FYI @jasonkatonica

jasonkatonica:
1. Fixed BufferOverflowException by using ByteBuffer.remaining() instead of array().length.

2. Added ShortBufferException to method signature for proper exception handling. Validates output buffer has sufficient space.

3. Enhanced input validation for decryption operations. Added check to ensure input length >= tag length before decryption

4. Improved buffer position handling. Uses ByteBuffer.get() to extract data from current position. Correctly updates output buffer position with actual output length

Added TestAESCCMByteBuffer.java with comprehensive test coverage:
- Tests for various ByteBuffer scenarios (standard, positioned, limited, sliced, direct)
- Validates exact-sized output buffers
- Tests empty buffer encryption/decryption
- Verifies proper exception handling for short buffers
- Parameterized tests for both AES-128 and AES-256 key sizes

Fixes: https://github.com/IBM/OpenJCEPlus/issues/1504

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java
@@ -161,7 +161,7 @@ public final class AESCCMCipher extends CipherSpi implements AESConstants, CCMCo
 
     @Override
     protected int engineDoFinal(ByteBuffer inputByteBuffer, ByteBuffer outputByteBuffer)
-            throws IllegalBlockSizeException, BadPaddingException, AEADBadTagException {
+            throws IllegalBlockSizeException, BadPaddingException, AEADBadTagException, ShortBufferException {
         //final String methodName = "byte[] enginedoFinal";
 
         if (inputByteBuffer == null) {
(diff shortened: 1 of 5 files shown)


johnpeck-us-ibm (approved):

JinhangZhang (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java
```diff
@@ -177,37 +177,42 @@ protected int engineDoFinal(ByteBuffer inputByteBuffer, ByteBuffer outputByteBuf
...
-            // OCKDebug.Msg(debPrefix, methodName, "NativeException seen");
-            if (!encrypting) {
-                AEADBadTagException abte = new AEADBadTagException(
-                        "Uanble to perform engine doFinal; Possibly a bad tag or bad padding or illegalBlockSize");
-                provider.setExceptionCause(abte, e);
-                throw abte;
-            } else {
-                throw provider.providerException("unable to perform to engineDoFinal ", e);
-            }
+            throw e;
```
KostasTsiounis: Why catch it at all just to rethrow it?

KostasTsiounis (changes requested):

taoliult (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java
jasonkatonica: Agreed, updated code to remove the catch block in this area for `ShortBufferException` and the re-throw.

KostasTsiounis (approved):
LGTM

PR #1541, opened 2026-06-22
Refactor JUnit 5 Tests for BaseTestDSASignatureInterop2 test.

Mohit-Rajbhar100698:
This change replace multiple test classes extending BaseTestDSASignatureInterop2 into JUnit 5 parameterized test class.


johnpeck-us-ibm (approved):

src/test/java/ibm/jceplus/junit/tests/TestArguments.java
```diff
@@ -55,7 +55,17 @@ public static Stream<Arguments> rsaMultithreadKeySizesAndProviders() {
...
+    }
+
+    /**
+     * Generates combination of only OpenJCEPlus (non-FIPS) provider
+     * with the BC provider for interoperability testing.
+     *
+     * @return Stream of Arguments containing (OpenJCEPlus, BC) pair
+     */
+    protected static Stream<Arguments> getOpenJCEPlusOnlyWithBCInteropProvider() {
+        return Stream.of(Arguments.of(TestProvider.OpenJCEPlus, TestProvider.BC));
```
jasonkatonica: This hardcodes `TestProvider.OpenJCEPlus`. There is however logic on line 106 below that honors a `-Dgroups` and allows tests to be filtered. Looking at other methods seems like the method `getOpenJCEPlusOnly()` may also have this same problem.

src/test/java/ibm/jceplus/junit/tests/TestArguments.java
```diff
@@ -55,7 +55,17 @@ public static Stream<Arguments> rsaMultithreadKeySizesAndProviders() {
...
-    }    
+    }
+
+    /**
+     * Generates combination of only OpenJCEPlus (non-FIPS) provider
+     * with the BC provider for interoperability testing.
+     *
+     * @return Stream of Arguments containing (OpenJCEPlus, BC) pair
+     */
+    protected static Stream<Arguments> getOpenJCEPlusOnlyWithBCInteropProvider() {
```
jasonkatonica, line 66: Its interesting that this is declared as `protected` where other methods are `public`. Perhaps all our methods are better off being `protected`. I did validate in the logs that TestDSASignatureInterop2 was running as expected so likely we could reduce scope in the other methods also.

jasonkatonica (changes requested):

src/test/java/ibm/jceplus/junit/tests/TestArguments.java
Mohit-Rajbhar100698: Yes @jasonkatonica , -Dgroups allows tests to be filtered based on the active provider.
But there is one issue. The method in this class(TestArguements) acts as the method source. If no argument is passed in -Dgroups, then by default it returns both providers through the method source [(line 110),](https://github.com/Mohit-Rajbhar100698/OpenJCEPlus/blob/9badfdfd10419716d33383b4c894de886742944b/src/test/java/ibm/jceplus/junit/tests/TestArguments.java#L110) 
In that case, what happens if a test class is supposed to run only with the OpenJCEPlus provider The method source would contain:
`Stream<TestProvider> containing:`
  `[0] → TestProvider.OpenJCEPlus`
  `[1] → TestProvider.OpenJCEPlusFIPS`
The test will then start failing because it is also executed with the second parameter from the method source. This is the case for tests like TestDSASignatureInterop2, TestChaCha20, and others.

So, for tests that are intended to use only the OpenJCEPlus provider, if nothing is passed through -Dgroups, they should receive only the OpenJCEPlus provider from the method source.

We can handle this in two ways:
1. As each test class has a @ BeforeEach annotated setup() method, we can add a condition in those test classes to skip execution if the OpenJCEPlusFIPS provider is passed through the method source.

```
if (provider.getProviderName().equals("OpenJCEPlusFIPS")) {
    assumeTrue(false); // this test is not supposed to run in FIPS mode
}
```
2. We should honor -Dgroups and remove the hardcoded TestProvider.OpenJCEPlus.
We can modify the existing getOpenJCEPlusOnly() method for test classes that use only the OpenJCEPlus provider, like this:
```
protected static Stream<TestProvider> getOpenJCEPlusOnly() {
    String[] groupPropertyTags = BaseTest.getTagsPropertyAsArray();

    if (groupPropertyTags.length == 0) {
        return Stream.of(TestProvider.OpenJCEPlus);
    }

    for (String tag : groupPropertyTags) {
        if (TestProvider.OpenJCEPlus.getProviderName().equalsIgnoreCase(tag)) {
            return Stream.of(TestProvider.OpenJCEPlus);
        }
    }
    return Stream.empty(); // -Dgroups was set but did not include OpenJCEPlus
}```

 This covers all cases for tests that use only the OpenJCEPlus provider:
* If -Dgroups is empty, the default OpenJCEPlus provider will be returned.
* If -Dgroups=OpenJCEPlus, it will return only the OpenJCEPlus provider.
* If -Dgroups=OpenJCEPlus,OpenJCEPlusFIPS, it will still return only the OpenJCEPlus provider.
* If -Dgroups=OpenJCEPlusFIPS, it will return an empty stream (so the test is automatically filtered out).

src/test/java/ibm/jceplus/junit/tests/TestArguments.java
Mohit-Rajbhar100698: ok I will reduce scope in other methods as well

src/test/java/ibm/jceplus/junit/tests/TestArguments.java
jasonkatonica: Hi @Mohit-Rajbhar100698 What you propose seems reasonable to me in the above `getOpenJCEPlusOnly` method. Can you describe the behavior what happens if `Stream.empty()` is returned? I would assume the test gets skipped if we return an empty stream to the test? This would allow us to filter out tests when needed.

Mohit-Rajbhar100698: Hi @jasonkatonica 

`Stream.empty()` will actually never get executed in a valid case, since only tests that are meant to run with the `OpenJCEPlus` provider will use this new getOpenJCEPlusOnly() method.

If `-Dgroups=OpenJCEPlusFIPS` is passed, all the tests tagged with `OpenJCEPlus` are filtered out before getOpenJCEPlusOnly() is called.

I tried this with `-Dgroups=OpenJCEPlusFIPS` and checked it in the debugger. It never entered `getOpenJCEPlusOnly()` because those `OpenJCEPlus` tests had already been filtered out.

`Stream.empty()` will only be returned if an OpenJCEPlusFIPS test(test that use this provider) calls `getOpenJCEPlusOnly()`. But I think that is an invalid case, because this method is only supposed to be used by OpenJCEPlus tests.

KostasTsiounis (approved):
LGTM

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):

jasonkatonica (approved):
LGTM

PR #1554, opened 2026-06-24
Prevent key pair gen in EdDSA/XDH constructor

### LINKED ISSUE #1521: Key Pair creation in XHDPrivateKeyImpl and EdDSAPrivateKeyImpl
The following files at the specified lines contain code to create key pairs XHDPrivateKeyImpl (lines 138 - 141) and EdDSAPrivateKeyImpl ( lines 95 - 99). This does not make sense and no one knows of a scenario where this is needed.

Therefore, this code should be removed and a exception thrown instead of the key pair creation. 

Please run all Tests. JTREG, JCK, etc to validate that the above is not a valid scenario. If there is a valid scenario, that should be documented in the code where the key pair is generated.

jasonkatonica:
Key pair generation in a private key constructor is not a valid scenario. This change ensures proper key construction and prevents unexpected behavior.

- Add validation to reject null key bytes/scalar in EdDSAPrivateKeyImpl and XDHPrivateKeyImpl
- Throw InvalidParameterException when attempting to create private keys without key material
- Add emergency fallback system properties for backward compatibility:
  * openjceplus.eddsa.allowKeyPairGenerationInConstructor
  * openjceplus.xdh.allowKeyPairGenerationInConstructor

Fixes: https://github.com/IBM/OpenJCEPlus/issues/1521

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/EdDSAPrivateKeyImpl.java
@@ -29,6 +29,14 @@ final class EdDSAPrivateKeyImpl extends PKCS8Key implements EdECPrivateKey {
 
     private static final long serialVersionUID = 1L;
 
+    // Emergency fallback property to allow legacy key pair generation behavior
+    // when h parameter is null. This should only be used in emergency situations
+    // as generating a key pair in a private key constructor is not a valid scenario.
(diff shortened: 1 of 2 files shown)


johnpeck-us-ibm (approved):

KostasTsiounis (approved):
LGTM

JinhangZhang (approved):
LGTM

PR #1565, opened 2026-07-01
Fix AESCCMCipher.engineDoFinal to use inputLen

### LINKED ISSUE #1564: AESCCMCipher.engineDoFinal output buffer size check uses input.length instead of inputLen
In [AESCCMCipher.java](https://github.com/IBM/OpenJCEPlus/blob/main/src/main/java/com/ibm/crypto/plus/provider/AESCCMCipher.java), the 5-argument overload of `engineDoFinal` performs an output buffer size check that incorrectly uses input.length (the full backing array length) rather than `inputLen` (the number of bytes the caller actually wants to encrypt/decrypt).

**Affected code (lines 225–235):**
```
if (encrypting) {
    if ((output.length - outputOffset) < (input.length + tagLenInBytes)) {  // BUG: input.length
        throw new ShortBufferException(
                "The output buffer is too small to hold the encryption result.");
    }
} else { // decrypting
    if ((output.length - outputOffset) < (input.length - tagLenInBytes)) {  // BUG: input.length
        throw new ShortBufferException(
                "The output buffer is too small to hold the decryption result.");
    }
}
```
**Expected code:**
```
if (encrypting) {
    if ((output.length - outputOffset) < (inputLen + tagLenInBytes)) {
        throw new ShortBufferException(
                "The output buffer is too small to hold the encryption result.");
    }
} else { // decrypting
    if ((output.length - outputOffset) < (inputLen - tagLenInBytes)) {
        throw new ShortBufferException(
                "The output buffer is too small to hold the decryption result.");
    }
}
```
**Reproducer scenario**

Consider a caller that provides a large input array but only wants to process a slice:
```
byte[] input = new byte[100];
int inputOffset = 40;
int inputLen = 10;
int tagLenInBytes = 16;  // e.g. 128-bit tag

// Caller allocates the exact right output size based on inputLen
byte[] output = new byte[inputLen + tagLenInBytes];  // 26 bytes — correct

// This call throws ShortBufferException even though output is correctly sized:
// check computes (26 - 0) < (100 + 16) → 26 < 116 → true → spurious exception
int outputLen = engineDoFinal(input, inputOffset, inputLen, output, 0);
```

The output buffer is correctly sized for the actual work being requested (inputLen + tagLenInBytes = 26), but the check compares it against the entire input array length plus the tag (input.length + tagLenInBytes = 116), causing a spurious `ShortBufferException`.

This also affects the internal `engineDoFinal(byte[], int, int)` overload, which allocates `output = new byte[inputLen + tagLenInBytes]` and then calls the 5-arg form — the internal call will always fail with a `ShortBufferException`whenever `inputLen != input.length` (i.e., whenever `inputOffset > 0` or the input array is larger than the slice being processed).

**Fix**

Replace `input.length` with `inputLen` in both the encrypt and decrypt output buffer size checks at lines 226 and 231.

jasonkatonica:
AESCCMCipher.engineDoFinal was using input.length (the full backing array length) instead of inputLen (the caller-specified byte count) when deciding whether the output buffer was large enough. This caused a false ShortBufferException whenever the caller passed a slice of a larger backing array via inputOffset + inputLen, because the backing array was wider than the actual data being processed.

Fix: replace input.length with inputLen in both the encrypt and decrypt output-size checks inside engineDoFinal.

Tests were also added to exercise slicing and ensure interop via sliced buffers is working as expected.

Fixes: https://github.com/IBM/OpenJCEPlus/issues/1564

### DIFF
--- a/src/test/java/ibm/jceplus/junit/base/BaseTestAESCCMInteropBC.java
+++ b/src/test/java/ibm/jceplus/junit/base/BaseTestAESCCMInteropBC.java
@@ -289,6 +289,129 @@ public class BaseTestAESCCMInteropBC extends BaseTestJunit5Interop {
         }
     } // end testAESCCM()
 
+    /**
+     * Tests that engineDoFinal correctly uses inputLen (not input.length) for the output buffer
+     * size check during cross-provider interop with BouncyCastle.
+     *
+     * Exercises two cross-provider scenarios, each with a large input backing array where only a
+     * slice is passed via inputOffset + inputLen, and the output buffer is exactly the right size
+     * for that slice.
+     *
+     * Scenario A: OpenJCEPlus encrypts a slice → BC decrypts → plaintext verified.
+     * Scenario B: BC encrypts → OpenJCEPlus decrypts a slice → plaintext verified.
+     */
+    @Test
+    public void testAESCCMDoFinalWithInputSliceInterop() throws Exception {
+        // Fixed parameters for deterministic, readable test.
+        int ccmTagLengthBits = 128;
+        int tagLenInBytes    = ccmTagLengthBits / 8; // 16
+        byte[] iv = {(byte) 0x01, (byte) 0x02, (byte) 0x03, (byte) 0x04,
+                     (byte) 0x05, (byte) 0x06, (byte) 0x07, (byte) 0x08,
+                     (byte) 0x09, (byte) 0x0a, (byte) 0x0b, (byte) 0x0c,
+                     (byte) 0x0d};
+        byte[] aadBytes = {(byte) 0x09, (byte) 0x10, (byte) 0x11, (byte) 0x12,
+                           (byte) 0x13, (byte) 0x14, (byte) 0x15, (byte) 0x16};
+
+        // Generate a shared 128-bit AES key via OpenJCEPlus.
+        KeyGenerator keyGenerator = KeyGenerator.getInstance("AES", getProviderName());
+        keyGenerator.init(128);
+        SecretKey key = keyGenerator.generateKey();
+        byte[] aesKeyBytes = key.getEncoded();
+        SecretKeySpec keySpec = new SecretKeySpec(aesKeyBytes, "AES");
+
+        // The actual plaintext payload is 20 bytes, embedded inside a 100-byte backing array
+        // at offset 40. This exercises the fix for issue #1564 where engineDoFinal incorrectly
+        // used input.length instead of inputLen for the output buffer size check.
+        // See: https://github.com/IBM/OpenJCEPlus/issues/1564 for more information on failing scenarios.
+        byte[] plaintext = "InteropSliceTest1234".getBytes(); // exactly 20 bytes
+        int inputOffset = 40;
+        int inputLen    = plaintext.length; // 20
+        byte[] inputBacking = new byte[100];
+        System.arraycopy(plaintext, 0, inputBacking, inputOffset, inputLen);
+
+        // -----------------------------------------------------------------------
+        // Scenario A: OpenJCEPlus encrypts the slice → BC decrypts → verify
+        // -----------------------------------------------------------------------
+        ibm.security.internal.spec.CCMParameterSpec ccmParamEnc =
+                new ibm.security.internal.spec.CCMParameterSpec(ccmTagLengthBits, iv);
+        Cipher encCipherOJP = Cipher.getInstance("AES/CCM/NoPadding", getProviderName());
+        encCipherOJP.init(Cipher.ENCRYPT_MODE, keySpec, ccmParamEnc);
+        encCipherOJP.updateAAD(aadBytes);
+
+        // Output buffer sized exactly for the slice: inputLen + tagLenInBytes = 36 bytes.
+        // Before the fix for issue #1564, this would throw a ShortBufferException
+        // because the check incorrectly used input.length (100) instead of inputLen (20).
+        // See: https://github.com/IBM/OpenJCEPlus/issues/1564 for more information on failing scenarios.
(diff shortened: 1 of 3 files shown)


johnpeck-us-ibm (approved):

JinhangZhang (approved):
LGTM

taoliult (approved):
LGTM

src/test/java/ibm/jceplus/junit/base/BaseTestAESCCMInteropBC.java
```diff
@@ -289,6 +289,123 @@ public void testAESCCM() throws Exception {
...
+
+        // Generate a shared 128-bit AES key via OpenJCEPlus.
+        KeyGenerator keyGenerator = KeyGenerator.getInstance("AES", getProviderName());
+        keyGenerator.init(128);
+        SecretKey key = keyGenerator.generateKey();
+        byte[] aesKeyBytes = key.getEncoded();
+        SecretKeySpec keySpec = new SecretKeySpec(aesKeyBytes, "AES");
+
+        // The actual plaintext payload is 20 bytes, embedded inside a 100-byte backing array
+        // at offset 40 — so inputLen (20) << input.length (100).
```
KostasTsiounis: Should we refer to the problematic scenario that we are fixing in the test? Reading this on its own in the future might not make that much sense if one doesn't have context.

src/test/java/ibm/jceplus/junit/base/BaseTestAESCCMInteropBC.java
```diff
@@ -289,6 +289,123 @@ public void testAESCCM() throws Exception {
...
+        // Scenario A: OpenJCEPlus encrypts the slice → BC decrypts → verify
+        // -----------------------------------------------------------------------
+        ibm.security.internal.spec.CCMParameterSpec ccmParamEnc =
+                new ibm.security.internal.spec.CCMParameterSpec(ccmTagLengthBits, iv);
+        Cipher encCipherOJP = Cipher.getInstance("AES/CCM/NoPadding", getProviderName());
+        encCipherOJP.init(Cipher.ENCRYPT_MODE, keySpec, ccmParamEnc);
+        encCipherOJP.updateAAD(aadBytes);
+
+        // Output buffer sized exactly for the slice: inputLen + tagLenInBytes = 36 bytes.
+        // input.length + tagLenInBytes would be 116, which is larger than the output buffer.
```
KostasTsiounis: Similar comment here.

src/test/java/ibm/jceplus/junit/base/BaseTestAESCCMInteropBC.java
```diff
@@ -289,6 +289,123 @@ public void testAESCCM() throws Exception {
...
+        System.arraycopy(ciphertextB, 0, cipherBacking, cipherInputOffset, cipherInputLen);
+
+        ibm.security.internal.spec.CCMParameterSpec ccmParamDec =
+                new ibm.security.internal.spec.CCMParameterSpec(ccmTagLengthBits, iv);
+        Cipher decCipherOJP = Cipher.getInstance("AES/CCM/NoPadding", getProviderName());
+        decCipherOJP.init(Cipher.DECRYPT_MODE, keySpec, ccmParamDec);
+        decCipherOJP.updateAAD(aadBytes);
+
+        // Output buffer sized exactly for the decrypted plaintext: cipherInputLen - tagLenInBytes = 20.
+        // input.length - tagLenInBytes would be 84, which is larger than the output buffer.
```
KostasTsiounis: Same here.

src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
```diff
@@ -288,6 +288,196 @@ public void testAESCCM() throws Exception {
...
+        int inputLen    = 10;
+        byte[] input    = new byte[100]; // input.length >> inputLen
+        new java.security.SecureRandom().nextBytes(input);
+
+        CCMParameterSpec ccmParameterSpec = new CCMParameterSpec(ccmTagLengthBits, iv);
+        Cipher cipher = Cipher.getInstance("AES/CCM/NoPadding", getProviderName());
+        cipher.init(Cipher.ENCRYPT_MODE, keySpec, ccmParameterSpec);
+
+        // The output buffer is exactly the right size for the slice being encrypted:
+        // inputLen + tagLenInBytes = 26 bytes. input.length + tagLenInBytes would be 116.
```
KostasTsiounis: Again, it's not clear as to why we're mentioning this if you see it on its own.

jasonkatonica: Added additional comments to help describe and refer to the original issue that might help future readers.

src/test/java/ibm/jceplus/junit/base/BaseTestAESCCMInteropBC.java
jasonkatonica: Added additional comments to help describe and refer to the original issue that might help future readers.

src/test/java/ibm/jceplus/junit/base/BaseTestAESCCMInteropBC.java
jasonkatonica: Added additional comments to help describe and refer to the original issue that might help future readers.

src/test/java/ibm/jceplus/junit/base/BaseTestAESCCMInteropBC.java
jasonkatonica: Added additional comments to help describe and refer to the original issue that might help future readers.

KostasTsiounis (approved):
LGTM

PR #1580, opened 2026-07-08
Set Default SecureRandom Algorithm 

### LINKED ISSUE #1553: Inconsistent SecureRandom registration and usage
Based on the actual service registration, `SHA512DRBG` is registered **BEFORE** `SHA256DRBG` in OpenJCEPlus Java 25, making it the default `SecureRandom` algorithm.

```java
import java.security.SecureRandom;

public class TestSecureRandom {

        SecureRandom secureRandom = new SecureRandom();
        System.out.println("Provider: " + secureRandom.getProvider());
        System.out.println("Default Algorithm: " + secureRandom.getAlgorithm());
    }
}
```
```
Provider: OpenJCEPlus version 25
Default Algorithm: SHA512DRBG
```
The full registration service list: https://github.com/IBM/OpenJCEPlus/issues/1553#issuecomment-4785411321
That output was collected from `putService(Service s)` method in `java/security/Provider.java`.

That is causing this JTREG: `java/security/SecureRandom/DefaultAlgo.java` fail.

In addition, that is inconsistent with what OpenJCEPlus provides to other algorithms:
https://github.com/IBM/OpenJCEPlus/blob/deb675c3e94f0afb22b74a40e9f5cb1aab36d615/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlus.java#L218

```java
// Get SecureRandom to use for crypto operations.
    //
    java.security.SecureRandom getSecureRandom(java.security.SecureRandom userSecureRandom) {
        try {
            return java.security.SecureRandom.getInstance("SHA256DRBG", this);
        } catch (NoSuchAlgorithmException e) {
            throw new ProviderException("SecureRandom not available");
        }
    }
```

Does the new service registration preserve the orders?

@jasonkatonica

johnpeck-us-ibm:
Add code to be able to set the Default SecureRandom algorithm from provider.

fixes: #1553

### DIFF
--- a/PROVIDER_CONFIG_FORMAT.md
+++ b/PROVIDER_CONFIG_FORMAT.md
@@ -35,6 +35,7 @@ default = <true|false|1|0>
 - **name**: Unique identifier for the provider (required)
 - **description**: Human-readable description of the provider (required)
 - **default**: Whether to load default provider attributes (optional, values: `true`, `false`, `1`, `0`)
+- **securerandomdefault**: Determines which Secure Random Algorithm is the default one if this provider is first in the provider list
 
 #### Examples:
 ```properties
@@ -48,6 +49,9 @@ description = OpenJCEPlusFIPS-test Provider
 default = true
 ```
 
+# Set Default Secure Random algorithm
+securerandomdefault = SHA256DRBG
+```properties
 #### Parser Behavior:
 - Keys with exactly 1 part are checked for `name`, `description`, or `default`
 - Invalid single-part keys throw an `IOException`
@@ -60,7 +64,6 @@ default = true
 Register cryptographic service implementations.
 
 #### Format:
-```properties
 Service.<ServiceType>.<AlgorithmName> = <ImplementationClassName>
 ```
 
@@ -114,7 +117,7 @@ Service.SecretKeyFactory.PBKDF2WithHmacSHA256 = com.ibm.crypto.plus.provider.PBK
 - **KeyFactory**: Key conversion and specification
 - **KeyGenerator**: Symmetric key generators
 - **KeyPairGenerator**: Asymmetric key pair generators
-- **MAC**: Message Authentication Code engines
+- **Mac**: Message Authentication Code engines
 - **MessageDigest**: Hash/digest algorithms
 - **SecretKeyFactory**: Secret key factories
 - **SecureRandom**: Random number generators
--- a/src/main/java/com/ibm/crypto/plus/provider/DefaultProviderAttrs.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/DefaultProviderAttrs.java
@@ -371,6 +371,7 @@ class DefaultProviderAttrs {
         + " # SecureRandom\n"
         + " # =======================================================================\n"
         + " #\n"
+        + "securerandomdefault = SHA256DRBG\n"
         + "SecureRandom.SHA256DRBG.alias.add = HASHDRBG, SHA2DRBG\n"
         + "SecureRandom.SHA256DRBG.attr.add.ThreadSafe = true\n"
         + "Service.SecureRandom.SHA256DRBG = com.ibm.crypto.plus.provider.HASHDRBG$SHA256DRBG\n"
--- a/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
@@ -201,6 +201,10 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
     // SecureRandom in FIPS mode.
     //
     java.security.SecureRandom getSecureRandom(java.security.SecureRandom userSecureRandom) {
+        /* Unlike OpenJCEPlus we do not want this algorithm set by the config file 
+         * this should only be updated manually in the provider to make sure we 
+         * are using the RNG that FIPS requires.
+         */
         try {
             return java.security.SecureRandom.getInstance("SHA256DRBG", this);
         } catch (NoSuchAlgorithmException e) {
--- a/src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
@@ -14,6 +14,8 @@ import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -36,6 +38,7 @@ public class ProviderServiceReader {
     private Set<String> setDefAttributes = null;
     private Properties defPr = null;
     private boolean def = false;
+    private String secureRandDef = null;
     private static Debug debug = Debug.getInstance("jceplus");
     
     /**
@@ -125,9 +128,9 @@ public class ProviderServiceReader {
      */
     public List<ServiceDefinition> readServices() throws IOException {
         List<ServiceDefinition> services = new ArrayList<>();
-        Set<String> setAliases = new HashSet<>();
-        Set<String> setAttributes = new HashSet<>();
-        Set<String> setServices = new HashSet<>();
+        Set<String> setAliases = new LinkedHashSet<>();
+        Set<String> setAttributes = new LinkedHashSet<>();
+        Set<String> setServices = new LinkedHashSet<>();
         BufferedReader rd = null;
         Properties pr = new Properties();
 
@@ -164,6 +167,8 @@ public class ProviderServiceReader {
                     description = pr.getProperty(key);
                 } else if (parts.length == 1 && parts[0].equalsIgnoreCase("default")) {
                     defaults = pr.getProperty(key);
+                } else if (parts.length == 1 && parts[0].equalsIgnoreCase("securerandomdefault")) {
+                    secureRandDef = pr.getProperty(key);
                 } else {
                     throw new IOException("Invalid key: " + key);
                 }
@@ -217,7 +222,7 @@ public class ProviderServiceReader {
             throw new IOException("File issue: " + e.getMessage());
         }
        
-        return services;
+        return orderServices(services);
     }
        
     /**
@@ -301,7 +306,7 @@ public class ProviderServiceReader {
      * @return a map of attribute key-value pairs
      */
     private Map<String, String> processAttributes(String[] parts, Set<String> configAttrs, Properties defaultPr, Properties configPr) {
-        Map<String, String> attributes = new HashMap<>();
+        Map<String, String> attributes = new LinkedHashMap<>();
         String search = parts[1] + "." + parts[2] + ".attr";
 
         //Only .add, and .delete are supported for Attributes
@@ -365,7 +370,36 @@ public class ProviderServiceReader {
         }
         return attributes;
     }
-    
+
+    /**
+     * Orders services so that the SecureRandom service whose algorithm matches
+     * {@code secureRandDef} is placed first, with all other services following
+     * in their original order.
+     *
+     * @param services the list of services to order
+     * @return a new list with the designated SecureRandom service placed first
+     */
+    public List<ServiceDefinition> orderServices(List<ServiceDefinition> services) {
+        List<ServiceDefinition> ordered = new ArrayList<>();
+        ServiceDefinition secureRandom = null;
+
+        for (ServiceDefinition service : services) {
+            if (secureRandDef != null
+                    && service.getType().equalsIgnoreCase("SecureRandom")
+                    && service.getAlgorithm().equalsIgnoreCase(secureRandDef)) {
+                secureRandom = service;
+            } else {
+                ordered.add(service);
+            }
+        }
+
+        if (secureRandom != null) {
+            ordered.add(0, secureRandom);
+        }
+
+        return ordered;
+    }
+
     /**
      * Filters services by type.
      * 
@@ -432,5 +466,14 @@ public class ProviderServiceReader {
      */
     public String getDesc() {
         return description;
-    }     
+    }
+
+    /**
+     * Gets the default Secure Random algorithm that was read in from the config file.
+     * 
+     * @return a String that contains the default Secure Random algorithm
+     */
+    public String getDefSecRnd() {
+        return secureRandDef;
+    }
 }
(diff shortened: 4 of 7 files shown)


src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -166,6 +166,10 @@ public boolean isFIPS() {
     // SecureRandom in FIPS mode.
     //
     java.security.SecureRandom getSecureRandom(java.security.SecureRandom userSecureRandom) {
+        /* Unlike OpenJCEPlus we do not want this algorithm set by the config file
```
jasonkatonica, line 169: Any reason to restrict this? Sure we want the default to be specific for FIPS at the moment but in the future users could bring their own fips certified random libraries and backends. Having this behave the same for both providers similar to other algorithms in the fips provider seems more strait forward.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
```diff
@@ -34,8 +35,6 @@
 
 @Tag(Tags.OPENJCEPLUS_NAME)
 @Tag(Tags.OPENJCEPLUS_FIPS_NAME)
-@Tag(Tags.OPENJCEPLUS_MULTITHREAD_NAME)
-@Tag(Tags.OPENJCEPLUS_FIPS_MULTITHREAD_NAME)
```
jasonkatonica, line 38: Any reason we needed to disable the entire file instead of synchronizing just the one new random test below due to provider insertion/deletion?

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
```diff
@@ -232,6 +231,56 @@ public void testDefServicesDelAttribute() throws Exception {
         System.out.println("TestAttr2: " + attr2);
     }
 
+    @Test
```
jasonkatonica, line 234: Can the test also tolerate OpenJCEPlusFIPS to ensure we are getting the random we expect? It does not seem the test handles FIPS expecations.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -368,7 +373,36 @@ private Map<String, String> processAttributes(String[] parts, Set<String> config
...
+
+    /**
+     * Orders services so that the SecureRandom service whose algorithm matches
+     * {@code secureRandDef} is placed first, with all other services following
+     * in their original order.
+     *
+     * @param services the list of services to order
+     * @return a new list with the designated SecureRandom service placed first
+     */
+    public List<ServiceDefinition> orderServices(List<ServiceDefinition> services) {
```
jasonkatonica, line 385: Scope here can be `private` ?

PROVIDER_CONFIG_FORMAT.md
```diff
@@ -48,6 +49,9 @@ description = OpenJCEPlusFIPS-test Provider
 default = true
 ```
 
+# Set Default Secure Random algorihtm
```
jasonkatonica: typo here should be `algorithm` instead of `algorihtm`.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
johnpeck-us-ibm: The issue with this being changeable is that it could be set to something that we do not want it set too via a config parameter. It is just a safe guard to make sure that this does not change unintentionally.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
johnpeck-us-ibm: I tried syncing the test and it does not  work. Besides this I would not expect provider creations to be multi threaded. I am not sure the Java is set up to allow provider creation in that manner anyway.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: Not if we want to use it in the tests which we do today.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
johnpeck-us-ibm: I will add a test to verify. I can not do what I did in this test, because the FIPS provider is not be initialized using a config file like OpenJCEPlus.

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: Updated.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
KostasTsiounis: I think that's ok given that we also don't provide a `configure()` method through the `OpenJCEPlusFIPS` provider. So, if anyone wanted to use their own attributes they would still have to use `OpenJCEPlus` that allows this to be user-specified.

src/main/java/com/ibm/crypto/plus/provider/DefaultProviderAttrs.java
```diff
@@ -363,6 +363,7 @@ class DefaultProviderAttrs {
         + " # SecureRandom\n"
         + " # =======================================================================\n"
         + " #\n"
+        + "securerandomdefault = SHA256DRBG\n"
```
KostasTsiounis, line 366: Should we follow a pattern more similar to the existing ones? Like `SecureRandom.default = SHA256DRBG`.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
KostasTsiounis, line 385: Is there a chance that someone will expect that the algorithm to be used by default is the one they actually specified first in the config file, instead of the one indicated with the specific property?

Is that something we'd want to allow if that property hasn't be set? And if not, we should probably document it.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
```diff
@@ -232,6 +231,84 @@ public void testDefServicesDelAttribute() throws Exception {
...
+            + "default = true\n"
+            + "securerandomdefault = SHA512DRBG";
+
+        SecureRandom secureRandom = new SecureRandom();
+
+        List<String> acceptableValues = List.of("NativePRNG", "NativePRNGBlocking", "NativePRNGNonBlocking", "DRBG");
+        assertTrue (acceptableValues.contains(secureRandom.getAlgorithm()), "SecureRandom not SunJCE - " + secureRandom.getAlgorithm());
+
+        //Now Put OpenJCEPlus first in the list.
+        Provider provider = (Provider) Class.forName("com.ibm.crypto.plus.provider.OpenJCEPlus").getDeclaredConstructor().newInstance();
```
KostasTsiounis, line 249: Why are we using reflection here?

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
```diff
@@ -232,6 +231,84 @@ public void testDefServicesDelAttribute() throws Exception {
...
+    // This test does not play well in multi-thread. So, do not include this in those tests.
+    public void testDefaultSecureRandomFIPS() throws Exception {
+
+        SecureRandom secureRandom = new SecureRandom();
+
+        List<String> acceptableValues = List.of("NativePRNG", "NativePRNGBlocking", "NativePRNGNonBlocking", "DRBG");
+        assertTrue (acceptableValues.contains(secureRandom.getAlgorithm()), "SecureRandom not SunJCE");
+
+        //Now Put OpenJCEPlus first in the list.
+        Provider provider = (Provider) Class.forName("com.ibm.crypto.plus.provider.OpenJCEPlusFIPS").getDeclaredConstructor().newInstance();
```
KostasTsiounis, line 294: Similar comment here.

PROVIDER_CONFIG_FORMAT.md
```diff
@@ -48,6 +49,9 @@ description = OpenJCEPlusFIPS-test Provider
 default = true
 ```
 
+# Set Default Secure Random algorithm
```
KostasTsiounis, line 52: Do we also need a ```properties here as well?

PROVIDER_CONFIG_FORMAT.md
```diff
@@ -35,6 +35,7 @@ default = <true|false|1|0>
 - **name**: Unique identifier for the provider (required)
 - **description**: Human-readable description of the provider (required)
 - **default**: Whether to load default provider attributes (optional, values: `true`, `false`, `1`, `0`)
+- **securerandomdefault**: Determines which Secure Random Algorithm is the default one if this provider is first in the provider list
```
KostasTsiounis, line 38: Should we also document that if it's not set, it defaults to a value? That is if we decide that we won't honour user-specifies order if it's not.

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
jasonkatonica: OK makes sense to me.

src/main/java/com/ibm/crypto/plus/provider/DefaultProviderAttrs.java
johnpeck-us-ibm: I did this way since this was a general parameter like name, and description.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
johnpeck-us-ibm: The first in the list behavior is not something that is documented anywhere that I could find. Unfortunately, with out this the default Secure Random Alg is random. In my testing I have seen it switch back and forth between the two we define.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
johnpeck-us-ibm: We don't have too, It is just one of a few ways this can be done.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
johnpeck-us-ibm: We don't have too, It is just one of a few ways this can be done.

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: Only if the configuration is not a propertyconfig file like in ProviderDefAttrs.config but is stored in a variable like in DefaultProviderAttrs.java

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: We can not determine the one that will be selected. It is random. So, not sure what to say here.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
jasonkatonica: Sure makes sense this can be in another location. A test validating that FIPS is correct just protects us from accidentally returning the wrong algorithm for the FIPS paths in the future even though we dont use a config file currently.

johnpeck-us-ibm: Not sure what you are asking fro. I added 'testDefaultSecureRandomFIPS' which validates the behavior of 'new SecureRandom'. Our code in OpenJCE uses provider.getSecureRandom() and in there the Algorithm is hard coded and not affected by the config parameter.

jasonkatonica: OK i see the test now `testDefaultSecureRandomFIPS` that you added which asserts that we get the correct random instance back. That looks good to me.

jasonkatonica (approved):
LGTM

taoliult (approved):
LGTM

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
KostasTsiounis: Wouldn't it be simpler to do `Provider provider = new com.ibm.crypto.plus.provider.OpenJCEPlus();`

PROVIDER_CONFIG_FORMAT.md
KostasTsiounis: No, I'm talking about the way the `md` file is shown. I think we need a ```properties here. Otherwise it doesn't render properly.

PROVIDER_CONFIG_FORMAT.md
KostasTsiounis: If I'm getting this right, when someone doesn't specify this in the configuration, this defaults to `SHA256DRBG`. Should we mention this in this `md` file?

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
KostasTsiounis: No, I didn't say remove this. It was more of a question for potential additional functionality if we want to or can even do it. Whether we could honour someone's order if the default wasn't set. Of course that might not even be possible.

johnpeck-us-ibm: We can not honor the first item they set as the default if the default is not called out. I have seen this switch back ans forth randomly. We could but we would have to move away from using the Properties class. 

Note: There is nothing in any documentation that talks about the first one defined is the default in any documentation. I see this only in the comments in the Oracle code.

src/test/java/ibm/jceplus/junit/tests/TestProviderServices.java
johnpeck-us-ibm: Ether way works. This does follow how we do it in our test code though.

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: As indicated in the comment with out this setting the default could be either SHA256DRBG or SHA512DRBG what is selected is random. It could be either one.

PROVIDER_CONFIG_FORMAT.md
johnpeck-us-ibm: Fixed.

JinhangZhang (approved):
LGTM

KostasTsiounis (approved):
LGTM

jasonkatonica (commented):
LGTM

PR #1587, opened 2026-07-08
Fix AESCCMCipher.engineDoFinal to use inputLen

jasonkatonica:
AESCCMCipher.engineDoFinal was using input.length (the full backing array length) instead of inputLen (the caller-specified byte count) when deciding whether the output buffer was large enough. This caused a false ShortBufferException whenever the caller passed a slice of a larger backing array via inputOffset + inputLen, because the backing array was wider than the actual data being processed.

Fix: replace input.length with inputLen in both the encrypt and decrypt output-size checks inside engineDoFinal.

Tests were also added to exercise slicing and ensure interop via sliced buffers is working as expected.

Fixes: https://github.com/IBM/OpenJCEPlus/issues/1564

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1565


johnpeck-us-ibm (approved):

KostasTsiounis (approved):
LGTM

PR #1593, opened 2026-07-09
Fix missing return on failure in HMAC_init

jasonkatonica:
When GetPrimitiveArrayCritical returned NULL for keyNative, the code threw an OCK exception but fell through to call HMAC_init_internal with a NULL keyNative pointer, leading to a potential invalid memory access.

Add an early return of FAIL_HMAC_INTERNAL_INIT after the exception is thrown, and add the missing gslogFunctionExit call inside the debug block to keep exit logging consistent with other error paths.

Fixes #1253


johnpeck-us-ibm (approved):

KostasTsiounis (approved):
LGTM

PR #1617, opened 2026-07-13
Use ICC_EVP_PKEY_size for XDH key agreement buffer sizing

taoliult:
Replace ICC_EVP_PKEY_derive size query with ICC_EVP_PKEY_size call to determine buffer size before deriving shared secret. Removes platform-specific workarounds and simplifies the implementation.

### DIFF
--- a/src/main/native/ock/ECKey.c
+++ b/src/main/native/ock/ECKey.c
@@ -2274,24 +2274,24 @@ Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_ECKEY_1computeECDH
 /*
  * Class:     com_ibm_crypto_plus_provider_ock_NativeOCKImplementation
  * Method:    XECKEY_computeECDHSecret
- * Signature: (JJJJI)[B
+ * Signature: (JJJJ)[B
  */
 JNIEXPORT jbyteArray JNICALL
 Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_XECKEY_1computeECDHSecret(
     JNIEnv *env, jclass thisObj, jlong ockContextId, jlong genCtx,
-    jlong pubXecKeyId, jlong privXecKeyId, jint secretBufferSize) {
+    jlong pubXecKeyId, jlong privXecKeyId) {
     static const char *functionName =
         "NativeInterface_XECKEY_1computeECDHSecret";
 
-    ICC_CTX            *ockCtx       = (ICC_CTX *)((intptr_t)ockContextId);
-    ICC_EVP_PKEY       *ockPubXecKey = (ICC_EVP_PKEY *)((intptr_t)pubXecKeyId);
-    const ICC_EVP_PKEY *ockPrivXecKey =
+    ICC_CTX            *ockCtx          = (ICC_CTX *)((intptr_t)ockContextId);
+    ICC_EVP_PKEY       *ockPubXecKey    = (ICC_EVP_PKEY *)((intptr_t)pubXecKeyId);
+    const ICC_EVP_PKEY *ockPrivXecKey   =
         (const ICC_EVP_PKEY *)((intptr_t)privXecKeyId);
     ICC_EVP_PKEY_CTX *gen_ctx           = NULL;
     jbyteArray        secretBytes       = NULL;
     unsigned char    *secretBytesNative = NULL;
     jboolean          isCopy            = 0;
-    size_t            required_len      = 0;
+    int               required_len      = 0;
     size_t            secret_key_len    = 0;
     int               rc                = 0;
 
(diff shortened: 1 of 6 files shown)


src/main/native/ock/ECKey.c
```diff
@@ -2318,32 +2317,17 @@ Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_XECKEY_1computeECD
...
-        if (ICC_OSSL_SUCCESS != rc) {
-            throwOCKException(env, 0,
-                    "ICC_EVP_PKEY_derive failed to get secret size");
-            goto cleanup;
-        }
-        if (required_len == 0) {
-            throwOCKException(env, 0, "Derived secret size is zero");
-            goto cleanup;
-        }
+    /* NEW APPROACH:
```
KostasTsiounis: Do we need to say this is a `NEW APPROACH`?

src/main/native/ock/ECKey.c
```diff
@@ -2318,32 +2317,17 @@ Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_XECKEY_1computeECD
...
+    secret_key_len = ICC_EVP_PKEY_size(ockCtx, ockPubXecKey);
+
+    if (secret_key_len <= 0) {
+        throwOCKException(env, 0, "Derived secret size is zero");
+        goto cleanup;
     }
-    secret_key_len = required_len;
 
-    secretBytes = (*env)->NewByteArray(env, secret_key_len); /* Create Java secret bytes array */
+    secretBytes = (*env)->NewByteArray(env, (jsize)secret_key_len); /* Create Java secret bytes array */
```
KostasTsiounis: I think a space is required after `(jsize)`.

KostasTsiounis (changes requested):

src/main/native/ock/ECKey.c
taoliult: Removed.

src/main/native/ock/ECKey.c
taoliult: Yes, added.

src/main/native/ock/ECKey.c
```diff
@@ -2291,7 +2291,6 @@ Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_XECKEY_1computeECD
     jbyteArray        secretBytes       = NULL;
     unsigned char    *secretBytesNative = NULL;
     jboolean          isCopy            = 0;
-    size_t            required_len      = 0;
     size_t            secret_key_len    = 0;
```
jasonkatonica, line 2295: [EVP_PKEY_get_size](https://docs.openssl.org/3.0/man3/EVP_PKEY_get_size/) returns an `int` and OCKC should be similar, should we declare it that way?

src/main/native/ock/ECKey.c
jasonkatonica: Method signature should be updated due to the last parameter being removed.

src/main/native/ock/ECKey.c
```diff
@@ -2318,32 +2317,15 @@ Java_com_ibm_crypto_plus_provider_ock_NativeOCKImplementation_XECKEY_1computeECD
...
-            throwOCKException(env, 0,
-                    "ICC_EVP_PKEY_derive failed to get secret size");
-            goto cleanup;
-        }
-        if (required_len == 0) {
-            throwOCKException(env, 0, "Derived secret size is zero");
-            goto cleanup;
-        }
+    /* Determine buffer size using ICC_EVP_PKEY_size with the public key */
+    secret_key_len = ICC_EVP_PKEY_size(ockCtx, ockPubXecKey);
```
jasonkatonica: Reading [ICC_EVP_PKEY_size](https://docs.openssl.org/3.0/man3/EVP_PKEY_get_size/) I am a bit worried that this API returns more memory then really represents the key we are about to derive. The doc there states:
```
It must be stressed that, unless the documentation for the operation that's being performed says otherwise, the size returned by EVP_PKEY_get_size() is only preliminary and not exact, so the final contents of the target buffer may be smaller. 
```
Perhaps we also have to make a call to `EVP_PKEY_get_bits` as mentioned in the same doc to ensure our buffer is not too big? Worry here is that if the buffer is larger then necessary as documented then we may return some trailing random bytes in the buffer as our derived key.

jasonkatonica (changes requested):

src/main/native/ock/ECKey.c
jasonkatonica: I also see that `ICC_EVP_PKEY_derive(ockCtx, gen_ctx, secretBytesNative, &secret_key_len);` would update `secret_key_len` with the actual number of byte written. Perhaps we can just ensure that this length matches what we expect there ( The length of the buffer we allocated as indicated by `ICC_EVP_PKEY_size` )

No need to use `EVP_PKEY_get_bits` in this approach.

src/main/native/ock/ECKey.c
taoliult: Yes, `ICC_EVP_PKEY_size` returns `int`, so `required_len` of type int is introduced to match its return type. The existing `secret_key_len` is kept as `size_t` because `ICC_EVP_PKEY_derive` requires a `size_t *` for its length parameter. `required_len` is cast to `size_t` and assigned to `secret_key_len` after the `<= 0` error check passes.

src/main/native/ock/ECKey.c
taoliult: Yes, updated.

src/main/native/ock/ECKey.c
taoliult: Agree. Since `ICC_EVP_PKEY_size` returns a preliminary upper-bound size and `ICC_EVP_PKEY_derive` updates `secret_key_len` with the actual number of bytes written, the allocated buffer may be larger than the real derived secret. 

To avoid returning trailing random bytes, after `ICC_EVP_PKEY_derive` we compare `required_len` (the allocated size) against `secret_key_len` (the actual written size). If they differ, we allocate a new correctly-sized Java array `trimmedBytes`, pin `secretBytes` with `GetPrimitiveArrayCritical` to get a read-only native pointer `trimSrc`, copy only the actual bytes into `trimmedBytes` via `SetByteArrayRegion`, then release the pin with `JNI_ABORT` (read-only, no write-back needed), delete the oversized array, and return `trimmedBytes` instead.

jasonkatonica: An alternative also could be to simply fail if less bytes then the size of the buffer is returned. I am unsure how often that would really happen or not.

taoliult: Updated,  will `throwOCKException` if `(secret_key_len != (size_t)required_len)`.

src/main/native/ock/ECKey.c
jasonkatonica: This has changed to use an `int` and looks good to me.

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

JinhangZhang (approved):
LGTM

johnpeck-us-ibm (approved):

PR #1625, opened 2026-07-15
Update test tags and method sources

KostasTsiounis:
Methods within the `TestArguments` class are utilized through the `MethodSource` annotation to parameterize class attributes (e.g., provider and key sizes).

This change allows the passing of parameters to these methods. The parameters pertain to information about the calling class, like defined tags or other static field, like lists of key sizes.

The test tags are, also, updated to only include specific providers and a dedicated tag for multithreaded tests. The top level tests are altered to look for tag expressions that indicate provider and the multithreaded tag when looking for appropriate tests.

### DIFF
--- a/src/test/java/ibm/jceplus/junit/tests/TestArguments.java
+++ b/src/test/java/ibm/jceplus/junit/tests/TestArguments.java
@@ -10,6 +10,8 @@ package ibm.jceplus.junit.tests;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.junit.jupiter.params.provider.Arguments;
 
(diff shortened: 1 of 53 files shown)


johnpeck-us-ibm (approved):

src/test/java/ibm/jceplus/junit/tests/parameters/resolvers/ProviderListParameterResolver.java
```diff
@@ -0,0 +1,36 @@
...
+        Type type = parameterContext.getParameter().getParameterizedType();
+        if (!(type instanceof ParameterizedType pt)) return false;
+        return pt.getRawType() == Set.class &&
+                pt.getActualTypeArguments()[0] == String.class;
+    }
+
+    @Override
+    public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
+        Set<String> providers = extensionContext.getTags().stream()
+                                        .filter(pn -> !pn.equalsIgnoreCase("multithread"))
```
jasonkatonica: Should this use the new enumeration for multithread instead in Tags.java?

jasonkatonica (changes requested):
Also appears that a different number of tests are running due to these changes. Perhaps this is intentional?

Perhaps the exclusion of multithreaded tests from TestAll can account for the difference.

KostasTsiounis:
> Also appears that a different number of tests are running due to these changes. Perhaps this is intentional?
> 
> Perhaps the exclusion of multithreaded tests from TestAll can account for the difference.

After checking the logs, I found that that's exactly what happens. Not running the `TestRSAMultiThread` as part of the single threaded `TestAll` accounts for this difference.

src/test/java/ibm/jceplus/junit/tests/TestArguments.java
```diff
@@ -63,81 +44,53 @@ protected static Stream<Arguments> getOpenJCEPlusWithSUNInteropProvider() {
...
         } else {
+            enabledProviders = new ArrayList<>();
             for (String tag : groupPropertyTags) {
-                if (TestProvider.OpenJCEPlus.getProviderName().equalsIgnoreCase(tag)) {
-                    enabledProviders.add(TestProvider.OpenJCEPlus);
-                } else if (TestProvider.OpenJCEPlusFIPS.getProviderName().equalsIgnoreCase(tag)) {
-                    enabledProviders.add(TestProvider.OpenJCEPlusFIPS);
+                try {
+                    TestProvider tp = TestProvider.valueOf(tag);
+                    if (taggedProviders.contains(tp)) {
```
Mohit-Rajbhar100698, line 89: One doubt here 
What if a test is suppose to be run only for OpenJCEPlus(tagged with OpenJCEPlus) and we pass “OpenJCEPLusFIPS” in -DGroups. 
Then enabledProviders can be null.

As “getEnabledProviders” method is used as method source for the tests which are OpenJCEplus test only.

Mohit-Rajbhar100698:
And Requires Rebase from main. 
Some Tests like (TestPBMAC1Interop) using (getOpenJCEPlusOnly) methodsource

src/test/java/ibm/jceplus/junit/tests/TestArguments.java
Mohit-Rajbhar100698: Tested it 
when "-Dgroups=OpenJCEPlusFIPS" any tests tagged only with `OpenJCEPlus` are already skipped during test filteration, then getEnabledProviders() never execute for them.

this is good. not a issue

src/test/java/ibm/jceplus/junit/tests/parameters/resolvers/ProviderListParameterResolver.java
KostasTsiounis: Switched to use `Tags.MULTITHREAD_NAME`.

jasonkatonica (approved):
LGTM

PR #1647, opened 2026-07-16
Validate expanded PQC private key encodings

JinhangZhang:
Update PQCPrivateKey to validate ML-DSA and ML-KEM private key material using the RFC 9881 / RFC 9935 expandedKey CHOICE encoding.

The validation checks that the private key material contained in the PKCS#8 privateKey OCTET STRING is encoded as a DER OCTET STRING with the expected expanded key length for the corresponding parameter set. Unsupported private key encodings are rejected with an InvalidKeyException.

Backported from: https://github.com/IBM/OpenJCEPlus/pull/1520

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,6 +532,8 @@ OpenJCEPlus provider enhances the security of Java applications by providing an 
 
 No keytool or certificate support was added other than what is already in a given Java runtime environment.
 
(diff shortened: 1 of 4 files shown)


johnpeck-us-ibm (approved):

KostasTsiounis (approved):
LGTM

taoliult (approved):

jasonkatonica (approved):
LGTM

PR #1665, opened 2026-07-21
Add Support to Select native Crypto Provider

johnpeck-us-ibm:
Code was added to allow the use of  a Provider.Service attrobute called "NativeProvider" to specify the native back end to use. Only OCK and OPENSSL are support. Any thing specified in the "NativeProvider" attribute other than OPENSSL will result in OCK being used.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1512

Signed-off-by: johnpeck-us-ibm [johnpeck@us.ibm.com](mailto:johnpeck@us.ibm.com)


taoliult (approved):

src/main/java/com/ibm/crypto/plus/provider/RSASignature.java
```diff
@@ -35,7 +35,7 @@ abstract class RSASignature extends SignatureSpi {
         try {
             this.provider = provider;
             this.ockDigestAlgo = ockDigestAlgo;
-            this.signature = Signature.getInstance(ockDigestAlgo, provider);
```
jasonkatonica, line 38: This change is missing from other releases. Perhaps we have to do another PR for all the others from main on back after these merge?

PR #1668, opened 2026-07-23
Introduce SystemAccessUtils to all Security Manager permission checks

taoliult:
Centralize all security manager permission checks in SystemAccessUtils wrapper methods instead of AccessController.doPrivileged() calls throughout the codebase.

SystemAccessUtils now wraps all permission-requiring operations:
- System property access (getSystemProperty)
- File system access (fileExists, newFileReader, getFileCanonicalPath)
- Library loading (loadLibrary)
- Class loading (forName)

This centralizes permission checks in one utility class, making the codebase cleaner and easier to maintain across JDK versions (SecurityManager in JDK 21 and earlier, removed AccessController.doPrivileged() in JDK 25+).

This is a back port PR from PR: https://github.com/IBM/OpenJCEPlus/pull/1700

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
@@ -21,9 +21,11 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
     // Field serialVersionUID per tag [SERIALIZATION] in DesignNotes.txt
     private static final long serialVersionUID = 929669768004683845L;
 
-    private static final boolean printFipsDeveloperModeWarning = Boolean.parseBoolean(System.getProperty("openjceplus.fips.devmodewarn", "true"));
+    private static final boolean printFipsDeveloperModeWarning = Boolean.parseBoolean(
+            SystemAccessUtils.getSystemProperty("openjceplus.fips.devmodewarn", "true"));
 
-    private static final boolean allowNonOAEPFIPS = Boolean.parseBoolean(System.getProperty("com.ibm.openjceplusfips.allowNonOAEP", "false"));
+    private static final boolean allowNonOAEPFIPS = Boolean.parseBoolean(
+            SystemAccessUtils.getSystemProperty("com.ibm.openjceplusfips.allowNonOAEP", "false"));
 
     private static final String info = "OpenJCEPlusFIPS Provider implements the following:\n" +
 
@@ -67,8 +69,8 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
         supportedPlatforms.put("Arch", List.of("amd64", "ppc64", "s390x"));
         supportedPlatforms.put("OS", List.of("Linux", "AIX", "Windows", "z/OS"));
 
-        osName = System.getProperty("os.name");
-        osArch = System.getProperty("os.arch");;
+        osName = SystemAccessUtils.getSystemProperty("os.name");
+        osArch = SystemAccessUtils.getSystemProperty("os.arch");
 
         boolean isOsSupported, isArchSupported;
         // Check whether the OpenJCEPlus FIPS is supported.
@@ -180,7 +182,7 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
     private static String getDebugDate(String className) {
         String versionDate = "Unknown";
         try {
-            Class<?> thisClass = Class.forName(className);
+            Class<?> thisClass = SystemAccessUtils.forName(className);
             Package thisPackage = thisClass.getPackage();
             String versionInfo = thisPackage.getImplementationVersion();
             int index = versionInfo.indexOf("_");
--- a/src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
@@ -9,11 +9,8 @@
 package com.ibm.crypto.plus.provider;
 
 import java.io.BufferedReader;
-import java.io.FileReader;
 import java.io.IOException;
 import java.io.StringReader;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -139,11 +136,11 @@ public class ProviderServiceReader {
                 throw new IOException("No file or BufferedReader specified");
             } else if (null == filePath && this.reader != null) {
                 rd = this.reader;
-            } else if (filePath != null && !Files.exists(Paths.get(filePath))) {
+            } else if (filePath != null && !SystemAccessUtils.fileExists(filePath)) {
                 throw new IOException("File not found: " + filePath);
             } else {
                 // this filePath != null && Files.exists(Paths.get(filePath))
-                rd = new BufferedReader(new FileReader(filePath));
+                rd = new BufferedReader(SystemAccessUtils.newFileReader(filePath));
             }
 
             pr.load(rd);
--- a/src/main/java/com/ibm/crypto/plus/provider/SystemAccessUtils.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/SystemAccessUtils.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright IBM Corp. 2026
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms provided by IBM in the LICENSE file that accompanied
+ * this code, including the "Classpath" Exception described therein.
+ */
+
+package com.ibm.crypto.plus.provider;
+
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
+
+/**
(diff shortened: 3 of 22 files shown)


src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -94,6 +96,10 @@ public OpenJCEPlusFIPS() {
         if (debug != null) {
             debug.println("New OpenJCEPlusFIPS instance");
         }
+        // Eagerly verify that the FIPS native adapter can be obtained.
+        // Without this check, the provider silently loads in a non-FIPS
+        // state when the SecurityManager blocks the OCK library from loading.
+        NativeOCKAdapterFIPS.getInstance();
```
johnpeck-us-ibm: What happens if this call fails? Should this be put into a try catch and a unique message be display?

johnpeck-us-ibm (changes requested):

src/main/java/com/ibm/crypto/plus/provider/base/Digest.java
```diff
@@ -39,7 +38,7 @@ public final class Digest implements Cloneable {
     final static int[] digestLengths = {32, 48, 64, 28, 20};
 
     //disable caching mechanism for windows OS
-    final static private boolean isWindows = System.getProperty("os.name").startsWith("Windows");
+    final static private boolean isWindows = SystemAccessUtils.doPrivileged(() -> System.getProperty("os.name")).startsWith("Windows");
```
KostasTsiounis: Should we change this to something that most resembles the existing call? Could we do something like `SystemAccessUtils.getProperty("os.name");`

KostasTsiounis (changes requested):

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
```diff
@@ -143,7 +143,7 @@ public List<ServiceDefinition> readServices() throws IOException {
                 throw new IOException("File not found: " + filePath);
```
JinhangZhang, line 140: Should we add a privileged check on this Files.exists() function?

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
taoliult: Removed these lines in OpenJCEPlusFIPS constructor since it is no longer needed.

`SystemAccessUtils.runPrivileged()`, which wraps sensitive operation in `AccessController.doPrivileged()`. The `NativeOCKAdapter` constructor now calls `initializeContext()` inside that privileged block. `initializeContext()` has a try catch that converts every possible failure — `UnsatisfiedLinkError`, `SecurityException`, `OCKException` — into a `ProviderException`. Because `ProviderException` is unchecked, it propagates straight through `doPrivileged` and crashes the `NativeOCKAdapterFIPS` constructor. There is no code path that lets initialization fail silently.

src/main/java/com/ibm/crypto/plus/provider/base/Digest.java
taoliult: I considered adding a `SystemAccessUtils.getProperty()` method before, but the problem is it doesn't stop there — we would also need separate methods for `System.load()`, `File.exists()`, `File.getAbsolutePath()`, `new FileReader()`, `Class.forName()`, OCK context initialization, and so on. That would turn `SystemAccessUtils` into an ever-growing class with one wrapper per JDK operation.

The three existing methods — `doPrivileged()`, `doPrivilegedChecked()`, and `runPrivileged()` — I think they can cover all cases in the current codebase: value-returning without checked exceptions, value-returning with checked exceptions, and void operations respectively. Any future privileged operation can be wrapped with one of these three without touching SystemAccessUtils.

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
taoliult: Yes, updated.

johnpeck-us-ibm (approved):

src/main/java/com/ibm/crypto/plus/provider/SystemAccessUtils.java
```diff
@@ -0,0 +1,78 @@
...
+     * @return the value returned by {@code action}
+     * @throws Exception any checked exception thrown by {@code action}
+     */
+    @SuppressWarnings("removal")
+    public static <T> T doPrivilegedChecked(Callable<T> action) throws Exception {
+        try {
+            return AccessController.doPrivileged(
+                    (java.security.PrivilegedExceptionAction<T>) action::call);
+        } catch (java.security.PrivilegedActionException pae) {
+            throw pae.getException();
```
jasonkatonica: Can we more simply just throw the original exception here such that we know what may be wrong? Additionally `pae.getException()` could return `null`?

src/main/java/com/ibm/crypto/plus/provider/ock/NativeOCKImplementation.java
```diff
@@ -170,31 +171,23 @@ static void preloadOCK() {
...
                 return true;
             } catch (Throwable t) {
-                if (debug != null) {
-                    debug.println("Failed to load : " + libraryName);
-                }
-            }
-        } else {
-            if (debug != null) {
-                debug.println("Skipping load of " + libraryName);
+                return false;
```
jasonkatonica: Seems like some debugs were lost here such as the case where we get into this catch block?

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/SystemAccessUtils.java
taoliult: The checked exception thrown by `action.call()` is wrapped in a `PrivilegedActionException`. If we throw `pae`, the caller receives the `PrivilegedActionException` wrapper rather than the original exception thrown by `action.call()`.

For example, if the action throws an `IOException`:

`throw pae;` - The caller receives a `PrivilegedActionException` whose cause is the `IOException`.

`throw pae.getException();` - The caller receives the original `IOException` directly.

I think both of them are ok, but maybe throwing the original exception makes it clearer to the caller?

I updated the code to unwrap the exception when available and throw `PrivilegedActionException` directly if `getException()` returns `null`:

```
Exception ex = pae.getException();
throw ex != null ? ex : pae;
```

src/main/java/com/ibm/crypto/plus/provider/ock/NativeOCKImplementation.java
taoliult: Yes, added the debug.

jasonkatonica: Do we also need a trace like before after line 180 above to cover the successful path of a load?

src/main/java/com/ibm/crypto/plus/provider/SystemAccessUtils.java
jasonkatonica: OK seems like a reasonable approach to me.

src/main/java/com/ibm/crypto/plus/provider/ock/NativeOCKImplementation.java
taoliult: Previously, there was a single debug output after the load operation that checked the `loaded` boolean to print either `"Loaded : "` or `"Skipped (not found): "`. Code as below:

```
        if (debug != null) {
            debug.println((loaded ? "Loaded : " : "Skipped (not found): ")
                    + libraryFile.getAbsolutePath());
        }
```

Since we added debug output in the try catch block to print the exception stack trace, this exception would trigger both the debug output and the existing `Skipped (not found):` message above. This could be confusing.

So, I removed the single debug output and added more specific debug messages before each return so that the message is more accurate.

* "File not found: <path>" — when the file does not exist
* "Loaded: <path>" — when System.load() succeeds
* "Failed to load <path>: <ex>" — when System.load() throws, including the exception detail.

jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/base/Digest.java
jasonkatonica: Hi @taoliult I also lean toward it being a better approach what Kostas mentioned above for a few reasons. I think it will be easier to ensure that a call that needs a privledged action actually calls the utility class. We can review it as so on PRs made to main. Also doing this will avoid having generic lambda functions throughout the code making it a bit cleaner. For example all calls would do a `SystemAccessUtils.getProperty` instead of having to write a lambda expression for each.

taoliult: Yes, updated. SystemAccessUtils now wraps all permission-requiring operations:

- System property access (getSystemProperty)
- File system access (fileExists, newFileReader, getFileCanonicalPath)
- Library loading (loadLibrary)
- Class loading (forName)

KostasTsiounis (approved):
LGTM

JinhangZhang (approved):
LGTM

jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/ProviderServiceReader.java
jasonkatonica: Does this need another full round of PRs to make this match in all the releases ?

PR #1689, opened 2026-07-29
Fix iteration counter reset for AESCCM parameterized  tests

Mohit-Rajbhar100698:
The iteration counter was shared across parameterized test runs, causing subsequent provider executions to skip the test loop. Reset the counter before each parameterized execution and update debug logs.

Changes:
- Reset the iteration counter in TestAESCCM.java and TestAESCCM2.java during test setup for each parameterized execution.
- Update debug logs.

Signed-off-by: Mohit Rajbhar [mohit.rajbhar@ibm.com](mailto:mohit.rajbhar@ibm.com)

PR #1715, opened 2026-08-19
Fix iteration counter reset for AESCCM parameterized  tests

Mohit-Rajbhar100698:
The iteration counter was shared across parameterized test runs, causing subsequent provider executions to skip the test loop. Reset the counter before each parameterized execution and update debug logs.

Changes:
- Reset the iteration counter in TestAESCCM.java and TestAESCCM2.java during test setup for each parameterized execution.
- Update debug logs.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1689

### DIFF
--- a/src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
+++ b/src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
@@ -69,6 +69,7 @@ public class TestAESCCM extends BaseTest {
     @BeforeEach
     public void setUp() throws Exception {
         setAndInsertProvider(provider);
+        iterationCounter = 0;
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


johnpeck-us-ibm (approved):

src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
johnpeck-us-ibm, line 43: This should not be static if we are going to run this test multithread at a time in the future.

johnpeck-us-ibm, line 46: Same comment as above.

johnpeck-us-ibm, line 63: Same comment as above. Each instance of the test should have there own value here.

johnpeck-us-ibm (changes requested):

src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
jasonkatonica: Agreed that this can be a problem with the use of statics. Mohit we will need another round of PRs to fix this starting with `main` then backported again to each release.

Mohit-Rajbhar100698: Ok , I will open another PR to fix this

jasonkatonica (approved):
LGTM. Be sure to open an issue for the use of `statics` in this repostiroy and link the issue to this PR.

johnpeck-us-ibm (approved):

PR #1720, opened 2026-08-19
Add configAlgName for NativeCryptoSelector backend lookup in PBES2Core

taoliult:
Follow the same pattern as `PBES1Core`, pass the registered `Cipher` service name (e.g. "`PBEWithHmacSHA1AndAES_128`") to `PBKDF2Core` and `AESCipher` so that `NativeCryptoSelector` resolves the backend against the same service as the caller.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/PBES2Core.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/PBES2Core.java
@@ -42,7 +42,7 @@ abstract class PBES2Core extends CipherSpi {
      */
     PBES2Core(String kdfAlgo, String cipherAlgo, int keySize, OpenJCEPlusProvider provider)
         throws NoSuchAlgorithmException, NoSuchPaddingException {
-        
+
         blkSize = AESCipher.AES_BLOCK_SIZE;
         this.cipherAlgo = cipherAlgo;
         keyLength = keySize * 8;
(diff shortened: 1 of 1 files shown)


jasonkatonica (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/PBES2Core.java
```diff
@@ -40,39 +40,39 @@ abstract class PBES2Core extends CipherSpi {
...
+
         blkSize = AESCipher.AES_BLOCK_SIZE;
         this.cipherAlgo = cipherAlgo;
         keyLength = keySize * 8;
         pbeAlgo = "PBEWith" + kdfAlgo + "And" + cipherAlgo + "_" + keyLength;
         this.provider = provider;
 
         if (cipherAlgo.equalsIgnoreCase("AES")) {
-            cipher = new AESCipher(provider);
+            cipher = new AESCipher(provider, configAlgName);
```
KostasTsiounis: The `configAlgName` seems to be the same as `pbeAlgo`. I don't think we need another parameter.

KostasTsiounis (changes requested):

KostasTsiounis (approved):
LGTM

src/main/java/com/ibm/crypto/plus/provider/PBES2Core.java
taoliult: Yes, updated to use the `pbeAlgo`.

johnpeck-us-ibm (approved):

PR #1723, opened 2026-08-20
Remove static from test variables

Mohit-Rajbhar100698:
Remove the static modifier from the test variables to ensure each instance maintains its own state and to support future multithreaded test execution.
Changes:
- In src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java file.
- In src/test/java/ibm/jceplus/junit/tests/TestAESCCM2.java file.

### DIFF
--- a/src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
+++ b/src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
@@ -39,28 +39,24 @@ public class TestAESCCM extends BaseTest {
     @Parameter(0)
     TestProvider provider;
 
-    public static int iterationLimit = 100;
-    public static int iterationCounter = 0;
+    public static final int iterationLimit = 100;
+    public int iterationCounter = 0;
 
     // The plainText string to be encrypted and decrypted will be selected randomly for each iteration.
-    public static String plainText = null;
+    public String plainText = null;
 
-    public static String plainTextShort = "A short text string to be encrypted.";
-    public static String plainTextMedium = "A medium text string to be encrypted. A medium text string to be encrypted. A medium text string to be encrypted. A medium text string to be encrypted. A medium text string to be encrypted.";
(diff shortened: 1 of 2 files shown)


src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
```diff
@@ -39,32 +39,32 @@ public class TestAESCCM extends BaseTest {
     @Parameter(0)
     TestProvider provider;
 
-    public static int iterationLimit = 100;
-    public static int iterationCounter = 0;
+    public int iterationLimit = 100;
+    public int iterationCounter = 0;
```
johnpeck-us-ibm, line 44: Not all these need to be non-static and most of these probably should be static final. The variables that need to be non-static are the ones that will change during an execution of the code. Also, when running multiple threads the variable will need to be unique for each thread otherwise every thread is using the same variable.

johnpeck-us-ibm (changes requested):

src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
```diff
@@ -39,28 +39,28 @@ public class TestAESCCM extends BaseTest {
...
-    public static int AESKeySize128 = 128;
-    public static int AESKeySize192 = 192;
-    public static int AESKeySize256 = 256;
+    public static final int AESKeySize128 = 128;
+    public static final int AESKeySize192 = 192;
+    public static final int AESKeySize256 = 256;
 
     // This CCM TAG LENGTH is specified in bits.  Valid values are:  32, 48, 64, 60, 96, 112, and 128
     // Although initialized here, the ccmTagLength will actually be selected randomly for each iteration.
     public static int ccmTagLength = 128;
```
Mohit-Rajbhar100698: This seems to be a dead variable . There is no any references to this variable.
Should I remove it ?

src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
Mohit-Rajbhar100698: Updated.

I made the variables that do not change and have constant values static final. The variables that change during execution are kept non-static so that each thread can have its own instance.

src/test/java/ibm/jceplus/junit/tests/TestAESCCM.java
johnpeck-us-ibm: Yes. If not referenced please remove it.

Mohit-Rajbhar100698: Removed

johnpeck-us-ibm (approved):

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

PR #1727, opened 2026-08-20
Print warning when services requested instead of initialization

KostasTsiounis:
The warnings for the developer mode and the z platforms were printed as part of the constructor of the `OpenJCEPlusFIPS` provider.

This led to unnecessary prints when the provider was only loaded to check against other requested providers.

The warnings are now only printed when one requests services from `OpenJCEPlusFIPS`. A check is, also, added to ensure the warnings are only printed once.

### DIFF
--- a/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
+++ b/src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
@@ -14,6 +14,7 @@ import java.security.ProviderException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
 
@@ -57,6 +58,10 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
     // to find ourselves or run the risk of not being in the list.
     private static volatile OpenJCEPlusFIPS instance;
 
+    // Guards the one-time printing of the developer-mode / s390x warnings,
+    // shared across all instances of this provider.
+    private static volatile boolean warningsPrinted = false;
+
     private static boolean ockInitialized = false;
 
     private static final boolean isFIPSCertifiedPlatform;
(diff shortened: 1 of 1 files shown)


src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -149,6 +137,43 @@ OpenJCEPlusProvider getProvider() {
         }
     }
 
+    // Print developer-mode and s390x warnings the first time a service is
+    // requested from this provider instance, rather than at construction time.
+    private void printWarningsOnce() {
+        if (!warningsPrinted.compareAndSet(false, true)) {
```
jasonkatonica: Can we read the boolean here without grabbing the lock each time? the compare and swap may add a bit of overhead on each of these checks.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
```diff
@@ -55,6 +57,9 @@ public final class OpenJCEPlusFIPS extends OpenJCEPlusProvider {
     // to find ourselves or run the risk of not being in the list.
     private static volatile OpenJCEPlusFIPS instance;
 
+    // Guards the one-time printing of the developer-mode / s390x warnings.
+    private final AtomicBoolean warningsPrinted = new AtomicBoolean(false);
```
jasonkatonica: Should this be `static` and `volatile`. It seems like we want to have a single `AtomicBoolean` here not one per provider instance.

jasonkatonica (changes requested):

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
KostasTsiounis: Switch to boolean with double check and synchronization scheme to avoid unnecessary locking.

src/main/java/com/ibm/crypto/plus/provider/OpenJCEPlusFIPS.java
KostasTsiounis: Added both to match new way of synchronizing.

jasonkatonica (approved):
LGTM

johnpeck-us-ibm (approved):

taoliult (approved):
LGTM

JinhangZhang (approved):
LGTM

PR #1728, opened 2026-08-20
Document security policy grant required for non-Semeru JDKs

taoliult:
Users on JDKs other than IBM Semeru must manually add the openjceplus grant block to their JDK policy file. Add a new README section with the full grant, setup steps, and a note on the file: codeBase alternative for JAR-based installations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
   - [Run All Tests](#run-all-tests)
   - [Run Single Test](#run-single-test)
 - [OpenJCEPlus and OpenJCEPlusFIPS Provider SDK Installation](#openjceplus-and-openjceplusfips-provider-sdk-installation)
+- [Security Policy Configuration](#security-policy-configuration)
 - [Configuration Options](#configuration-options)
 - [Features and Algorithms](#features-and-algorithms)
 - [Contributions](#contributions)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -288,6 +289,62 @@ take effect.
...
+```
+
+Alternatively, if your application uses a custom policy file (specified via `-Djava.security.policy`), add the grant block there instead.
+
+### Step 2 – Add the `openjceplus` grant block
+
+Open the policy file and append the following grant block:
+
+```
+grant codeBase "jrt:/openjceplus" {
```
jasonkatonica: In this syntax i believe that `openjceplus` is the module name. At this time the only runtime that i am aware of that has openjceplus as a module is IBM Semeru. Is this syntax correct for non Semeru JDKS that do not have a `openjceplus` module built in.

README.md
```diff
@@ -288,6 +289,62 @@ take effect.
...
+
+### Step 3 – Verify the change
+
+Restart your application. If a `SecurityManager` is in use, the `AccessControlException` errors related to `openjceplus` should no longer appear. You can confirm the policy is being picked up with:
+
+```console
+java -Djava.security.debug=access,domain -version 2>&1 | grep openjceplus
+```
+
+**NOTE**: The `jrt:/openjceplus` code base is only resolvable when `openjceplus` is installed as a named module in the runtime image. If you are loading it from a JAR via `-Xbootclasspath/a`, the `codeBase` URL will differ and you should instead use a `file:` URL pointing to your JAR, for example:
```
jasonkatonica: This is the missing information i didnt notice when writting my comment above. Perhaps we should lead with the syntax for using a `file:` since Semeru users will have this built in? We can then add a NOTE to state semeru users have this built in using the `jrt:/openjceplus` syntax.

README.md
```diff
@@ -288,6 +289,62 @@ take effect.
         ```console
         -Djgskit.library.path=$ANYDIRECTORY
         ```
+## Security Policy Configuration for Non-Semeru JDKs
```
jasonkatonica: Seems this title is off a bit given the section is for all java jdks , not just non-semeru.

jasonkatonica (changes requested):

README.md
taoliult: Agree, updated.

README.md
taoliult: Agree, updated.

README.md
taoliult: Yes, removed "`for Non-Semeru JDKs`", just use "`Security Policy Configuration`" as the title.

jasonkatonica (approved):
LGTM

README.md
```diff
@@ -288,6 +289,60 @@ take effect.
         ```console
         -Djgskit.library.path=$ANYDIRECTORY
         ```
+## Security Policy Configuration
+
+**NOTE**: This section applies only when the JVM Security Manager is enabled. The Security Manager was deprecated in Java 17 and removed in Java 24.
+
+When running on a JDK that does **not** bundle `OpenJCEPlus` (i.e., any JDK other than IBM Semeru), the JVM security manager requires an explicit policy grant so that the `openjceplus` module can perform the operations it needs. Without this grant, you may encounter `AccessControlException` errors at runtime.
```
KostasTsiounis: I wouldn't refer to it as a `module` since it's not one in every JDK other than Semeru.

KostasTsiounis (changes requested):

JinhangZhang (approved):
LGTM

README.md
taoliult: If we should avoid using the term "module" here, would it be better to replace it with "provider" or "code"? Like "`openjceplus provider`" or "`openjceplus code`", or any other suggestions?

KostasTsiounis: Given that your are talking about "operations it needs", I think `code` is more appropriate. @jasonkatonica what do you think?

jasonkatonica: `openjceplus code` sounds good to me also given this is not always a module ( unless you count this code running on the `unnamed` module as a module )

taoliult: Or just  `openjceplus`, without `code` or `module`, like "so that the `openjceplus` can perform the operations it needs." @jasonkatonica @KostasTsiounis

KostasTsiounis: I don't think it reads as well that way.

taoliult: Updated. Just use `openjceplus`, without referring to it as a `module` or `code`.

johnpeck-us-ibm (approved):

README.md
```diff
@@ -288,6 +289,60 @@ take effect.
         ```console
         -Djgskit.library.path=$ANYDIRECTORY
         ```
+## Security Policy Configuration
+
+**NOTE**: This section applies only when the JVM Security Manager is enabled. The Security Manager was deprecated in Java 17 and removed in Java 24.
+
+When running on a JDK that does **not** bundle `OpenJCEPlus` (i.e., any JDK other than IBM Semeru), the JVM security manager requires an explicit policy grant so that the `openjceplus` can perform the operations it needs. Without this grant, you may encounter `AccessControlException` errors at runtime.
+
+IBM Semeru ships with this grant pre-configured in its `default.policy`. If you use any other JDK you must add it manually.
```
jasonkatonica: We should also add a sentence that mentions that users who make direct use of the classes in use by `OpenJCEPlus` should also add rules accordingly allowing their application to also make use of the below classes directly.

jasonkatonica (changes requested):

README.md
taoliult: Updated the sentence to mention this and also points it to the new added "`Custom Policy Files and AccessControlException`" subsection for detailed explanations and example.

taoliult:
@jasonkatonica @johnpeck-us-ibm @KostasTsiounis @JinhangZhang 

Added a new subsection "Custom Policy Files and AccessControlException" to explain that the Security Manager checks permissions on every frame of the call stack — meaning the calling code also needs the required permission in its own policy grant block. This subsection explains the symptom, root cause, and how to fix it. Please review.

jasonkatonica (approved):
LGTM

KostasTsiounis (approved):
LGTM

PR #1740, opened 2026-08-25
Print warning when services requested instead of initialization

KostasTsiounis:
The warnings for the developer mode and the z platforms were printed as part of the constructor of the `OpenJCEPlusFIPS` provider.

This led to unnecessary prints when the provider was only loaded to check against other requested providers.

The warnings are now only printed when one requests services from `OpenJCEPlusFIPS`. A check is, also, added to ensure the warnings are only printed once.

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1727


JinhangZhang (approved):
LGTM

taoliult (approved):

johnpeck-us-ibm (approved):

Issue #1567, opened 2026-07-06
Support generic ML-DSA


jasonkatonica:
According to https://openjdk.org/jeps/497 when encoding and decoding `ML-DSA` keys you can use the `ML-DSA` `KeyFactory` implementation to convert an `ML-DSA` private key from its PKCS #8 encoding, or an `ML-DSA` public key to or from its X.509 encoding. For example, to convert a `ML-DSA` private key to its PKCS #8 encoding, and vice-versa:
```
KeyFactory f = KeyFactory.getInstance("ML-DSA");
PKCS8EncodedKeySpec p8spec = f.getKeySpec(kp.getPrivate(),
                                          PKCS8EncodedKeySpec.class);
PrivateKey sk2 = f.generatePrivate(p8spec);
```
Similarly, to convert a ML-DSA public key to its X.509 encoding, and vice-versa:
```
X509EncodedKeySpec x509spec = f.getKeySpec(kp.getPublic(),
                                           X509EncodedKeySpec.class);
PublicKey pk2 = f.generatePublic(x509spec);
```
The `KeyFactory` implementation can also translate a key from another security provider using the `translateKey` method, as long as its encoding format is supported. The `getAlgorithm` method of a Key object generated by an `ML-DSA` `KeyPairGenerator` or `KeyFactory` implementation always returns the family name `ML-DSA`, regardless of whether the `KeyPairGenerator` or `KeyFactory` was instantiated with the `ML-DSA` family name or one of the parameter-set names. 

The `getParams` method of an `ML-DSA` key returns a `NamedParameterSpec` object that matches the key's parameter-set name. If a `KeyFactory` object is instantiated with a family name, it encodes or decodes `ML-DSA` keys in the family with any parameter set. If it is instantiated with a parameter-set name, it only encodes or decodes `ML-DSA` keys that use that parameter set; otherwise, the `translateKey` method throws an `InvalidKeyException`, and the `generatePrivate`, `generatePublic`, and `getKeySpec` methods throw an `InvalidKeySpecException`.

The encoding used by the `ML-DSA` `KeyFactory` is defined in a draft IETF RFC. We will track changes in this draft until it is published.

openjceplus should align a similar behaviour as `SUN` provider for the generic `ML-DSA`

A failed test:
```
import java.security.KeyFactory;
import java.security.KeyPair;
import java.security.KeyPairGenerator;
import java.security.PrivateKey;
import java.security.Provider;
import java.security.PublicKey;
[...]

PR #1667, opened 2026-07-22
Add generic ML-DSA KeyFactory and Signature

jasonkatonica:
Register a family-level `ML-DSA` `KeyFactory` and `Signature` service so that callers can use the algorithm name `ML-DSA` without specifying a parameter set, matching the JEP 497 / JDK 24+ API contract.

- Register `PQCKeyFactory$MLDSA` and `PQCSignatureImpl$MLDSA` as the `ML-DSA` service. Remove the `ML-DSA` alias from `ML-DSA-65` (it was a
  mis-mapping that silently directed all generic lookups to ML-DSA-65).
- Split the single `name` field into `familyName` (returned by `getAlgorithm()` such as `ML-DSA`) and `paramSetName` (the concrete parameter set such as `ML-DSA-65`). Add getters and setters for these values such that other code can act accordingly.
- Add tests to `BaseTestPQCKeyInterop`. These tests cover all three `ML-DSA` parameter sets via `@ParameterizedTest`.
- Add tests to `BaseTestPQCKeys`. Expand alias coverage to include case variants, OID aliases, and bare OID strings. Added tests for `getAlgorithm()` family-name correctness for all ML-DSA and ML-KEM aliases
- Add tests to `BaseTestPQCSignature` for generic `ML-DSA` sign/verify tests to all three parameter sets.

Fixes: https://github.com/IBM/OpenJCEPlus/issues/1567

Signed-off-by: Jason Katonica <katonica@us.ibm.com>

PR #1760, opened 2026-09-09
Add generic ML-DSA KeyFactory and Signature

jasonkatonica:
Register a family-level `ML-DSA` `KeyFactory` and `Signature` service so
that callers can use the algorithm name `ML-DSA` without specifying a
parameter set, matching the JEP 497 / JDK 24+ API contract.

- Register `PQCKeyFactory$MLDSA` and `PQCSignatureImpl$MLDSA` as the
`ML-DSA` service. Remove the `ML-DSA` alias from `ML-DSA-65` (it was a
  mis-mapping that silently directed all generic lookups to ML-DSA-65).
- Split the single `name` field into `familyName` (returned by
`getAlgorithm()` such as `ML-DSA`) and `paramSetName` (the concrete
parameter set such as `ML-DSA-65`). Add getters and setters for these
values such that other code can act accordingly.
- Add tests to `BaseTestPQCKeyInterop`. These tests cover all three
`ML-DSA` parameter sets via `@ParameterizedTest`.
- Add tests to `BaseTestPQCKeys`. Expand alias coverage to include case
variants, OID aliases, and bare OID strings. Added tests for
`getAlgorithm()` family-name correctness for all ML-DSA and ML-KEM
aliases
- Add tests to `BaseTestPQCSignature` for generic `ML-DSA` sign/verify
tests to all three parameter sets.

Fixes: https://github.com/IBM/OpenJCEPlus/issues/1567

Back-ported from: https://github.com/IBM/OpenJCEPlus/pull/1667


KostasTsiounis (approved):
LGTM

taoliult (approved):

PR #1788, opened 2026-09-14
Enable SHA3 digests for RSASSA-PSS

JinhangZhang:
Add SHA3-224, SHA3-256, SHA3-384, and SHA3-512 support to the RSASSA-PSS implementation.

Normalize digest names before validating PSS parameters so that message digest and MGF1 digest comparisons are based on canonical algorithm names.


johnpeck-us-ibm:
Should part of this work be moving these tests to the parameterized ones?

src/main/java/com/ibm/crypto/plus/provider/base/SignatureRSAPSS.java
```diff
@@ -75,76 +76,67 @@ public synchronized void setParameter(String digestAlgo, int saltlen, int traile
...
-                break;
-            case "SHA5":
-            case "SHA-5":
-            case "SHA512":
-            case "SHA-512":
-                digestAlgoOCK = "SHA512";
-                break;
-            default:
-                digestAlgoOCK = digestAlgo;
+    private static String toOCKDigestName(String digestAlgo) {
```
KostasTsiounis, line 79: I wouldn't use `OCK` in the name, as the same method will be used for `OpenSSL` when the algorithm is implemented there as well.

KostasTsiounis (changes requested):

PR #1790, opened 2026-09-14
PBES2Parameters: omit default HmacSHA1 prf field in DER encoding

taoliult:
According to JDK-8371156: https://bugs.openjdk.org/browse/JDK-8371156

The `PBKDF2` `prf` field defaults to `HmacSHA1` and must be omitted from the `DER` encoding when using the default value. `engineGetEncoded()` was always encoding this field, causing the `ImportPassKeyAlg` regression test to fail when using `PBEWithHmacSHA1AndAES_128`.

Since this upstream change only applies to JDK 26 and later, PRs will only be created for the Main and JDK 27 branches.


src/main/java/com/ibm/crypto/plus/provider/PBES2Parameters.java
```diff
@@ -336,12 +336,15 @@ protected byte[] engineGetEncoded() throws IOException {
             pBKDF2_params.putInteger(keysize / 8); // derived key length (in octets)
         }
 
-        DerOutputStream prf = new DerOutputStream();
-        // algorithm is id-hmacWith<MD>
-        prf.putOID(kdfAlgo_OID);
-        // parameters is 'NULL'
-        prf.putNull();
-        pBKDF2_params.write(DerValue.tag_Sequence, prf);
+        // HmacSHA1 is the default and must not be encoded.
```
jasonkatonica, line 339: Can we add a test that exercises this path and forces the previous failure.

jasonkatonica (changes requested):

PR #1770, opened 2026-09-10
Remove reflection from provider instantiation in tests

michael-ma-dev:
Multiple places use reflection to create providers unnecessarily. 

Replace `Class.forName(...).getDeclaredConstructor().newInstance()` with direct constructors for `OpenJCEPlus`, `OpenJCEPlusFIPS`, and `BouncyCastleProvider`.

Fixes: https://github.com/IBM/OpenJCEPlus/issues/1709
Signed-off-by: Michael Ma <michael.ma@ibm.com>

PR #1803, opened 2026-09-15
Enable JMH performance testing for OpenSSL backend

KostasTsiounis:


src/test/java/ibm/jceplus/jmh/JMHBase.java
```diff
@@ -176,6 +183,28 @@ protected void insertProvider(String provider) throws Exception {
                 java.security.Security.removeProvider("OpenJCEPlus");
             }
             java.security.Security.insertProviderAt(myProvider, 1);
+        } else if (provider.equalsIgnoreCase("OpenJCEPlus-OpenSSL")) {
+            Provider myProvider = java.security.Security.getProvider("OpenJCEPlus-OpenSSL");
+            if (myProvider == null) {
+                Object ojp = Class.forName("com.ibm.crypto.plus.provider.OpenJCEPlus")
```
jasonkatonica: Do we have to use reflection here and line 200?

JenkinsfilePerformance
```diff
@@ -226,10 +229,20 @@ def run(platform) {
                     echo "Binaries fetched"
                     externalLibrary.getMaven(software)
                     echo "Maven fetched"
+                    if (software != "zos") {
+                        externalLibrary.getOpenSSL(hardware, software)
```
jasonkatonica: Should we do this all the time or only when OpenSSL is selected. Perhaps this can be moved to after line 244.

jasonkatonica: Does OPENSSL_RELEASE value get picked up automatically by `externalLibrary.getOpenSSL(hardware, software)` without passing it to that method in this location?

jasonkatonica (changes requested):

JenkinsfilePerformance
KostasTsiounis: > Should we do this all the time or only when OpenSSL is selected. Perhaps this can be moved to after line 244.

The way the build scripts are set up now, we always need OpenSSL and OCK to do the full build, regardless of whether the provider is used or not. We could change this in the future if we decide this is the path forward.

src/test/java/ibm/jceplus/jmh/JMHBase.java
KostasTsiounis: I will rebase after https://github.com/IBM/OpenJCEPlus/pull/1770 is merged and fix that.

JenkinsfilePerformance
KostasTsiounis: > Does OPENSSL_RELEASE value get picked up automatically by `externalLibrary.getOpenSSL(hardware, software)` without passing it to that method in this location?

It seems to work as expected both in the regular and performance pipeline.

PR #1804, opened 2026-09-16
Add getParams support for PQC keys

JinhangZhang:
Implement getParams() in PQCPrivateKey and PQCPublicKey to return the specific ML-KEM or ML-DSA parameter set as a NamedParameterSpec.

Store the parameter specification as transient state and reconstruct it when PQC keys are created or decoded.

Add coverage for all supported ML-KEM and ML-DSA parameter sets.


src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeys.java
```diff
@@ -288,6 +289,33 @@ public void genWithAlgParameterSpecMLDSAFaiure(String algParamSpecName) throws E
...
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = {
+            "ML-KEM-512",
+            "ML-KEM-768",
+            "ML-KEM-1024",
+            "ML-DSA-44",
+            "ML-DSA-65",
+            "ML-DSA-87"
```
jasonkatonica, line 299: Can we update these to also exercise the generic family names, not just specific parameter sets.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeys.java
```diff
@@ -288,6 +289,33 @@ public void genWithAlgParameterSpecMLDSAFaiure(String algParamSpecName) throws E
...
+                (NamedParameterSpec) keyPair.getPrivate().getParams();
+        assertEquals(algorithm, privateParams.getName());
+
+        assertTrue(keyPair.getPublic().getParams() instanceof NamedParameterSpec);
+        NamedParameterSpec publicParams =
+                (NamedParameterSpec) keyPair.getPublic().getParams();
+        assertEquals(algorithm, publicParams.getName());
+    }
+
     /**
```
jasonkatonica, line 319: Can we also add a interop test for `getParams()`. Our providers behavior should match other providers in this area.

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
```diff
@@ -31,8 +32,8 @@ final class PQCPrivateKey extends PKCS8Key {
     private String familyName;      // algorithm family name returned by getAlgorithm()
     private String paramSetName; // specific parameter-set name (e.g. "ML-DSA-65")
 
+    private transient NamedParameterSpec params;
```
jasonkatonica, line 35: Is this redundant with `paramSetName` above? Seems like we could simply just construct a new `NamedParameterSpec` in the new `getParams` method using `this.paramSetName` and return. Same comment/question for `PQCPublicKey.java` changes.

jasonkatonica (changes requested):

johnpeck-us-ibm:
Are we doing this for some JDK tests that need this? The parameters are all hard coded and if you know the alg you know the parameters since they do not change.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeys.java
JinhangZhang: Updated

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeys.java
JinhangZhang: Updated

src/main/java/com/ibm/crypto/plus/provider/PQCPrivateKey.java
JinhangZhang: You are right, i updated the code and use the paramSetName instead

JinhangZhang:
> Are we doing this for some JDK tests that need this? The parameters are all hard coded and if you know the alg you know the parameters since they do not change.

Not specifically for a JDK test. The main reason is to align our PQC key behavior with the upstream JCA API and the JDK providers.

`PublicKey` and `PrivateKey` now both extend `AsymmetricKey`, and `AsymmetricKey` introduced a `getParams()` method. Its default implementation returns `null`, so our PQC keys currently inherit that behavior unless we override it.

For PQC keys, `getAlgorithm()` now returns only the family name, such as `ML-KEM` or `ML-DSA`, not the specific parameter set. The JDK providers override `getParams()` for these keys and return the specific parameter set as a `NamedParameterSpec`, so this change makes OpenJCEPlus consistent with that behavior.

I agree that if a caller already requested `ML-KEM-768` directly from a `KeyPairGenerator`, the information is redundant, but it is useful when the key is created through a generic ML-KEM/ML-DSA generator or decoded from an encoded key, because getAlgorithm() only exposes the family name while the encoded key still identifies a specific parameter set.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
```diff
@@ -1080,4 +1081,60 @@ public void testMLKEMGetAlgorithmConsistentAcrossProviders(String paramSetName)
...
+                ((NamedParameterSpec) openjceplusKeyPair.getPublic().getParams()).getName());
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+            "ML-KEM",
+            "ML-KEM-512",
+            "ML-KEM-768",
+            "ML-KEM-1024"
+    })
```
KostasTsiounis: Why not just add these parameters to the previous test?

KostasTsiounis (changes requested):

johnpeck-us-ibm:
> > Are we doing this for some JDK tests that need this? The parameters are all hard coded and if you know the alg you know the parameters since they do not change.
> 
> Not specifically for a JDK test. The main reason is to align our PQC key behavior with the upstream JCA API and the JDK providers.
> 
> `PublicKey` and `PrivateKey` now both extend `AsymmetricKey`, and `AsymmetricKey` introduced a `getParams()` method. Its default implementation returns `null`, so our PQC keys currently inherit that behavior unless we override it.
> 
> For PQC keys, `getAlgorithm()` now returns only the family name, such as `ML-KEM` or `ML-DSA`, not the specific parameter set. The JDK providers override `getParams()` for these keys and return the specific parameter set as a `NamedParameterSpec`, so this change makes OpenJCEPlus consistent with that behavior.
> 
> I agree that if a caller already requested `ML-KEM-768` directly from a `KeyPairGenerator`, the information is redundant, but it is useful when the key is created through a generic ML-KEM/ML-DSA generator or decoded from an encoded key, because getAlgorithm() only exposes the family name while the encoded key still identifies a specific parameter set.

I am fine with this change since it is being introduced as a result of the updates to the base classes.

getAlgorithm() should always return the actual algorithm of the key. The concept of a family name fro PQC Keys is an Oracle-specific construct.

Unlike RSA, DSA, and other traditional key types, the algorithm name for PQC keys identifies the key size and all associated parameters. PQC key encodings include the exact key algorithm, like ML-DSA-44

There is no single OID for ML-DSA or ML-KEM. Instead, the OIDs are defined for each specific algorithm and parameter set, such as ML_DSA_44 or ML_KEM_512.

By comparison, RSA has a general RSA OID, with parameters within the key determining characteristics such as key size. PQC algorithms do not work that way, as the specific algorithm identifier already defines the parameter set.

src/test/java/ibm/jceplus/junit/base/BaseTestPQCKeyInterop.java
JinhangZhang: I originally kept them separate because the Oracle interop test uses different providers for ML-KEM and ML-DSA (`SunJCE` for ML-KEM and `SUN` for ML-DSA). But the actual test logic is identical, so I agree they can be combined and just select the appropriate interop provider based on the algorithm family.

JinhangZhang:
> > > Are we doing this for some JDK tests that need this? The parameters are all hard coded and if you know the alg you know the parameters since they do not change.
> > 
> > 
> > Not specifically for a JDK test. The main reason is to align our PQC key behavior with the upstream JCA API and the JDK providers.
> > `PublicKey` and `PrivateKey` now both extend `AsymmetricKey`, and `AsymmetricKey` introduced a `getParams()` method. Its default implementation returns `null`, so our PQC keys currently inherit that behavior unless we override it.
> > For PQC keys, `getAlgorithm()` now returns only the family name, such as `ML-KEM` or `ML-DSA`, not the specific parameter set. The JDK providers override `getParams()` for these keys and return the specific parameter set as a `NamedParameterSpec`, so this change makes OpenJCEPlus consistent with that behavior.
> > I agree that if a caller already requested `ML-KEM-768` directly from a `KeyPairGenerator`, the information is redundant, but it is useful when the key is created through a generic ML-KEM/ML-DSA generator or decoded from an encoded key, because getAlgorithm() only exposes the family name while the encoded key still identifies a specific parameter set.
> 
> I am fine with this change since it is being introduced as a result of the updates to the base classes.
> 
> getAlgorithm() should always return the actual algorithm of the key. The concept of a family name fro PQC Keys is an Oracle-specific construct.
> 
> Unlike RSA, DSA, and other traditional key types, the algorithm name for PQC keys identifies the key size and all associated parameters. PQC key encodings include the exact key algorithm, like ML-DSA-44
> 
> There is no single OID for ML-DSA or ML-KEM. Instead, the OIDs are defined for each specific algorithm and parameter set, such as ML_DSA_44 or ML_KEM_512.
> 
> By comparison, RSA has a general RSA OID, with parameters within the key determining characteristics such as key size. PQC algorithms do not work that way, as the specific algorithm identifier already defines the parameter set.

That makes sense, and I agree. I am not suggesting that the family name is part of the PQC algorithm encoding model. The encoded key still identifies the exact algorithm/parameter set through its specific OID, such as ML-DSA-44 or ML-KEM-512.

My intent with this change is only to align with the current JDK API/base-class behavior, where `getAlgorithm()` exposes the family name and `AsymmetricKey.getParams()` provides the specific parameter set. So this does not change how the PQC key is encoded or identified internally; it just exposes the parameter-set information through the newer `getParams()` API.

